PR #70, opened 2020-10-11
Partial LLVM-6 support

wsmoses:
This restores Enzyme's ability to build against LLVM-6. We still should recommend against using LLVM-6, however, as it is not tested in CI (ubuntu doesn't have a package for) and a nontrivial amount of bugs in TBAA on the LLVM side.


wsmoses:
Resolves #69

vchuravy (approved):
Nice!

I don't personally care about LLVM 6 anymore,

PR #77, opened 2020-10-18
Additional tools to ease debugging of errors produced by Enzyme

wsmoses:
To follow #76 

Introduces helper utilities `__enzyme_float(ptr, size)` `__enzyme_double(ptr, size)` `__enzyme_integer(ptr, size)` and `__enzyme_pointer(ptr, size)` which explicitly informs type analysis of the type of [ptr, ptr+size).

Begins use of optimization remarks to provide debug information where type information not able to be determine. This allows source-code information about where type information needs to be injected.


vchuravy (approved):
Nice! Remarks are a great addition

PR #86, opened 2020-11-02
Handle rust pointer types, Fix Activity Analysis Bug, & Support more custom type analyses

wsmoses:


vchuravy (approved):
The rust rule is peculiar. ;)

PR #87, opened 2020-11-04
Restore exported C symbol

wsmoses:
This needs to return for Enzyme.jl to be happy (and yes we'll need to rebuild the binaries).


vchuravy:
Yeah having that seems key ;)

vchuravy (approved):

PR #97, opened 2020-12-20
Create static lib

tiberiusferreira:
The default created lib using the MODULE config generates a "bundle" in MacOS, which gives the following error when trying to link against in order to use the C api: `ld: can't link with bundle (MH_BUNDLE) only dylibs (MH_DYLIB)` .

It would be nice if there was a way to create a static lib (not dynamic, so no runtime surprises), this PR enables it. If this is not the best way to do it, let me know.


wsmoses:
Could you also add the static library build to the other build configurations in that file (if LLVM version < 8, windows)?

tiberiusferreira:
I'm not sure what the equivalent would be for the targets using `add_llvm_loadable_module` since I can't test them locally, so I tried just adding the STATIC keyword there.

wsmoses:
The answer is seemingly no, given the "Enzyme CI" failures. I'll take a look at this week and see how to fix.

wsmoses:
Upon reflection, for now we should just static library build for LLVM versions before 8 -- there are some intricacies in the build process that shouldn't delay this PR.

PR #107, opened 2021-01-23
Fix Julia TBAA

wsmoses:


vchuravy (approved):
Can we get away with removing the Julia tbaa logic from Enzyme proper since we now pass our own typetree?

wsmoses:
I don't think we can eliminate it yet -- Julia currently only pass the input/return trees, but there's still a potential need for information about load/stores.

PR #124, opened 2021-02-06
LLVM 11+ Vector handling and conditional external shared library build

wsmoses:


vchuravy (approved):
Can we add a build of the external shared library to CI?

wsmoses:
Yes, but in a separate PR.

PR #162, opened 2021-03-15
Misc cache optimizations per PTX

wsmoses:


enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -85,264 +85,367 @@ bool is_load_uncacheable(
...
-    }
-    assert(found != uncacheable_args.end());
-    if (found->second) {
-      mustcache = true;
-    }
-  } else {
+      assert(found != uncacheable_args.end());
+      if (found->second) {
+        mustcache = true;
+        // EmitWarning("UncacheableOrigin", arg->getDebugLoc(), oldFunc,
```
vchuravy, line 133: Is this warning still relevant?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2317,23 +2419,19 @@ Function *CreatePrimalAndGradient(
...
-          ? augmenteddata->can_modref_map
-          : compute_uncacheable_load_map(gutils->oldFunc, AA, TLI,
-                                         unnecessaryInstructionsTmp,
-                                         _uncacheable_argsPP, topLevel);
+      augmenteddata ? augmenteddata->can_modref_map
+                    : CA.compute_uncacheable_load_map();
   gutils->can_modref_map = &can_modref_map;
 
-  gutils->forceContexts();
+  // gutils->forceContexts();
```
vchuravy, line 2434: rm

vchuravy (approved):

enzyme/Enzyme/EnzymeLogic.cpp
wsmoses: I'd like to preserve this for the time being. It does seem reasonable that we can eliminate it and rely on the loop bounds to be generated laziliy, but if later we find this not to be the case it would be nice to see this as the place to return to.

enzyme/Enzyme/EnzymeLogic.cpp
wsmoses: Potentially still relevant, though an Argument doesn't have debug info associated to attach to =/

PR #164, opened 2021-03-19
Re-enable attributor and use proper memory management for compiler internals

wsmoses:
This contains a breaking change to the C ABI.


vchuravy (approved):
Confirmed good from the Julia side

PR #165, opened 2021-03-22
Fix example code

tgymnich:
The example code on the website is missing a cast to a void pointer.


wsmoses (approved):
LGTM. I think this is only necessary for C++ (hence the earlier code).

PR #169, opened 2021-03-31
[TypeAnalysis] Begin tensor core support

wsmoses:


vchuravy:
Can we split this PR  into it's three components? I would like to review the shared mem fwd separately.

wsmoses:
Agreed and split [and fixed CI]. Take a look at just the tensor core starting change here, and other parts to come.

PR #171, opened 2021-04-03
Fix address space bug in cache LICM

wsmoses:
I believe this should remedy the julia issue you saw @vchuravy

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -3090,10 +3090,16 @@ Value *GradientUtils::lookupM(Value *val, IRBuilder<> &BuilderM,
                                   /*extraSize*/ lim);
 
               auto dst_arg = v.CreateBitCast(
-                  outer, Type::getInt8PtrTy(inst->getContext()));
+                  outer,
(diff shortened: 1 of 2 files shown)


vchuravy:
Can confirm that this fixes the module verification error I encountered.

vchuravy:
But we seem to synthesize shared memory writes with the wrong alignment. Running under `compute-sanitizer --tool memcheck --launch-timeout=0 --show-backtrace=no --target-processes=all` yields:

```
========= Invalid __shared__ write of size 16 bytes
=========     at 0xb0 in julia_dvolumerhs__5923(CuDeviceArray<Float32,int=5,int=1>,CuDeviceArray<Float32,int=5,int=1>,CuDeviceArray<Float32,int=5,int=1>,CuDeviceArray<Float32,int=5,int=1>,CuDeviceArray<Float32,int=5,int=1>,CuDeviceArray<Float32,int=5,int=1>,Float32,CuDeviceArray<Float32,int=2,int=1>,CuDeviceArray<Float32,int=2,int=1>,Int64)
=========     by thread (2,1,1) in block (0,0,0)
=========     Address 0x274 is misaligned
```

PR #172, opened 2021-04-04
Begin alignment investigation

wsmoses:


vchuravy (commented):
Fixes the alignment disagreement for shared memory.

vchuravy (approved):

PR #177, opened 2021-04-19
Primitive Shared Memory Forwarding

wsmoses:


enzyme/test/Enzyme/sharedcachefwd.ll
```diff
@@ -0,0 +1,420 @@
+; RUN: if [ %llvmver -ge 9 ]; then %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -enzyme-shared-forward -mem2reg -sroa -instsimplify -early-cse -simplifycfg -S | FileCheck %s; fi
+
+; TODO the code generated here may cause illegal reads as the phi recopmutation does not check that a load is legal to recompute for
+; bounds reasons, only aliasing reasons. This code illustrates an example where it is incorrect and should be remedied. Now, the value
+; will never be used and not present a correctness error, but may present a segfault where one need not occur.
```
vchuravy, line 5: How hard is this to fix?

vchuravy (commented):
Do we have a benchmark where we saw this to be of immediate value? IIRC the general idea here is instead of forwarding the value of the load from the forward to backward pass, we instead check whether or not the load will be ever be clobbered and if it will not be clobbered refetch the value in the reverse?

enzyme/test/Enzyme/sharedcachefwd.ll
wsmoses: So the fix simply requires making a phi node in the reverse pass and conditionally executing the inputs. From that standpoint its relatively easy. The harder part is that it breaks the assumption that every forward block corresponds to one reverse block. I thought a bit about trying to work around this with a subfunction we call and immediately inline (like we do for differential memcpy, etc), but that won't work as easily here. Regardless the bug isn't introduced by this PR (though this test case does demonstrate it).

wsmoses:
The cuda matrix multiply directly uses it (the code that generated this IR). This is explicitly disabled by default presently though as it makes an assumption about the size of the block.

vchuravy (approved):

PR #182, opened 2021-04-26
Forward mode

tgymnich:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -208,7 +208,8 @@ class AdjointGenerator
         eraseIfUnused(inst);
         if (gutils->isConstantInstruction(&inst))
           return;
-        if (Mode != DerivativeMode::Reverse && Mode != DerivativeMode::Both)
+        if (Mode != DerivativeMode::Reverse && Mode != DerivativeMode::Both &&
```
wsmoses, line 211: This should probably just be if augmented forward, unless I'm misreading

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -979,15 +983,6 @@ class AdjointGenerator
 
   void visitBinaryOperator(llvm::BinaryOperator &BO) {
     eraseIfUnused(BO);
-    if (gutils->isConstantInstruction(&BO))
```
wsmoses, line 982: I think you'll still want the activity checks for fwd mode.

wsmoses: To avoid inadvertently differentiating a +[integer] b

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -1348,6 +1365,105 @@ class AdjointGenerator
...
+        setDiffe(&BO, idiff, Builder2);
+      } else if (constantval1) {
+        Value *idiff = Builder2.CreateFMul(orig_op1, dif1);
+        setDiffe(&BO, idiff, Builder2);
+      } else {
+        Value *idiff0 =
+            Builder2.CreateFMul(dif0, gutils->getNewFromOriginal(orig_op0));
+        Value *idiff1 =
+            Builder2.CreateFMul(dif1, gutils->getNewFromOriginal(orig_op1));
+        addToDiffe(&BO, idiff0, Builder2, addingType);
```
wsmoses, line 1408: Why is this dual += whereas above it is dual =.

I might also restructure this code to 

if (!constantval0) {
   diffe(op0) = ...
}
if (!constantval1) {
  diffe(op1) = ..
}

there'd be less cases that way and perhaps easier to debug.

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -326,6 +327,12 @@ class Enzyme : public ModulePass {
 
       constants.push_back(ty);
 
+      for (auto i = constants.begin(); i != constants.end() && fwdMode; ++i) {
```
wsmoses, line 330: This is reasonable a reasonable start.

Something to consider (need not be now). you may alternatively want to change the rule for auto detecting diffe type to have this be dup by default for forward mode. This way you can catch if someone passes the out_diff metadata to a forward mode call as an error.

enzyme/Enzyme/GradientUtils.h
```diff
@@ -71,7 +71,7 @@
 
 using namespace llvm;
 
-enum class DerivativeMode { Forward, Reverse, Both };
+enum class DerivativeMode { AugmentedForward, Forward, Reverse, Both };
```
wsmoses, line 74: Need not be this PR but I may rename this so old Forward is unambiguously becomes "Augmented Forward Pass of reverse mode" and the new one is "Forward Mode AD"

wsmoses (changes requested):
Suggestions inline.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: Activity checks are on line 1414 and 1418

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -1348,6 +1365,105 @@ class AdjointGenerator
...
+      } else {
+        idiff1 = ConstantFP::get(addingType, 0.0);
+      }
+
+      Value *idiff2;
+      if (!constantval1) {
+        idiff2 =
+            Builder2.CreateFMul(gutils->getNewFromOriginal(orig_op0), dif1);
+      } else {
+        idiff2 = ConstantFP::get(addingType, 0.0);
```
wsmoses, line 1447: If something is inactive (constant) then there isn't a shadow for you to update, right?

enzyme/Enzyme/AdjointGenerator.h
wsmoses: Yes, for the operands of the individiual operator. It may still be useful to do an initial check of the entire instruction is inactive (and thus you don't need to run createBinaryOperatorDual /  createBinaryOperatorAdjoint at all)

enzyme/Enzyme/AdjointGenerator.h
tgymnich: do you mean `DerivativeMode::Forward`?

wsmoses: Yeah I mean the existing forward pass in reverse mode. I guess to me "augmented forward" feels more like the "augmented forward pass needed for reverse mode" than "forward mode" so perhaps a name change may be warranted?

enzyme/Enzyme/GradientUtils.h
tgymnich: How about:
AugmentedForward -> Forward
Forward -> Passthrough

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -818,14 +837,22 @@ class Enzyme : public ModulePass {
...
+    for (auto CI : toLowerFwd) {
+      successful &= HandleAutoDiff(CI, TLI, PostOpt, true);
       Changed = true;
       if (!successful)
         break;
     }
+
     for (auto CI : toLowerI) {
-      successful &= HandleAutoDiff(CI, TLI, PostOpt);
+      successful &= HandleAutoDiff(CI, TLI, PostOpt, false);
```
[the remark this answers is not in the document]
tgymnich: done 044f04515de5bfc166c9716072f7122294947844

enzyme/Enzyme/GradientUtils.h
wsmoses: Passthrough feels weird in that I wouldn't understand what is meant by it.

Maybe:
AugmentedForward => ForwardMode
Forward => ReverseModePrimal
Reverse => ReverseModeGradient
Both      => ReverseModeCombined

@vchuravy thoughts as well?

tgymnich: shouldn't it be: AugmentedForward => ForwardModeDirectionalDerivative

wsmoses: I don't think there's forward mode AD that isn't directional in this sense so it feel redundant?

vchuravy: ++ for the ReverseMode rename.

tgymnich: done 5b45bf589e27fd83f9c586798aa38979e3317eef

enzyme/Enzyme/Enzyme.cpp
tgymnich: Since we've now defined the calling convention for the forward mode, I went ahead and reworked this in 476ef9da884901061662f40ffad92ff28b07d765

tgymnich:
@wsmoses @ludgerpaehler @vchuravy this should be ready now.

wsmoses (approved):

PR #193, opened 2021-05-17
Forward mode load store

tgymnich:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -378,15 +384,37 @@ class AdjointGenerator
...
-        ((DiffeGradientUtils *)gutils)
-            ->addToInvertedPtrDiffe(I.getOperand(0), prediff, Builder2,
-                                    alignment, OrigOffset);
+      switch (Mode) {
+      case DerivativeMode::ForwardMode: {
+        IRBuilder<> Builder2(&I);
+        getForwardBuilder(Builder2);
+
+        auto diff = Builder2.CreateLoad(
+            gutils->invertPointerM(I.getOperand(0), Builder2));
```
wsmoses, line 394: I'd move this invert pointer load inot the isconstantvalue check, since only needed in that case.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -511,13 +540,29 @@ class AdjointGenerator
...
           addToDiffe(orig_val, dif1, Builder2, FT);
         }
+        break;
+      }
+      case DerivativeMode::ForwardMode: {
+        IRBuilder<> Builder2(&SI);
+        getForwardBuilder(Builder2);
+
+        if (constantval) {
+          ts = setPtrDiffe(orig_ptr, Constant::getNullValue(valType), Builder2);
```
wsmoses, line 550: I don't think this set if load is inactive is correct. I think no store should happen in that case.

E.g. consider

```
%unused = load %x
%toret = load %x
ret %toret
```

The unused value would cause d_x to be set to zero, making the eventual derivative zero, which seems incorrect.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -955,6 +1000,22 @@ class AdjointGenerator
     Builder2.setFastMathFlags(getFast());
   }
 
+  inline void getForwardBuilder(IRBuilder<> &Builder2) {
```
wsmoses, line 1003: If you're going to make this new helper function, I'd suggest also reworking all the existing code that creates a forward builder (usually named BuilderZ or something) to use the helper as well.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2069,9 +2069,9 @@ void createTerminator(DiffeGradientUtils *gutils,
 
     auto retVal = inst->getOperand(0);
```
wsmoses, line 2071: Why is isConstantValue insufficient here and we need to also explicitly check the diffe_type?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2611,15 +2611,16 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
   assert(!todiff->empty());
 
   ReturnType retVal =
-      returnValue ? (dretPtr ? ReturnType::ArgsWithTwoReturns
-                             : ReturnType::ArgsWithReturn)
-                  : (dretPtr ? ReturnType::ArgsWithReturn : ReturnType::Args);
+      fwdMode       ? ReturnType::Return
```
wsmoses, line 2614: Technically won't foward mode need to be either return or two returns, depending on whether the return is active?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2820,7 +2821,9 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
 
     for (size_t i = 0; i < constant_args.size(); ++i) {
       auto arg = constant_args[i];
-      if (arg == DIFFE_TYPE::DUP_ARG) {
+      auto argTypeInfo = oldTypeInfo.Function->args().begin() + i;
+      if (arg == DIFFE_TYPE::DUP_ARG &&
+          !argTypeInfo->getType()->isPointerTy()) {
```
wsmoses, line 2826: Shouldn't all active arguments be dup_arg (so perhaps change that to an assert)? Also please use type analysis info here rather than a pure pointer check.

enzyme/Enzyme/FunctionUtils.cpp
```diff
@@ -1384,9 +1384,11 @@ Function *PreProcessCache::CloneFunctionWithReturns(
   F = preprocessForClone(F, topLevel);
   std::vector<Type *> RetTypes;
   if (returnValue == ReturnType::ArgsWithReturn ||
-      returnValue == ReturnType::ArgsWithTwoReturns ||
-      returnValue == ReturnType::Return ||
-      returnValue == ReturnType::TwoReturns)
+      returnValue == ReturnType::ArgsWithTwoReturns)
+    RetTypes.push_back(F->getReturnType());
+  if ((returnValue == ReturnType::Return ||
```
wsmoses, line 1389: The void type error here should also apply to the argswithreturn / argswithtworeturns right?

I'd suggest you move the check for void up the chain for consistency (and generally have it become an activity of return check).

enzyme/Enzyme/AdjointGenerator.h
tgymnich: makes sense 😄

enzyme/Enzyme/AdjointGenerator.h
tgymnich: I could only find two such instances of BuilderZ, moved the helpers to GradientUtils and made those instances use getForwardBuilder

enzyme/Enzyme/EnzymeLogic.cpp
tgymnich: Yes. I wanted to deal with that later.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: This is a store

enzyme/Enzyme/EnzymeLogic.cpp
tgymnich: Yes, this should become unnecessary once I've figured out the activity of return check

enzyme/Enzyme/EnzymeLogic.cpp
tgymnich: Yes, this should become unnecessary once I've figured out the activity of return check

enzyme/Enzyme/AdjointGenerator.h
wsmoses: Ah my mistake, this is fine then.

enzyme/Enzyme/FunctionUtils.cpp

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -270,10 +270,13 @@ class AdjointGenerator
         IRBuilder<> BuilderZ(newi);
         Value *newip = nullptr;
 
-        bool needShadow = is_value_needed_in_reverse<ValueType::ShadowPtr>(
-            TR, gutils, &I,
-            /*toplevel*/ Mode == DerivativeMode::ReverseModeCombined,
-            oldUnreachable);
+        bool needShadow =
+            Mode == DerivativeMode::ForwardMode
+                ? false
```
wsmoses, line 275: In retrospect I think the answer here is true if the loaded value is used as a pointer. In that case you need to make sure a shadow load is available in the fwd pass. In any case I'd leave a comment, then try testing this.

wsmoses, line 275: I think the correct answer here is that this should be true if the loaded value itself is used as a pointer. I'm fine leaving this as is for now, but at least make a note here and try writing a test that exhibits this property.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2831,8 +2841,8 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
 
     for (size_t i = 0; i < constant_args.size(); ++i) {
       auto arg = constant_args[i];
+      newArgs += 1;
```
wsmoses, line 2844: This should only be in the dup case, otherwise you increase twice for a constant arg (worthwhile adding a test for)

wsmoses (approved):
Only relatively minor comments. Please fix the dup issue and add the comment for load, then feel free to squash & rebase into main.

PR #194, opened 2021-05-17
Forward mode cast inst

tgymnich:
implemented forward mode CastInst

Tests will follow in #198

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -608,51 +608,104 @@ public:
         I.getOpcode() == CastInst::CastOps::PtrToInt)
       return;
 
-    if (Mode == DerivativeMode::ReverseModePrimal)
+    switch (Mode) {
+    case DerivativeMode::ReverseModePrimal: {
       return;
-
-    Value *orig_op0 = I.getOperand(0);
-    Value *op0 = gutils->getNewFromOriginal(orig_op0);
-
-    IRBuilder<> Builder2(I.getParent());
-    getReverseBuilder(Builder2);
-
-    if (!gutils->isConstantValue(orig_op0)) {
-      Value *dif = diffe(&I, Builder2);
-
-      size_t size = 1;
-      if (orig_op0->getType()->isSized())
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -608,51 +608,104 @@ class AdjointGenerator
...
+    case DerivativeMode::ForwardMode: {
+      Value *orig_op0 = I.getOperand(0);
+
+      IRBuilder<> Builder2(&I);
+      getForwardBuilder(Builder2);
+
+      if (!gutils->isConstantValue(orig_op0)) {
+        Value *dif = diffe(orig_op0, Builder2);
+
+        size_t size = 1;
```
wsmoses, line 672: I don't think you need the type check here, the fwd diff of a cast should just be setting the shadow of the return to cast(shadow(input)), regardless of type/cast, right?

wsmoses:
I still think this is worth creating a minimal ll test for. E.g. making a single function that does fpext of input and returns it.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: right

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -608,51 +608,90 @@ class AdjointGenerator
...
+      break;
+    }
+    case DerivativeMode::ForwardMode: {
+      Value *orig_op0 = I.getOperand(0);
+
+      IRBuilder<> Builder2(&I);
+      getForwardBuilder(Builder2);
+
+      if (!gutils->isConstantValue(orig_op0)) {
+        Value *dif = diffe(orig_op0, Builder2);
```
wsmoses, line 670: This still should be able to be simplified. E.g. it could have a single case like setDiffe(&I, Builder.CreateCast(diff, I.getType(), I.getOpcode())

tgymnich: done

wsmoses (approved):

PR #198, opened 2021-05-22
ForwardMode phi

tgymnich:
Added derivative of phi for forward mode

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -611,22 +611,16 @@ public:
     }
     }
 
-    size_t size = 1;
-    if (phi.getType()->isSized())
-      size = (gutils->newFunc->getParent()->getDataLayout().getTypeSizeInBits(
-                  phi.getType()) +
-              7) /
-             8;
+    BasicBlock *oBB = phi.getParent();
+    BasicBlock *nBB = gutils->getNewFromOriginal(oBB);
 
-    if (phi.getType()->isIntOrIntVectorTy() &&
-        TR.intType(size, &phi, /*errifnotfound*/ false) == BaseType::Pointer) {
-      return;
-    }
+    IRBuilder<> diffeBuilder(nBB->getFirstNonPHI());
+    diffeBuilder.setFastMathFlags(getFast());
 
-    IRBuilder<> Builder2(&phi);
-    getForwardBuilder(Builder2);
+    IRBuilder<> phiBuilder(&phi);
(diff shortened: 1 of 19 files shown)


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -395,6 +395,8 @@ class AdjointGenerator
           auto diff = Builder2.CreateLoad(
               gutils->invertPointerM(I.getOperand(0), Builder2));
           setDiffe(&I, diff, Builder2);
+        } else {
+          setDiffe(&I, Constant::getNullValue(I.getType()), Builder2);
```
wsmoses, line 399: Why is this necessary, there shouldn't be a diffe set for inactive values?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -1466,6 +1511,8 @@ class AdjointGenerator
 
     Type *addingType = BO.getType();
 
+    setDiffe(&BO, Constant::getNullValue(BO.getType()), Builder2);
```
wsmoses, line 1514: Why do we need this? If this op is constant we should never set it, and if it is active we should always set it to the computed derivative?

enzyme/Enzyme/AdjointGenerator.h
tgymnich: This was not supposed to be in this PR.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: Because I am using `addToDiffe` I need to "reset" variables that are used in loops. If I stop using `addToDiffe` I can get rid of this.

wsmoses: We should just use setDiffe for forward mode (and arguably can assert in addToDiffe if in forwardMode).

tgymnich: Already working on this 😄

wsmoses: Bump at potentially removing this?

wsmoses:
Besides the unnecessary setDiffe, LGTM.

wsmoses (approved):

enzyme/Enzyme/AdjointGenerator.h
tgymnich: completely forgot to remove this 🙈

PR #201, opened 2021-05-23
Support Bitcode Library

wsmoses:
Disabled by default.


vchuravy:
Can we separate out the invoke changes? I would like to hold on the bc changes until we have confirmed correctness, of the implementation.

wsmoses:
I'm fine separating out the invoke, though I'm inclined to get the bitcode library pieces in. I agree we should double check that reduction, but this change primarily enables the use of a bitcode library (off by default) and can have the reduction interchaged.

vchuravy (changes requested):
Fixed bitcode, including runtime correctness tests.

PR #221, opened 2021-07-01
Implement visitMemSetInst

tgymnich:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -1751,18 +1752,39 @@ class AdjointGenerator
...
-  void subTransferHelper(Type *secretty, BasicBlock *parent,
-                         Intrinsic::ID intrinsic, unsigned dstalign,
-                         unsigned srcalign, unsigned offset, Value *orig_dst,
-                         Value *orig_src, Value *length, Value *isVolatile,
-                         llvm::CallInst *MTI, bool allowForward = true) {
+  void subTransferHelper(Type *secretty, llvm::MemTransferInst &MTI,
+                         unsigned dstalign, unsigned srcalign, unsigned offset,
+                         Value *length, bool allowForward = true) {
+
+    BasicBlock *parent = MTI.getParent();
```
wsmoses, line 1759: I don't think you want to call this for forward mode (it is designed to do the += required by reverse, with correctness for types). You should simply be able to do a memcpy regardless of type, and since you only need to set not increment, this shouldn't be required?

wsmoses (approved):
I’d move the calculation of size after the forward mode sits it’s not used in that case but otherwise LGTM feel free to squash/merge.

PR #226, opened 2021-07-08
Use calls to cblas_daxpy to calculate the derivative of functions with cblas_ddot

reikdas:


enzyme/test/Enzyme/ReverseMode/cblas_ddot_double.ll
```diff
@@ -0,0 +1,191 @@
...
+@__const.main.n = private unnamed_addr constant [3 x double] [double 4.000000e+00, double 5.000000e+00, double 6.000000e+00], align 16
+
+; Function Attrs: noinline nounwind optnone uwtable
+define dso_local double @g(double* %m, double* %n) {
+entry:
+  %m.addr = alloca double*, align 8
+  %n.addr = alloca double*, align 8
+  %x = alloca double, align 8
+  %y = alloca double, align 8
+  store double* %m, double** %m.addr, align 8
```
wsmoses, line 42: Can you clean up the test a bit? There's no need for all of the m.addr / x, etc (e.g. run mem2reg on it).

Similarly I would run -mem2reg -instsimplify -simplifycfg after enzyme to make the gradient easier to read.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3724,6 +3726,117 @@ class AdjointGenerator
       return;
     }
 
+    if (funcName == "cblas_ddot") {
+      Type *castvals[2] = {call.getArgOperand(1)->getType(),
+                           call.getArgOperand(3)->getType()};
+      auto tupletype = StructType::get(call.getContext(), ArrayRef(castvals));
+      Value *tupleholder = BuilderZ.CreateAlloca(tupletype);
```
wsmoses, line 3733: Rather than making an allocation here and gep/storing into it, can you initialize to an undef then use an insertvalue?

Similarly it may be wise to first determine which values need caching and then only caching those values. For example. in ddot(a, b) you only need to cache a if b is active (since if b is inactive you shouldn't update the shadow of b, since its inactive [and otherwise may get incorrect behavior]). Additionally if a isn't overwritten between the forward and reverse (even if b is active), you don't need to cache it. The (poorly named) variable uncacheable_args_map should contain a map of arguments to whether or not the memory is overwritten.

enzyme/test/Enzyme/ReverseMode/cblas_ddot_double.ll
```diff
@@ -0,0 +1,191 @@
...
+  store double* %n, double** %n.addr, align 8
+  %0 = load double*, double** %m.addr, align 8
+  %1 = load double*, double** %n.addr, align 8
+  %call = call double @cblas_ddot(i32 3, double* %0, i32 1, double* %1, i32 1)
+  store double %call, double* %x, align 8
+  %2 = load double*, double** %m.addr, align 8
```
wsmoses, line 48: I'd also explicitly add a test where one of the inputs isn't active as well.

enzyme/Enzyme/AdjointGenerator.h
wsmoses: I might also write this to handle either ddot for doubles or sdot for floats in the same call

reikdas: > For example. in ddot(a, b) you only need to cache a if b is active (since if b is inactive you shouldn't update the shadow of b, since its inactive [and otherwise may get incorrect behavior]).

I am trying to understand this - in `ddot(a, b)`, there should never be a situation where either of them is inactive right? Unless you mean that `a` is inactive for `d/d_a ddot(a,b)` (for the cases where adjoint doesn't depend on `a`), and `b` is inactive for `d/d_b ddot(a,b)` (for the cases where adjoint doesn't depend on `b`).

wsmoses: It's possible for one of the two inputs to be inactive.

For example consider

```
double sum(double* a) {
  double b[3] = {1, 1, 1};
  return ddot(a, b);
}
```

Here b is inactive, even though a is active. As such we don't need to cache a since we don't need to (and in fact would be illegal to) compute the derivative of b.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3838,6 +3841,306 @@ class AdjointGenerator
...
+              BuilderZ.CreateBitCast(malins1, call.getArgOperand(1)->getType());
+          auto memcpy_size1 =
+              (int)(call.getFunction()
+                        ->getParent()
+                        ->getDataLayout()
+                        .getTypeAllocSize(
+                            newcall->getArgOperand(1)->getType())) *
+              dyn_cast<ConstantInt>(call.getArgOperand(0))->getSExtValue();
+#if LLVM_VERSION_MAJOR >= 10
+          BuilderZ.CreateMemCpy(arg1, MaybeAlign(), newcall->getArgOperand(1),
```
wsmoses, line 3882: I'm a bit worried about the correctness of the memcpy here. The spec specifies a stride (which so far has been tested wiht only 1). But suppose an stride of 4 were used, we won't copy all the data needed. Arguably in that case we should still allocate the same amount but do something like cache[i] = input[4*i], then in the reverse should use a size of 1 for the cached version (since we're storing at stride of 1).

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3838,6 +3841,306 @@ class AdjointGenerator
...
+        if (!gutils->isConstantValue(call.getArgOperand(3)) && xoverwrite)
+          CallInst::CreateFree(structarg1, firstdcall->getNextNode());
+        if (!gutils->isConstantValue(call.getArgOperand(1)) && yoverwrite)
+          CallInst::CreateFree(structarg2, seconddcall->getNextNode());
+      }
+      return;
+    }
+
+    if (funcName == "cblas_sdot" && called->isDeclaration()) {
+      Type *castvals[2] = {call.getArgOperand(1)->getType(),
```
wsmoses, line 3995: I'd have this be handled in the same code as above with a if funcName == "cblas_ddot" || funcName == "cblas_sdot". Then when it differs (per type), you can do a check whether it's ddot or sdot

enzyme/test/Enzyme/ReverseMode/cblas_sdot_nomod.ll
```diff
@@ -0,0 +1,90 @@
+;RUN: %opt < %s %loadEnzyme -enzyme -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
```
wsmoses, line 1: I might make a new subfolder called blas in reverse mode for ease.

enzyme/Enzyme/AdjointGenerator.h
reikdas: Just to clarify,

For a strided implementation where we do something like - 
```
g(double *m, double* m1, double *n, double *n1) { // Where m1, n1 are shadow args
  cblas_ddot(3, m, 2, n, 3); // 2 is the stride for m. 3 is the stride for n.
  ....
}
  
int caller() {
  double m[6] = {1, 2, 3, 4, 5, 6};
  double m1[6] = {0, 0, 0, 0, 0, 0};
  double n[9] = {1, 2, 3, 4, 5, 6, 7, 8, 9};
  double n1[9] = {0, 0, 0, 0, 0, 0, 0, 0, 0};
  g(m, m1, n, n1);
```
If cblas_ddot's source code is available, we store the derivatives wrt m in m1[0], m1[2], m1[4]. And the derivatives wrt n in n1[0], n1[3], n1[6].

If I understand correctly, in the strided memcpy implementation that you suggest, we would be storing the derivatives wrt m in m1[0], m1[1], m1[2]. And the derivatives wrt n in n1[0], n1[1], n1[2]. Is that okay?

wsmoses: So we still want the same end derivative -- but we don't need to cache the values in the same way. So even though here m has a stride of 2, we can condense the cache to have a stride of 1 when we copy into it. This should still produce the same result, so long as we now a stride of 1 when indexing into the cached m for the reverse.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3976,6 +3979,150 @@ class AdjointGenerator
...
+      Argument *xfuncarg = in_arg;
+      in_arg++;
+      in_arg++;
+      Argument *yfuncarg = in_arg;
+      bool xoverwrite = uncacheable_args.find(xfuncarg)->second;
+      bool yoverwrite = uncacheable_args.find(yfuncarg)->second;
+      bool xcache =
+          !gutils->isConstantValue(call.getArgOperand(3)) &&
+          (xoverwrite ||
+           cast<ConstantInt>(call.getArgOperand(2))->getSExtValue() != 1);
```
wsmoses, line 4011: This will cause an error, you can't assume that the argument is always a constant int. Moreover, you don't need to cache if it's not overwritten, even if the stride != 1.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3976,6 +3979,150 @@ class AdjointGenerator
...
+          (yoverwrite ||
+           cast<ConstantInt>(call.getArgOperand(4))->getSExtValue() != 1);
+      if ((Mode == DerivativeMode::ReverseModeCombined ||
+           Mode == DerivativeMode::ReverseModePrimal) &&
+          (xcache || ycache)) {
+        BuilderZ.SetInsertPoint(gutils->getNewFromOriginal(&call));
+        Value *arg1, *arg2;
+        auto size = ConstantExpr::getSizeOf(innerType);
+        if (xcache) {
+          auto dmemcpy = getOrInsertDifferentialMemcpyStrided(
```
wsmoses, line 4023: Change this to getOrInsertMemcpyStrided (since here it shouldn't be doing the differential memcpy, just a strided memcpy).

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3976,6 +3979,150 @@ class AdjointGenerator
...
+        }
+        CallInst *firstdcall, *seconddcall;
+        if (!gutils->isConstantValue(call.getArgOperand(3))) {
+          assert(!isa<Constant>(orig->getArgOperand(3)));
+          SmallVector<Value *, 6> args1 = {
+              lookup(gutils->getNewFromOriginal(orig->getArgOperand(0)),
+                     Builder2),
+              diffe(orig, Builder2),
+              structarg1,
+              Builder2.getInt32(1),
```
wsmoses, line 4098: This should be the actual stride of x, unless it's cached at which point it can be 1.

enzyme/Enzyme/Utils.cpp
```diff
@@ -158,6 +158,98 @@ Function *getOrInsertDifferentialFloatMemcpy(Module &M, PointerType *T,
...
+#else
+  Function *F = cast<Function>(M.getOrInsertFunction(name, FT));
+#endif
+
+  if (!F->empty())
+    return F;
+
+  F->setLinkage(Function::LinkageTypes::InternalLinkage);
+  F->addFnAttr(Attribute::ArgMemOnly);
+  F->addFnAttr(Attribute::NoUnwind);
```
wsmoses, line 186: Probably also worth adding readonly to src and writeonly (if sufficient llvm version) to dst.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3976,6 +3979,152 @@ class AdjointGenerator
...
+      }
+      if (Mode == DerivativeMode::ReverseModeCombined ||
+          Mode == DerivativeMode::ReverseModeGradient) {
+        IRBuilder<> Builder2(call.getParent());
+        getReverseBuilder(Builder2);
+        auto derivcall = gutils->oldFunc->getParent()->getOrInsertFunction(
+            dfuncName, Builder2.getVoidTy(), Builder2.getInt32Ty(), innerType,
+            call.getArgOperand(1)->getType(), Builder2.getInt32Ty(),
+            call.getArgOperand(3)->getType(), Builder2.getInt32Ty());
+        Value *structarg1 = lookup(
```
wsmoses, line 4064: For performance, it's good to avoid a lookup if not necessary. Can you initialize this to nullptr, then if it's not set from the cache, do this lookup?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3976,6 +3979,152 @@ class AdjointGenerator
...
+            structarg1 = BuilderZ.CreateExtractValue(cacheval, 0);
+            structarg2 = BuilderZ.CreateExtractValue(cacheval, 1);
+          } else if (xcache)
+            structarg1 = cacheval;
+          else if (ycache)
+            structarg2 = cacheval;
+        }
+        CallInst *firstdcall, *seconddcall;
+        if (!gutils->isConstantValue(call.getArgOperand(3))) {
+          assert(!isa<Constant>(orig->getArgOperand(3)));
```
wsmoses, line 4086: Why is this assertion here?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3976,6 +3979,152 @@ class AdjointGenerator
...
+              diffe(orig, Builder2),
+              structarg1,
+              estride,
+              gutils->invertPointerM(orig->getArgOperand(3), Builder2),
+              lookup(gutils->getNewFromOriginal(orig->getArgOperand(4)),
+                     Builder2)};
+          firstdcall = Builder2.CreateCall(derivcall, args1);
+        }
+        if (!gutils->isConstantValue(call.getArgOperand(1))) {
+          assert(!isa<Constant>(orig->getArgOperand(1)));
```
wsmoses, line 4103: and here?

enzyme/Enzyme/Utils.cpp
```diff
@@ -158,6 +158,98 @@ Function *getOrInsertDifferentialFloatMemcpy(Module &M, PointerType *T,
...
 }
 
+Function *getOrInsertMemcpyStrided(Module &M, PointerType *T, unsigned dstalign,
+                                   unsigned srcalign, Value *stride) {
+  Type *elementType = T->getElementType();
+  assert(elementType->isFloatingPointTy());
+  std::string name = "__enzyme_memcpy_" + tofltstr(elementType) + "da" +
+                     std::to_string(dstalign) + "sa" +
+                     std::to_string(srcalign) + "stride" +
+                     std::to_string(cast<ConstantInt>(stride)->getSExtValue());
```
wsmoses, line 168: Don't do a cast to constantint -- it may not be constant

enzyme/test/Enzyme/ReverseMode/blas/cblas_ddot_inactive_second.ll
```diff
@@ -0,0 +1,85 @@
+;RUN: %opt < %s %loadEnzyme -enzyme -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
+
+;#include <cblas.h>
+;
+;extern double __enzyme_autodiff(double*, double*, double*);
+;
+;double g(double *restrict n) {
+;    double m[3] = {1, 2, 3};
+;	 double x = cblas_ddot(3, m, 1, n, 1);
+;    double y = x*x;
```
wsmoses, line 10: Can you add a test where the stride is an inactive argument to g?

enzyme/Enzyme/Utils.cpp
reikdas: Could you please help me figure out an alternate way to make the name of the function unique for that particular stride number?

wsmoses:
Since you’re passing it as an argument I don’t think it needs to be unique?
Perhaps you need to unique it based off of the type of the stride var?

enzyme/test/Enzyme/ReverseMode/blas/cblas_ddot_inactive_second.ll
reikdas: Would this be an appropriate example? 

```
float g(float *restrict m, float *restrict n) {
  int stride = 3;
  float x = cblas_sdot(3, m, 2, n, stride);
  float y = x * x;
  return y;
}

int main() {
  float m[6] = {1, 2, 3, 101, 102, 103};
  float m1[6] = {0, 0, 0, 0, 0, 0};
  float n[9] = {4, 5, 6, 104, 105, 106, 7, 8, 9};
  float n1[9] = {0, 0, 0, 0, 0, 0, 0, 0, 0};
  __enzyme_autodiff((void*)g, m, m1, n, n1);
}
```

enzyme/Enzyme/AdjointGenerator.h
wsmoses: Bumping this. Also its wise to avoid a lookup if it's not used to avoid unnecessary caching.

wsmoses (approved):
Excluding minor comments (see the lookup, and stride), LGTM.

PR #229, opened 2021-07-10
Introduce enzyme_iter for preserving approximation levels through differentiation.

wsmoses:
Additionally fix & simplify accumulation of structs.

As a side note, merging this means there's only one more "based off of LLVM mainline" branch with unmerged commits.


vchuravy (approved):
Should also be added to docs

PR #232, opened 2021-07-16
Implement FNeg for Forward Mode

### LINKED ISSUE #230: FNeg in ForwardMode
Hi,
the implementation of unary minus for ForwardMode seems to be broken.
I'm using the current brew version (LLVM 12) on Linux and opt produces a segfault.
I traced the error to
https://github.com/wsmoses/Enzyme/blob/9be50337dbf98a6a0489e94660d15586466a4b34/enzyme/Enzyme/AdjointGenerator.h#L252
where it tries to get a ReverseBuilder which I suppose doesn't exist in ForwardMode.
But I'm not sure how to procede from there.

Testcase to reproduce:
```
// clang++ fneg.cpp -S -emit-llvm -o input.ll -O2 -fno-vectorize -fno-slp-vectorize -fno-unroll-loops
// opt input.ll -load=/home/linuxbrew/.linuxbrew/lib/LLVMEnzyme-12.so -enzyme -o output.ll -S
// clang++ -c output.ll

extern double __enzyme_fwddiff(void*, double, double);

double foo(double x){
    return -x;
}

double dfoo(double x){
    return __enzyme_fwddiff((void*)foo, x, 1.0);
}
```

Best
Markus

tgymnich:
fixes #230

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -235,8 +235,6 @@ public:
   void visitInstruction(llvm::Instruction &inst) {
     // TODO explicitly handle all instructions rather than using the catch all
     // below
-    if (Mode == DerivativeMode::ReverseModePrimal)
-      return;
 
 #if LLVM_VERSION_MAJOR >= 10
     if (auto *FPMO = dyn_cast<FPMathOperator>(&inst)) {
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -248,16 +246,37 @@ class AdjointGenerator
         Value *orig_op1 = FPMO->getOperand(0);
```
wsmoses, line 246: This also appears to be missing an `eraseIfUnused(*FPMO);`, mind adding it here for correctness.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -248,16 +246,37 @@ class AdjointGenerator
...
         bool constantval1 = gutils->isConstantValue(orig_op1);
```
wsmoses, line 247: I'd change this to a return here if constant rather than having the repeated checks.

enzyme/test/Enzyme/ForwardMode/fneg.ll
```diff
@@ -0,0 +1,32 @@
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
```
wsmoses, line 1: I think you're going to need to set a check for llvm version above a certain number here.

wsmoses (approved):
LGTM, please address comments.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: wouldn't line 244/242 already take care of this? `eraseIfUnused(inst);`

enzyme/Enzyme/AdjointGenerator.h
tgymnich: done

enzyme/test/Enzyme/ForwardMode/fneg.ll
tgymnich: done

enzyme/Enzyme/AdjointGenerator.h
wsmoses: Yes it would! The diff cut it off and I was a bit concerned.

tgymnich: allright 😄

PR #265, opened 2021-08-01
Fix double free error

wsmoses:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4476,6 +4476,103 @@ class AdjointGenerator
         return;
       }
 
+      // Functions that only modify pointers and don't allocate memory,
+      // needs to be run on shadow in primal
+      if (funcName ==
+          "_ZSt29_Rb_tree_insert_and_rebalancebPSt18_Rb_tree_node_baseS0_RS_") {
```
vchuravy, line 4482: This seems rather unfortunate (and orthogonal to this PR)

Could we solve this by using a bit code version of libstdc++?

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -3311,6 +3311,12 @@ void TypeAnalyzer::visitCallInst(CallInst &call) {
       updateAnalysis(&call, TypeTree(BaseType::Integer).Only(-1), &call);
       return;
     }
+    if (funcName == "__dynamic_cast" ||
+        funcName == "_ZSt18_Rb_tree_decrementPSt18_Rb_tree_node_base" ||
```
vchuravy, line 3315: What happens if someone uses -nostdlib and provides their own implementation?

vchuravy (commented):
There seems to be some handling of libstdc++? Can we split that out?

PR #300, opened 2021-08-16
ForwardMode IntrinsicInst

tgymnich:
- [x] IntrinsicInst
- [x] Tests

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -2887,8 +2887,6 @@ public:
         return;
       }
 #endif
-
-      case Intrinsic::lifetime_start:
       case Intrinsic::nvvm_sqrt_rn_d:
       case Intrinsic::sqrt: {
         Value *op = diffe(orig_ops[0], Builder2);
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -2940,24 +2940,20 @@ public:
       case Intrinsic::x86_sse_max_ps:
 #endif
       case Intrinsic::maxnum: {
-        auto op0 = gutils->getNewFromOriginal(orig_ops[0]);
-        auto op1 = gutils->getNewFromOriginal(orig_ops[1]);
-        Value *cmp = Builder2.CreateFCmpOLT(op0, op1);
-
-        if (!gutils->isConstantValue(orig_ops[0]) &&
+        if (!gutils->isConstantValue(orig_ops[0]) ||
             !gutils->isConstantValue(orig_ops[1])) {
-          Value *dif = Builder2.CreateSelect(cmp, diffe(orig_ops[0], Builder2),
-                                             diffe(orig_ops[1], Builder2));
-          setDiffe(&I, dif, Builder2);
-        } else if (!gutils->isConstantValue(orig_ops[0])) {
-          Value *dif = Builder2.CreateSelect(
-              cmp, ConstantFP::get(orig_ops[0]->getType(), 0),
-              diffe(orig_ops[0], Builder2));
-          setDiffe(&I, dif, Builder2);
-        } else if (!gutils->isConstantValue(orig_ops[1])) {
-          Value *dif =
-              Builder2.CreateSelect(cmp, diffe(orig_ops[1], Builder2),
-                                    ConstantFP::get(orig_ops[0]->getType(), 0));
+          Value *op0 = gutils->getNewFromOriginal(orig_ops[0]);
+          Value *op1 = gutils->getNewFromOriginal(orig_ops[1]);
+          Value *cmp = Builder2.CreateFCmpOLT(op0, op1);
+
+          Value *diffe0 = gutils->isConstantValue(orig_ops[0])
+                              ? 0
+                              : diffe(orig_ops[0], Builder2);
+          Value *diffe1 = gutils->isConstantValue(orig_ops[1])
+                              ? 0
+                              : diffe(orig_ops[1], Builder2);
+
+          Value *dif = Builder2.CreateSelect(cmp, diffe0, diffe1);
           setDiffe(&I, dif, Builder2);
         }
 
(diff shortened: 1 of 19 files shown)


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,442 @@ class AdjointGenerator
...
+      case Intrinsic::nvvm_membar_gl:
+      case Intrinsic::nvvm_membar_sys:
+      case Intrinsic::assume:
+      case Intrinsic::prefetch:
+      case Intrinsic::dbg_declare:
+      case Intrinsic::dbg_value:
+#if LLVM_VERSION_MAJOR > 6
+      case Intrinsic::dbg_label:
+#endif
+      case Intrinsic::dbg_addr:
```
wsmoses, line 2851: I think this would be better handled by a check for inactive value and instruction rather than the cases here.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,442 @@ class AdjointGenerator
...
+          auto und = UndefValue::get(orig_ops[1]->getType());
+          auto mask = ConstantAggregateZero::get(VectorType::get(
+              Type::getInt32Ty(und->getContext()),
+#if LLVM_VERSION_MAJOR >= 11
+              cast<VectorType>(und->getType())->getElementCount()));
+#else
+              cast<VectorType>(und->getType())->getNumElements()));
+#endif
+          auto accdiff = diffe(orig_ops[0], Builder2);
+          auto diff = Builder2.CreateShuffleVector(
```
wsmoses, line 2881: This is a reverse mode derivative, in forward mode I think the correct answer is to sum the shadows.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,442 @@ class AdjointGenerator
...
+              Builder2.CreateInsertElement(und, accdiff, (uint64_t)0), und,
+              mask);
+          setDiffe(&I, diff, Builder2);
+        }
+        return;
+      }
+#endif
+
+      case Intrinsic::lifetime_start:
```
wsmoses, line 2890: Why is lifetime start here?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,442 @@ class AdjointGenerator
...
+      case Intrinsic::x86_sse_max_ss:
+      case Intrinsic::x86_sse_max_ps:
+#endif
+      case Intrinsic::maxnum: {
+        auto op0 = gutils->getNewFromOriginal(orig_ops[0]);
+        auto op1 = gutils->getNewFromOriginal(orig_ops[1]);
+        Value *cmp = Builder2.CreateFCmpOLT(op0, op1);
+
+        if (!gutils->isConstantValue(orig_ops[0]) &&
+            !gutils->isConstantValue(orig_ops[1])) {
```
wsmoses, line 2949: I'd change this to have a variable like this `diffe0 = isConstantValue(op[0]) ? 0 : diffe(op[0])`. It would simplify the case work here.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,442 @@ class AdjointGenerator
...
+#if LLVM_VERSION_MAJOR < 10
+      case Intrinsic::x86_sse_min_ss:
+      case Intrinsic::x86_sse_min_ps:
+#endif
+      case Intrinsic::minnum: {
+        auto op0 = gutils->getNewFromOriginal(orig_ops[0]);
+        auto op1 = gutils->getNewFromOriginal(orig_ops[1]);
+        Value *cmp = Builder2.CreateFCmpOLT(op0, op1);
+
+        if (!gutils->isConstantValue(orig_ops[0]) &&
```
wsmoses, line 2977: Same comment as max

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,442 @@ class AdjointGenerator
...
+
+      case Intrinsic::fma: {
+        Value *dif = nullptr;
+
+        auto op1 = gutils->getNewFromOriginal(orig_ops[1]);
+        auto op2 = gutils->getNewFromOriginal(orig_ops[2]);
+
+        if (!gutils->isConstantValue(orig_ops[0])) {
+          dif = diffe(orig_ops[0], Builder2);
+        }
```
wsmoses, line 3005: Might as well set this to be zero otherwise, and thus simplify the conditional operands below.

enzyme/Enzyme/AdjointGenerator.h
wsmoses: Sorry, to clarify the 0 in the above should be a constant, not a literal zero [meaning nullptr]

tgymnich: I have written way too much swift code where any object can be represented a custom literal. 😄

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,407 @@ class AdjointGenerator
...
+          auto mask = ConstantAggregateZero::get(VectorType::get(
+              Type::getInt32Ty(und->getContext()),
+#if LLVM_VERSION_MAJOR >= 11
+              cast<VectorType>(und->getType())->getElementCount()));
+#else
+              cast<VectorType>(und->getType())->getNumElements()));
+#endif
+
+          auto vecdif = diffe(orig_ops[1], Builder2);
+          auto dif1 = Builder2.CreateShuffleVector(und, vecdif, mask);
```
wsmoses, line 2861: I don't think is right, this should also be an reduce fadd of the diff, right?

wsmoses (approved):
Fix and add a test for the reduce fadd, but otherwise LGTM.

enzyme/test/Enzyme/ForwardMode/vector_reduce_fadd.ll
```diff
@@ -0,0 +1,45 @@
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
+
+define float @tester1(<4 x float> %input) {
+entry:
+  %unord = call reassoc float @llvm.vector.reduce.fadd.v4f32(float -0.0, <4 x float> %input)
```
wsmoses, line 5: Can you make the first arg a parameter of the caller rather than a constant?

wsmoses (approved):
Minor nits re reduce but otherwise LGTM.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2822,6 +2825,400 @@ class AdjointGenerator
...
+#endif
+      {
+        if (gutils->isConstantInstruction(&I))
+          return;
+
+        auto accdif = gutils->isConstantValue(orig_ops[0])
+                          ? ConstantFP::get(orig_ops[0]->getType(), 0)
+                          : diffe(orig_ops[0], Builder2);
+
+        if (!gutils->isConstantValue(orig_ops[1])) {
```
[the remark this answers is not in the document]
tgymnich: `@llvm.vector.reduce.fadd.v4f32(float %"acc'", <4 x float> %"vec'")`

Wouldn't calling `diffe(orig_ops[1], Builder2);` on a constant print an error?

enzyme/test/Enzyme/ForwardMode/vector_reduce_fadd.ll
tgymnich: done

enzyme/Enzyme/AdjointGenerator.h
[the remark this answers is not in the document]
wsmoses: Yes you're correct. I think the correct behavior here is a constant 0 if the arg is inactive, for both the initial value and the vector arg. You can do the same conditional as in previous ones.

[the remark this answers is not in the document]
tgymnich: done

PR #308, opened 2021-08-22
Calculate adjoint of cblas_dgemm

reikdas:


reikdas:
@wsmoses The CI failures should be unrelated.

wsmoses:
Yup, those CI failures are needing to get https://reviews.llvm.org/D108221 landed.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -5980,6 +5980,456 @@ class AdjointGenerator
...
+      in_arg++;
+      Argument *xfuncarg = in_arg;
+      in_arg++;
+      in_arg++;
+      Argument *yfuncarg = in_arg;
+      bool xcache = !gutils->isConstantValue(call.getArgOperand(3)) &&
+                    uncacheable_args.find(xfuncarg)->second;
+      bool ycache = !gutils->isConstantValue(call.getArgOperand(1)) &&
+                    uncacheable_args.find(yfuncarg)->second;
+      if ((Mode == DerivativeMode::ReverseModeCombined ||
```
wsmoses, line 6016: Can you add asserts that the forward modes are unhandled

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -5980,6 +5980,456 @@ class AdjointGenerator
...
+          else if (Bcache)
+            structarg2 = cacheval;
+        }
+        if (!Acache)
+          structarg1 = lookup(gutils->getNewFromOriginal(call.getArgOperand(7)),
+                              Builder2);
+        if (!Bcache)
+          structarg2 = lookup(gutils->getNewFromOriginal(call.getArgOperand(9)),
+                              Builder2);
+        if (cast<ConstantInt>(call.getArgOperand(0))->getValue() == 102) {
```
wsmoses, line 6284: Can you compare with a character rather than a raw number? Also usually don't both upper and lowercase work?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -5980,6 +5980,456 @@ class AdjointGenerator
...
+                           Builder2);
+            sbldc = lookup(gutils->getNewFromOriginal(call.getArgOperand(4)),
+                           Builder2);
+          }
+        } else
+          assert(false && "Wrong value");
+        CallInst *safunccall, *sbfunccall;
+        if (aactive) {
+          if (cast<ConstantInt>(call.getArgOperand(2))->getValue() == 112 ||
+              cast<ConstantInt>(call.getArgOperand(2))->getValue() == 113) {
```
wsmoses, line 6316: Likewise character for the rest here.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -5980,6 +5980,456 @@ class AdjointGenerator
...
+              sblda,
+              gutils->invertPointerM(call.getArgOperand(12), Builder2),
+              sbldb,
+              doneval,
+              gutils->invertPointerM(call.getArgOperand(9), Builder2),
+              sbldc};
+          sbfunccall = Builder2.CreateCall(dfunc, sbfuncargs);
+        }
+        auto scfunc = gutils->oldFunc->getParent()->getOrInsertFunction(
+            scfuncname, Builder2.getVoidTy(), Builder2.getInt32Ty(),
```
wsmoses, line 6397: Are we always sure this is an int32? Can we instead use whatever the type of the constantint is (if, for some reason, it is say an int64, etc).

enzyme/test/Enzyme/ReverseMode/blas/cblas_dgemm_col_mod1_split.ll
```diff
@@ -0,0 +1,142 @@
...
+  %arrayidx = getelementptr inbounds double, double* %B, i64 0
+  store double 1.000000e+01, double* %arrayidx, align 8
+  ret void
+}
+
+define dso_local i32 @main() {
+entry:
+  %A = alloca [8 x double], align 16
+  %B = alloca [6 x double], align 16
+  %C = alloca [12 x double], align 16
```
wsmoses, line 58: Can you remove the main except for the call to autodiff (and perhaps just have arguments forwarded to the autodiff call)

enzyme/test/Enzyme/ReverseMode/blas/cblas_dgemm_col_mod1_split.ll
```diff
@@ -0,0 +1,142 @@
...
+;CHECK-NEXT:  %1 = extractvalue { double*, double* } %0, 0
+;CHECK-NEXT:  %2 = extractvalue { double*, double* } %0, 1
+;CHECK-NEXT:  call void @cblas_dgemm(i32 102, i32 111, i32 112, i32 4, i32 2, i32 3, double %alpha, double* nocapture readonly %"Cfoo'", i32 4, double* nocapture readonly %2, i32 3, double 1.000000e+00, double* %"Afoo'", i32 4)
+;CHECK-NEXT:  %3 = bitcast double* %1 to i8*
+;CHECK-NEXT:  tail call void @free(i8* %3)
+;CHECK-NEXT:  call void @cblas_dgemm(i32 102, i32 112, i32 111, i32 2, i32 3, i32 4, double %alpha, double* nocapture readonly %1, i32 4, double* nocapture readonly %"Cfoo'", i32 4, double 1.000000e+00, double* %"Bfoo'", i32 2)
+;CHECK-NEXT:  %4 = bitcast double* %2 to i8*
+;CHECK-NEXT:  tail call void @free(i8* %4)
+;CHECK-NEXT:  call void @cblas_dscal(i32 12, double %beta, double* %"Cfoo'", i32 1)
+;CHECK-NEXT:  ret { double, double } zeroinitializer
```
wsmoses, line 139: Why is this zero? Shouldn't we also consider the case where alpha and beta are active?

enzyme/Enzyme/AdjointGenerator.h
reikdas: The first three arguments are enums. I am not sure about chars that we can pass as arguments instead. Do you have an example?

Do you instead mean passing in the enum names like `CblasRowMajor`, `CblasNoTrans`, etc instead of the number?

PR #320, opened 2021-09-01
EnzymeLogic cleanup

tgymnich:
Split up CreatePrimalAndGradient and remove the fwd mode stuff

### DIFF
--- a/enzyme/Enzyme/CApi.cpp
+++ b/enzyme/Enzyme/CApi.cpp
@@ -323,15 +323,13 @@ LLVMBasicBlockRef EnzymeGradientUtilsAllocationBlock(GradientUtils *gutils) {
   return wrap(gutils->inversionAllocs);
 }
 
-LLVMValueRef EnzymeCreateDual(EnzymeLogicRef Logic, LLVMValueRef todiff,
-                              CDIFFE_TYPE retType, CDIFFE_TYPE *constant_args,
-                              size_t constant_args_size,
-                              EnzymeTypeAnalysisRef TA, uint8_t returnValue,
-                              uint8_t dretUsed, CDerivativeMode mode,
-                              LLVMTypeRef additionalArg, CFnTypeInfo typeInfo,
-                              uint8_t *_uncacheable_args,
-                              size_t uncacheable_args_size, uint8_t AtomicAdd,
-                              uint8_t PostOpt) {
+LLVMValueRef EnzymeCreateForwardDiff(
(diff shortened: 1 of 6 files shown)


enzyme/Enzyme/CApi.cpp
```diff
@@ -323,6 +323,31 @@ LLVMBasicBlockRef EnzymeGradientUtilsAllocationBlock(GradientUtils *gutils) {
   return wrap(gutils->inversionAllocs);
 }
 
+LLVMValueRef EnzymeCreateDual(EnzymeLogicRef Logic, LLVMValueRef todiff,
```
wsmoses, line 326: Perhaps CreateForwardDiff ?  Also you probably don't need to pass a derivative mode here (unless we want to leave an option to expand a new vector forward mode).

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -3390,6 +3389,349 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
...
+    }
+  }
+
+  for (Argument &Arg : gutils->newFunc->args()) {
+    if (Arg.hasAttribute(Attribute::Returned))
+      Arg.removeAttr(Attribute::Returned);
+    if (Arg.hasAttribute(Attribute::StructRet))
+      Arg.removeAttr(Attribute::StructRet);
+  }
+  if (gutils->newFunc->hasFnAttribute(Attribute::OptimizeNone))
```
wsmoses, line 3672: While we're cleaning up, can we move the function attribute code into a common function used by all methods.

enzyme/Enzyme/CApi.cpp
tgymnich: done

enzyme/Enzyme/EnzymeLogic.cpp
tgymnich: The function attribute code for CreateAugmentedPrimal is not quite the same, so ill leave it out.

wsmoses: How precisely do they differ?

tgymnich:
Let's remove CacheAnalysis and others in a future PR. We should should probably also add mode assertions to `legalRecompute`

wsmoses:
Should definitely still remove those mincut / etc parts, but can be subsequent PR.

wsmoses (approved):

PR #329, opened 2021-09-20
Use custom reverse in combined mode

wsmoses:
Resolves https://github.com/wsmoses/Enzyme/issues/327


tgymnich (approved):
LGTM, maybe a unit test would also be good to add.

```
define internal {} @fixgradient_square_(double* %0, double* %1, double* %2, double* %3) {
entry:
  %4 = call i8* @augment_square_(double* %0, double* %1, double* %2, double* %3)
  call void @gradient_square_(double* %0, double* %1, double* %2, double* %3, i8* %4)
  ret {} undef
}
```

Also the return type could be cleaned up in a future PR

PR #334, opened 2021-09-22
Fix Forward Mode Return Type

tgymnich:
Make forward mode return a struct only when needed

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -7604,10 +7604,14 @@ public:
 #endif
 
       if (!newcalled->getReturnType()->isVoidTy()) {
-        unsigned structidx =
-            retUsed && subretType != DIFFE_TYPE::CONSTANT ? 1 : 0;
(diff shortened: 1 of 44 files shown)


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -7604,10 +7604,15 @@ class AdjointGenerator
 #endif
 
       if (!newcalled->getReturnType()->isVoidTy()) {
-        unsigned structidx =
-            retUsed && subretType != DIFFE_TYPE::CONSTANT ? 1 : 0;
+        bool structret =
+            retUsed && subretType != DIFFE_TYPE::CONSTANT ? true : false;
```
wsmoses, line 7608: This need not be a ternary, right?

wsmoses (approved):

enzyme/Enzyme/AdjointGenerator.h
tgymnich: 😄 yes

PR #338, opened 2021-09-26
Handle StructReturn

tgymnich:


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -130,14 +139,45 @@ class Enzyme : public ModulePass {
...
-    SmallVector<Value *, 2> args;
+    bool fnsret = false;
+    if (cast<Function>(fn)->hasParamAttribute(0, Attribute::StructRet)) {
+      Type *fnsrety = cast<PointerType>(FT->getParamType(0));
+      Type *cisrety = cast<PointerType>(CI->getArgOperand(0)->getType());
 
-    unsigned truei = 0;
-    IRBuilder<> Builder(CI);
+      if (fnsret == cisret) {
+        llvm::errs() << "Struct return types of __ezyme_autodiff and function "
```
wsmoses, line 150: Enzyme is misspelled here.

tgymnich: fixed

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -130,14 +139,45 @@ class Enzyme : public ModulePass {
...
-    std::vector<DIFFE_TYPE> constants;
-    SmallVector<Value *, 2> args;
+    bool fnsret = false;
+    if (cast<Function>(fn)->hasParamAttribute(0, Attribute::StructRet)) {
+      Type *fnsrety = cast<PointerType>(FT->getParamType(0));
+      Type *cisrety = cast<PointerType>(CI->getArgOperand(0)->getType());
 
-    unsigned truei = 0;
-    IRBuilder<> Builder(CI);
+      if (fnsret == cisret) {
```
wsmoses, line 149: I might remove fnsret here since its always false. I might in fact remove the variable entirely and instead do truei++ here.

tgymnich: Fixed it. It was supposed to be `(fnsrety != cisrety)`. I also removed `fnsret`

enzyme/test/Enzyme/ForwardMode/sret.ll
```diff
@@ -0,0 +1,106 @@
...
+attributes #4 = { nounwind "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "less-precise-fpmad"="false" "no-frame-pointer-elim"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="false" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+fxsr,+mmx,+sse,+sse2,+x87" "unsafe-fp-math"="false" "use-soft-float"="false" }
+attributes #5 = { argmemonly nounwind }
+attributes #6 = { nounwind }
+
+
+; CHECK: define dso_local void @_Z7dsquared(%"struct.std::array"* noalias sret %agg.result, double %x)
+; CHECK-NEXT: entry:  
+; CHECK-NEXT:   %0 = alloca %"struct.std::array"
+; CHECK-NEXT:   call void @fwddiffe_Z6squared(%"struct.std::array"* %0, %"struct.std::array"* %agg.result, double %x, double 1.000000e+00)
```
wsmoses, line 79: I feel like rather than adding the extra argument to be differentiated, the better solution would be to skip the sret argument, then after doing the regular enzyme call which returns a tuple/structtype, we'd set the structreturn with that result.

tgymnich: The regular enzyme call is getting a function with this signature though: `void (%"struct.std::__1::array"*, double)`. I don't think it would be wise to rewrite the square function clang generated, in order for us to perform the regular enzyme call.

wsmoses: I see, rephrasing the above a bit -- in the case where the sret is only on enzyme_autodiff/fwddiff and not the actual function being differentiated, I don't think we should add the extra argument. However if sret is on the function being differentiated, I agree with this. Can you add a test for the other case as well?

PR #343, opened 2021-09-30
Fix malloc placeholder

tgymnich:


wsmoses (approved):
Add a test otherwise lgtm

tgymnich:
> Add a test otherwise lgtm

I still need to do something like #251

wsmoses:
I don't think you do? Reverse needs to keep around allocations from fwd to reverse, but forward mode should be able to get away with doing the same allocation/deallocation as the original program

tgymnich:
I just want to handle `enzyme_fromstack`

wsmoses:
Oh, `enzyme_fromstack` shouldn't be enabled for forward mode

tgymnich:
It is

wsmoses:
We'll there's the solution to the problem!

PR #346, opened 2021-10-02
Add intrinsic for determining necessary allocation size

wsmoses:


enzyme/test/Enzyme/ReverseMode/splitSize.ll
```diff
@@ -0,0 +1,38 @@
...
+  ret double %res
+}
+
+define void @test_derivative(double* %x, double* %dx) {
+entry:
+  %size = call i64 (double (double*)*, ...) @__enzyme_augmentsize(double (double*)* nonnull @tester, metadata !"enzyme_dup")
+  %cache = alloca i8, i64 %size, align 1
+  %0 = tail call { double } (double (double*)*, ...) @__enzyme_augmentfwd(double (double*)* nonnull @tester, metadata !"enzyme_allocated", i64 %size, metadata !"enzyme_tape", i8* %cache, double* %x, double* %dx)
+  tail call void (double (double*)*, ...) @__enzyme_reverse(double (double*)* nonnull @tester, metadata !"enzyme_allocated", i64 %size, metadata !"enzyme_tape", i8* %cache, double* %x, double* %dx)
+  ret void
```
wsmoses, line 19: @jedbrown @LeilaGhaffari

Obviously this is at the llvm level and using split reverse mode rather than forward mode, but this is an example of the automatic tape size magic intrinsic I was thinking of from our call.

vchuravy (approved):

enzyme/test/Enzyme/ReverseMode/splitSize.ll
jedbrown: Cool, Is the `metadata !"enzyme_dup"` where the sizes would go? So if `@tester` accepted an array of caller-specified size, those args would still be passed to `__enzyme_augmentsize`?

wsmoses: Not quite, so the enzyme_dup denotes that we want a gradient whose corresponding argument is active, and a pointer (see here for a more in depth description: https://enzyme.mit.edu/getting_started/CallingConvention/#types).

In this case we'd be using `__enzyme_augmentsize` to determine the size of memory we need to reserve for the tape, which we use as an argument to the stack allocation (alloca instruction). We then specify to `__enzyme_augmentfwd` and `__enzyme_reverse` how much memory we have allocated by the argument after the `!"enzyme_allocated"` metadata. In this case we passed `%size` the result of `__enzyme_augmentsize`, or precisely the amount of memory Enzyme requires. We could alternatively provide any amount of memory which is at least as large as `__enzyme_augmentsize` and succeed, otherwise Enzyme will throw a compile-time error.

So in essence what this PR does is:
* add functionality through a runtime function `__enzyme_augmentsize` to query the amount of memory required for the outermost level of Enzyme's cache.
* allow passing in that memory as an argument (the argument following `!"enzyme_tape"`), rather than having an extra-wide register returned and having the caller copy that into the tape.

This doesn't yet do the more advanced array of pointers caching we discussed on the call, but is a first step on that path.

PR #347, opened 2021-10-03
Custom Forward Derivative

tgymnich:


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -79,6 +79,263 @@ llvm::cl::opt<bool>
 
 namespace {
 
+static void
+handleCustomGradient(llvm::Module &M, llvm::GlobalVariable &g,
+                     std::vector<GlobalVariable *> &globalsToErase) {
```
wsmoses, line 84: Can you simplify this and the following methods a bit -- there seems to be a fair amount of common code.

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -79,6 +79,220 @@ llvm::cl::opt<bool>
...
 
+static void
+handleCustomDerivative(StringRef handlername, int numargs, llvm::Module &M,
+                       llvm::GlobalVariable &g,
+                       std::vector<GlobalVariable *> &globalsToErase) {
+  if (g.hasInitializer()) {
+    if (auto CA = dyn_cast<ConstantAggregate>(g.getInitializer())) {
+      if (CA->getNumOperands() != numargs) {
+        llvm::errs() << M << "\n";
+        llvm::errs() << "Use of " << handlername
```
wsmoses, line 90: This is a good start, though I still think making (maybe a template on the number of args) that extracts out the initializer/constantexpr code from this function and below would be really useful.

wsmoses (approved):

PR #348, opened 2021-10-03
More BLAS functions using BCPass

reikdas:
Changes have been sitting in a branch for a while.


wsmoses (approved):

vchuravy (changes requested):
The `bc` files are problemtatic. They are not actually stable across LLVM versions, and we will need a clear way of versioning and recreating them.

reikdas:
> The `bc` files are problemtatic. They are not actually stable across LLVM versions, and we will need a clear way of versioning and recreating them.

How do you think it should look?
I can think of two things - 
1. Add different bitcode files for each LLVM version that we support.
2. Just have the C file checked into the repository and let the user/the tests generate the bitcode files themselves.

vchuravy:
If the build system can autogenerate the bitcode files from C/C++ that would be ideal

PR #364, opened 2021-10-25
Add cblas level 1 routines and corresponding regression tests

tansongchen:
I'm currently looking on LIT's documentation to learn how to write more comprehensive test, but let's just first put it here.


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -6308,6 +6308,43 @@ class AdjointGenerator
...
+        innerType = Type::getFloatTy(call.getContext());
+        dfuncName = "cblas_sscal";
+      } else {
+        assert(false && "Unreachable");
+      }
+      if (Mode == DerivativeMode::ReverseModeCombined ||
+          Mode == DerivativeMode::ReverseModeGradient) {
+        IRBuilder<> Builder2(call.getParent());
+        getReverseBuilder(Builder2);
+        auto derivcall = gutils->oldFunc->getParent()->getOrInsertFunction(
```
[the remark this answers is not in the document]
tansongchen: I am not sure about the derivative but I derived it in the following way: suppose the output is z, x'[i] = a * x[i], so ∂z/∂x[i] = ∂z/∂x'[i] * ∂x'[i]/∂x[i] = a * ∂z/∂x'[i]

tansongchen:
> > The test passed for higher versions of LLVM but not for lower versions, and I think this is because I generated this .ll file with LLVM 13. @wsmoses Could you maybe tell a little bit on how to generate .ll files that are compatible with lower versions?
> 
> The error is that in lower versions of llvm arguments need actual names, not just %0. Add names to the functions and it should pass?

Thanks for the hints. By the way, I am curious about what the life cycle for llvm versions is

enzyme/test/Enzyme/ReverseMode/blas/cblas_dscal_nomod.ll
```diff
@@ -0,0 +1,67 @@
...
+@__const.main.m = private unnamed_addr constant [3 x double] [double 1.000000e+00, double 2.000000e+00, double 3.000000e+00], align 16
+@__const.main.d_m = private unnamed_addr constant [3 x double] [double 1.000000e+00, double 1.000000e+00, double 1.000000e+00], align 16
+@.str = private unnamed_addr constant [10 x i8] c"%f %f %f\0A\00", align 1
+
+; Function Attrs: noinline nounwind optnone ssp uwtable
+define void @f(double* noalias %0) #0 {
+  %2 = alloca double*, align 8
+  store double* %0, double** %2, align 8
+  %3 = load double*, double** %2, align 8
+  call void @cblas_dscal(i32 3, double 2.000000e+00, double* %3, i32 1)
```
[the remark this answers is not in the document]
tansongchen: Sure, this can be simplified (not sure why llvm didn't do that), and I agree that the double (scaling factor a) should be passed to f (so we can probably calculate the derivative of it), but what is the point of passing other integers (size and stride) to f?

tansongchen:
I uploaded the implementation of level 1 routines and level 2 gemv, but the test system is not working on my computer, I will fix it and add test cases for gemv later

tansongchen:
Also I noticed that there are some related implementation of gemm at #308 , but both of us are doing this in a bit awkward way. After testing gemv I will try to extract the common portion of many blas routines and do a refactoring

tansongchen:
@wsmoses All tests have passed. Two fails are due to other test cases

wsmoses (changes requested):

enzyme/test/Enzyme/ReverseMode/blas/cblas_daxpy_a.ll
```diff
@@ -0,0 +1,55 @@
...
+  tail call void @cblas_daxpy(i32 %n, double %a, double* %x, i32 1, double* %y, i32 1)
+  ret void
+}
+declare void @cblas_daxpy(i32, double, double*, i32, double*, i32)
+define i32 @main() {
+entry:
+  %m = alloca [3 x double], align 16
+  %d_m = alloca [3 x double], align 16
+  %n = alloca [3 x double], align 16
+  %d_n = alloca [3 x double], align 16
```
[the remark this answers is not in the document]
tansongchen: Do you mean removing the `llvm.lifetime.start` and `llvm.lifetime.end` calls? The `.c` file only contains variable definition and a call to `__enzyme_autodiff` and these calls are generated by clang. Is there a way to strip it when generating LLVM IR?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4580,6 +4686,547 @@ class AdjointGenerator
...
+        innerType = Type::getFloatTy(call.getContext());
+        dfuncName = "cblas_saxpy";
+      } else {
+        assert(false && "Unreachable");
+      }
+      if (Mode == DerivativeMode::ReverseModeCombined ||
+          Mode == DerivativeMode::ReverseModeGradient) {
+        IRBuilder<> Builder2(call.getParent());
+        getReverseBuilder(Builder2);
+        auto arg_n = call.getArgOperand(0), arg_x = call.getArgOperand(1),
```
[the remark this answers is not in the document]
tansongchen: I think we can also handle the case that one of them is not active. For example if x is active and y is a constant array, we swap x and y at some point; the adjoint of swap should cause x to be zeroed out

enzyme/test/Enzyme/ReverseMode/blas/cblas_daxpy_a.ll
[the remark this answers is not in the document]
tansongchen: I actually used O2 for this, but you are absolutely right about removing the main! We don't need "real" data here and it can greatly simplify our test cases

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4580,6 +4686,547 @@ class AdjointGenerator
...
+
+      if (Mode == DerivativeMode::ReverseModeGradient) {
+        eraseIfUnused(call, /*erase*/ true, /*check*/ false);
+      } else {
+        eraseIfUnused(call);
+      }
+      return true;
+    }
+
+    if ((funcName == "cblas_sgemv" || funcName == "cblas_dgemv") &&
```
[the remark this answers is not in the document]
tansongchen: Sure, I will update this pr to include level 1 functions only

enzyme/Enzyme/AdjointGenerator.h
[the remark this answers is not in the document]
tansongchen: OK I will use assert now but later I will implement a strided memset to handle the case where x is active and y is not

vchuravy:
@tansongchen you marked all the conversations as resolved, but there is no new commit yet? Maybe you forgotten to push?

tansongchen:
> @tansongchen you marked all the conversations as resolved, but there is no new commit yet? Maybe you forgotten to push?

Just pushed

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4506,12 +4522,23 @@ class AdjointGenerator
 
   bool handleBLAS(CallInst &call, Function *called, StringRef funcName,
                   const std::map<Argument *, bool> &uncacheable_args) {
+    if (!called->isDeclaration()) {
```
wsmoses, line 4525: I would still do the high-level blas handler even if the code has a definition

enzyme/test/Enzyme/ReverseMode/blas/cblas_daxpy_a.ll
```diff
@@ -0,0 +1,26 @@
+;RUN: %opt < %s %loadEnzyme -enzyme -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
+
+define void @wrapper(i32 %n, double %a, double* %x, double* %y) {
+entry:
+  tail call void @cblas_daxpy(i32 %n, double %a, double* %x, i32 1, double* %y, i32 1)
+  ret void
```
wsmoses, line 6: Can you pass in the incx and incy as variables?

enzyme/test/Enzyme/ReverseMode/blas/cblas_daxpy_a.ll
```diff
@@ -0,0 +1,26 @@
...
+}
```
wsmoses, line 7: Can you make a test where x and/or y are overwritten after the blas call (to check caching works)

enzyme/test/Enzyme/ReverseMode/blas/cblas_dnrm2.ll
```diff
@@ -0,0 +1,29 @@
+;RUN: %opt < %s %loadEnzyme -enzyme -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
+
+define double @wrapper(double* %x) {
+entry:
+  %call = tail call double @cblas_dnrm2(i32 3, double* %x, i32 1)
```
wsmoses, line 5: Can you pass n and inc as variables?

enzyme/test/Enzyme/ReverseMode/blas/cblas_dnrm2.ll
```diff
@@ -0,0 +1,29 @@
...
+  %mul = fmul double %call, %call
+  ret double %mul
```
wsmoses, line 7: Remove the square to simplify the test

enzyme/test/Enzyme/ReverseMode/blas/cblas_dnrm2.ll
```diff
@@ -0,0 +1,29 @@
...
+
+declare double @__enzyme_autodiff(i8*, double*, double*)
+
+;CHECK:define internal void @diffewrapper(double* %x, double* %"x'", double %differeturn) {
+;CHECK-NEXT:entry:
+;CHECK-NEXT:  %call = tail call double @cblas_dnrm2(i32 3, double* %x, i32 1)
+;CHECK-NEXT:  %m0diffecall = fmul fast double %differeturn, %call
+;CHECK-NEXT:  %m1diffecall = fmul fast double %differeturn, %call
+;CHECK-NEXT:  %0 = fadd fast double %m0diffecall, %m1diffecall
+;CHECK-NEXT:  %1 = fdiv fast double %0, %call
```
wsmoses, line 26: Also add a caching test here.

enzyme/test/Enzyme/ReverseMode/blas/cblas_dscal_a.ll
```diff
@@ -0,0 +1,26 @@
+;RUN: %opt < %s %loadEnzyme -enzyme -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
+
+define void @wrapper(double %a, double* %x) {
+entry:
+  tail call void @cblas_dscal(i32 3, double %a, double* %x, i32 1)
+  ret void
```
wsmoses, line 6: Pass as size/inc as args.

enzyme/test/Enzyme/ReverseMode/blas/cblas_dscal_a.ll
```diff
@@ -0,0 +1,26 @@
...
+}
```
wsmoses, line 7: Add a cache test as well.

enzyme/test/Enzyme/ReverseMode/blas/cblas_dswap.ll
```diff
@@ -0,0 +1,24 @@
+;RUN: %opt < %s %loadEnzyme -enzyme -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
+
+define void @wrapper(i32 %n, double* %x, double* %y) {
+entry:
+  tail call void @cblas_dswap(i32 %n, double* %x, i32 1, double* %y, i32 1)
+  ret void
```
wsmoses, line 6: Add a test that overwriting x/y doesn't cause a cache.

enzyme/test/Enzyme/ReverseMode/blas/cblas_daxpy_a.ll
tansongchen: Did you mean defining incx, incy as local variables in `wrapper()`? Why we need to do that?

tansongchen: or you mean adding incx and incy to the arguments of `wrapper()`? In this case it will be the same as `cblas_daxpy` and we probably don't need a wrapper at all

tansongchen: I am a little bit confused because in the previous `ddot` test cases the wrapper function accepts pointers only, and integers are also passed as literals so I assume that was fine

wsmoses: Yes (and it would be wise to update the ddot test as well, though that doesn't need to be this PR).

tansongchen: Oh you are right making them variables indeed help us find more bugs (the variables need to be `getNewFromOriginal`'d before using malloc/memcpy)

wsmoses:
This will likely need to take into account operand bundles as well once that lands, see https://github.com/wsmoses/Enzyme/pull/491

enzyme/test/Enzyme/ReverseMode/blas/cblas_daxpy_a.ll
wsmoses: I would also update the other blas tests as well if you have a chance to check for the same issue.

tansongchen:
@wsmoses I am looking at the operand bundles, and I am wondering why the ValueType for x is Shadow at https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/Enzyme/AdjointGenerator.h#L4777 ? I assume that since `memcpy` is using the x value this ValueType should be Primal.

wsmoses:
So the reason is that we’re copying into the shadow

tansongchen:
I see, I thought Shadow was only used for derivative values... Maybe we should update the description at https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/Enzyme/Utils.h#L258 later

tansongchen:
Operand bundles are added, and test cases are fixed according to the new convention.

tansongchen:
@wsmoses Any comments on the latest push?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4725,52 +5003,49 @@ class AdjointGenerator
...
-      }
-      Type *castvals[2] = {call.getArgOperand(1)->getType(),
-                           call.getArgOperand(3)->getType()};
+    auto module = gutils->oldFunc->getParent();
+    auto &context = call.getContext();
+    Type *voidType = Type::getVoidTy(context);
+    Type *scalarType = getBLASInnerType(funcName, context);
+    bool handled = false;
+
+    if (funcName == "cblas_ddot" || funcName == "cblas_sdot") {
```
wsmoses, line 5012: This should still only apply if this is a declaration.

tansongchen: But you have mentioned last week we should still do the high-level blas handler even if the code has a definition? What is the criteria for handling or not handling here?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4715,7 +4721,279 @@ class AdjointGenerator
...
+
+  enum BLASType { Void, Integer, Real, RealPointer };
+  struct BLASArgument {
+    std::string name;
+    BLASType type;
+  };
+  enum GradientOpType { BLASCall, HelperCall, AddToDiffe };
+  struct GradientOp {
+    GradientOpType gradientOpType;
+    std::string callee;
```
wsmoses, line 4761: Can you describe these struct class members and their intended usage?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4715,7 +4721,279 @@ class AdjointGenerator
...
+    BLASType returnType;
+    std::vector<std::string> args;
+  };
+  struct ChainRule {
+    std::vector<GradientOp> gradientOps;
+  };
+  struct BLASAdjointGenerator {
+    std::string name;
+    BLASType returnType;
```
wsmoses, line 4770: There's so much going on here that I'm inclined to put this in a separate BLAS file

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4715,7 +4721,279 @@ class AdjointGenerator
...
+    auto currentModule = gutils->oldFunc->getParent();
+    auto calledFunc = call.getCalledFunction();
+    char scalarTypeName = funcName.str()[6];
+    std::map<char, Type *> scalarTypeLookup = {
+        {'s', Type::getFloatTy(call.getContext())},
+        {'d', Type::getDoubleTy(call.getContext())}};
+    Type *scalarType = scalarTypeLookup.at(scalarTypeName);
+    auto size = ConstantExpr::getSizeOf(scalarType);
+    std::vector<Value *> args(g.signature.size());
+    std::vector<Type *> types(g.signature.size());
```
wsmoses, line 4792: What is the difference between actives and activities?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4715,7 +4721,279 @@ class AdjointGenerator
...
+             ++cacheRank) {
+          unsigned index = cacheArgIndices[cacheRank];
+          auto toCopy = gutils->getNewFromOriginal(args[index]);
+          auto malloc =
+              CallInst::CreateMalloc(&newCall, size->getType(), scalarType,
+                                     size, args[sz[index]], nullptr, "");
+          auto bitcast = BuilderZ.CreateBitCast(malloc, toCopy->getType());
+          auto stridedMemcpy = getOrInsertMemcpyStrided(
+              *currentModule, PointerType::getUnqual(scalarType), 0, 0);
+          BuilderZ.CreateCall(stridedMemcpy, {bitcast, toCopy, args[sz[index]],
```
wsmoses, line 4859: I'm not sure this is generalizable like this when we get to more advanced data layouts like matricies.

tansongchen: I took a look at the GEMV's description and I found that in `lda != m` case, this approach will result in copying `lda × n` many elements instead of `m × n` (which is less efficient). So you are right! I need to implement a new memcpy function to handle submatrix copying

tansongchen:
I haven't quite understood what is happening with the Julia nightly CI. It seems that other PRs are also failing on this?

wsmoses:
Yeah feel free to ignore Julia nightly CI.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4836,78 +4869,55 @@ class AdjointGenerator
...
-              gutils->getInvertedBundles(
-                  &call,
-                  {ValueType::None, ValueType::Shadow, ValueType::None,
-                   ycache ? ValueType::None : ValueType::Primal,
-                   ValueType::None},
-                  Builder2, /*lookup*/ true));
+          Value *args[6] = {new_n,    diff_ret, new_y,
+                            new_incy, diff_x,   diff_incx};
+          ValueType types[5] = {
+              ValueType::None, ValueType::Shadow, ValueType::None,
```
wsmoses, line 4908: Can you hold off on the ddot changes for now?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4919,6 +4929,416 @@ class AdjointGenerator
...
+      if (Mode == DerivativeMode::ReverseModeGradient) {
+        eraseIfUnused(call, /*erase*/ true, /*check*/ false);
+      } else {
+        eraseIfUnused(call);
+      }
+    }
+
+    if ((funcName == "cblas_sswap" || funcName == "cblas_dswap") &&
+        called->isDeclaration()) {
+      // sswap(int n, float *x, int incx, float *y, int incy)
```
wsmoses, line 4941: You may also need to cache n, incx, incy here if they are passed by reference as demonstrated in https://github.com/EnzymeAD/Enzyme/pull/528

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4919,6 +4929,416 @@ class AdjointGenerator
...
+           arg_incy = call.getArgOperand(4);
+      auto has_diff_x = !gutils->isConstantValue(arg_x),
+           has_diff_y = !gutils->isConstantValue(arg_y);
+      if (Mode == DerivativeMode::ReverseModeCombined ||
+          Mode == DerivativeMode::ReverseModeGradient) {
+        IRBuilder<> Builder2(call.getParent());
+        getReverseBuilder(Builder2);
+        if (has_diff_x) {
+#if LLVM_VERSION_MAJOR >= 11
+          auto swapCall =
```
wsmoses, line 4954: Can you add to differentialuseanalysis.h a corresponding note that neither primal x, nor primal y are needed for the reverse pass here.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4919,6 +4929,416 @@ class AdjointGenerator
...
+      auto arg_n = call.getArgOperand(0), arg_x = call.getArgOperand(1),
+           arg_incx = call.getArgOperand(2), arg_y = call.getArgOperand(3),
+           arg_incy = call.getArgOperand(4);
+      auto has_diff_x = !gutils->isConstantValue(arg_x),
+           has_diff_y = !gutils->isConstantValue(arg_y);
+      if (Mode == DerivativeMode::ReverseModeCombined ||
+          Mode == DerivativeMode::ReverseModeGradient) {
+        IRBuilder<> Builder2(call.getParent());
+        getReverseBuilder(Builder2);
+        if (has_diff_x) {
```
wsmoses, line 4952: Do you not want to check both has_diff_x and has_diff_y?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4919,6 +4929,416 @@ class AdjointGenerator
...
+                new_n,  ConstantFP::get(scalarType, 1.0),
+                diff_y, new_incy,
+                diff_x, new_incx};
+            Builder2.CreateCall(
+                axpyCall, args,
+                gutils->getInvertedBundles(&call,
+                                           {ValueType::None, ValueType::Shadow,
+                                            ValueType::None, ValueType::Shadow,
+                                            ValueType::None},
+                                           Builder2, true));
```
wsmoses, line 5019: You should handle the case where x or y are inactive

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4704,7 +4710,36 @@ class AdjointGenerator
     }
   }
 
-  bool handleBLAS(llvm::CallInst &call, Function *called, StringRef funcName,
+  std::string getBLASName(StringRef funcName, const std::string &deriv) {
+    return funcName.str().substr(0, 7) + deriv;
+  }
+
+  Type *getBLASInnerType(StringRef funcName, LLVMContext &context) {
+    char scalarTypeName = funcName.str()[6];
```
wsmoses, line 4718: Remove for now?

wsmoses:
Modulo the rebase on https://github.com/EnzymeAD/Enzyme/pull/528 and potentially the pass-by-reference handling required this looks good!

Also we're going to want to list the blas routines behaviors in differential use analysis and activity analysis, since in theory we should be able to have stronger information than a generic call fallback.

vchuravy:
bump! It would be really good to get this done. This is the biggest blocker for Enzyme.jl currently

enzyme/Enzyme/AdjointGenerator.h
tansongchen: sure

enzyme/Enzyme/AdjointGenerator.h
tansongchen: I think I would first merge the code I currently have, which haven't handled pass-by-reference. This is already useful in many cases (c/c++ applications and Enzyme.jl). Handling fortran-style BLAS is way more complicated so that it would be better done when we have TableGen

enzyme/Enzyme/AdjointGenerator.h
tansongchen: It is checked by the assert statement below

enzyme/Enzyme/AdjointGenerator.h
tansongchen: We don't need to handle either of these cases.
Case 1: x active, y inactive, then y have no gradient value to add to dx, so do nothing;
Case 2: x inactive, y active, then x have no gradient to be added to, so do nothing.
In summary: Unless x and y are both active, do nothing.

enzyme/Enzyme/AdjointGenerator.h
tansongchen: Which function should I add to? and is there some example to see how to add them

tansongchen:
Let's get this merged step-by-step. I am firstly adding swap and copy functions here, and then the remaining level 1 functions, and then level 2, and then level 3. A helper function is implemented to handle the case where we swap a active vector with a constant vector.

wsmoses:
Reopening to preseve tests which we should add cc @ZuseZ4

wsmoses:
superceeded by upstream

PR #371, opened 2021-10-30
Simplify argument handling and add enzyme_nofree option

wsmoses:
@jedbrown @LeilaGhaffari While ideally Enzyme won't need to allocate any extra memory for your use case, the `enzyme_nofree` added here will allow you to run split mode multiple times on the same tape.

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -885,6 +885,9 @@ public:
     case DerivativeMode::ReverseModeCombined: {
       return;
     }
+    case DerivativeMode::ForwardModeVector:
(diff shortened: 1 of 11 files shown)


LeilaGhaffari:
Thanks, @wsmoses for taking care of this so quickly. 
I checked out this branch and made it. Then, I tried libCEED by adding `-fno-unroll-loops` to the flags and replaced `enzyme_tape`s with `enzyme_nofree` inside the code but the compilation fails. Am I missing something?

wsmoses:
Try adding enzyme_nofree, rather than replacing enzyme_tape with enzyme_nofree.

vchuravy (approved):

PR #375, opened 2021-11-03
Vector mode

tgymnich:


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -681,11 +698,16 @@ class Enzyme : public ModulePass {
     const AugmentedReturn *aug;
     switch (mode) {
     case DerivativeMode::ForwardModeVector:
+      newFunc = Logic.CreateForwardDiff(
+          cast<Function>(fn), retType, constants, TLI, TA,
+          /*should return*/ false, /*dretPtr*/ false, mode, /* width */ 3,
```
wsmoses, line 703: It would be nice to have this automatically deduced rather than hard coded, or perhaps come as a parameter?

tgymnich: Of course this is just my quick and dirty proof of concept

tgymnich: It’s mainly a draft of the C calling convention for vector mode

wsmoses: Yeah for sure, just noting this for the future.

wsmoses:
Per discussion, mind first refactoring out the vector mode flag to a parameter as a dependency PR?

enzyme/test/Enzyme/ForwardModeVector/sub.ll
```diff
@@ -0,0 +1,38 @@
...
+; Function Attrs: noinline nounwind readnone uwtable
+define double @tester(double %x, double %y) {
+entry:
+  %0 = fsub fast double %x, %y
+  ret double %0
+}
+
+define %struct.Gradients @test_derivative(double %x, double %y) {
+entry:
+  %0 = tail call %struct.Gradients (double (double, double)*, ...) @__enzyme_fwddiff(double (double, double)* nonnull @tester, metadata !"enzyme_dupn", i64 2, double %x, double 1.0, double 0.0, metadata !"enzyme_dupn", i64 2, double %y, double 0.0, double 1.0)
```
wsmoses, line 17: Is it useful to have dupn instead of dup? We can have a single var for the width like enzyme_width, and then all dup's are defined to be the corresponding width?

tgymnich: Sure let's do this.

PR #378, opened 2021-11-12
Fix gradient struct return on ARM

tgymnich:
When using autodiff to receive all the gradients in a struct,
the return type of `CI` on ARM emited by clang is `%struct.Foo = type { double, double }` instead of `{ double, double }` on x86. This make the following check fail: `if (diffret->getType() == CI->getType())`

To fix this we can just compare the struct layouts as follows: `CIsty->isLayoutIdentical(diffretsty)`.

All other tests  in `check-enzyme` pass on M1.

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -862,6 +862,8 @@ public:
         }
       }
     }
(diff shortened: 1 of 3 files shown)


wsmoses (approved):
Add a test, otherwise LGTM

PR #380, opened 2021-11-15
ARM Bugfixes

tgymnich:
handle insertvalue for arrays


wsmoses (approved):
I didn't realize this was legal on arrays.

Add an LLVM test (and also double check type analysis is fine)?

PR #381, opened 2021-11-17
Don't get diffe of constant return

tgymnich:


wsmoses (approved):
Add a test case?

tgymnich:
This should not change the generated code since, `diffe(...)` would just return a constant 0 if we pass a constant value. This PR just avoid the warning printed by diffe(...)`. The code path should already be covered by unit tests (e.g.`constant.ll`).

PR #389, opened 2021-12-09
Refactor blas

wsmoses:


vchuravy (approved):

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4107,6 +4107,191 @@ class AdjointGenerator
...
   }
 
+  bool handleBLAS(llvm::CallInst &call, Function *called, StringRef funcName,
+                  const std::map<Argument *, bool> &uncacheable_args) {
+    CallInst *const newCall = cast<CallInst>(gutils->getNewFromOriginal(&call));
+    IRBuilder<> BuilderZ(newCall);
+    BuilderZ.setFastMathFlags(getFast());
+
+    if ((funcName == "cblas_ddot" || funcName == "cblas_sdot") &&
+        called->isDeclaration()) {
```
tgymnich, line 4117:
```suggestion
    if (called->isDeclaration()) {
```

tgymnich (approved):

enzyme/Enzyme/AdjointGenerator.h
wsmoses: While I agree we can do that here, the goal of this refactor is that like the MPI function, this can contain all the blas code without getting in the way of the main call handler.

tgymnich: there are further checks down below for the respective blas functions

wsmoses: Yes, but this part of the code is specific to ddot (with the specific checks for the double/float variant). This code will not work, for example, for gemm.

tgymnich: ok got it

PR #395, opened 2021-12-14
ForwardMode handle function pointers

tgymnich:


enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -2680,6 +2683,10 @@ Constant *GradientUtils::GetOrCreateShadowFunction(EnzymeLogic &Logic,
     llvm::errs() << "warning: assuming realloc only creates pointers\n";
     type_args.Return.insert({-1, -1}, BaseType::Pointer);
   }
+  
+  switch (mode) {
```
vchuravy, line 2687: Dead code?

tgymnich: yes

enzyme/Enzyme/FunctionUtils.cpp
```diff
@@ -1632,8 +1617,31 @@ Function *PreProcessCache::CloneFunctionWithReturns(
...
+    ValueToValueMapTy *VMapO, bool diffeReturnArg, llvm::Type *additionalArg) {
+  assert(!F->empty());
+  F = preprocessForClone(F, mode);
+  llvm::ValueToValueMapTy VMap;
+  llvm::FunctionType *FTy =
+      getFunctionTypeForClone(F->getFunctionType(), additionalArg,
+                              constant_args, diffeReturnArg, returnValue);
+
+  for (BasicBlock &BB : *F) {
+    for (Instruction &I : BB) {
```
wsmoses, line 1637: For efficiency sake, can you just use the terminator of the block, rather than looping through all instructions?

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -2711,43 +2711,71 @@ Constant *GradientUtils::GetOrCreateShadowFunction(EnzymeLogic &Logic,
...
+  case DerivativeMode::ForwardMode: {
+    Constant *newf =
+        Logic.CreateForwardDiff(fn, retType, types, TLI, TA, false, false, mode,
+                                nullptr, type_args, uncacheable_args);
+
+    if (!newf)
+      newf = UndefValue::get(fn->getType());
+
+    std::string globalname = ("_enzyme_" + fn->getName() + "'").str();
```
wsmoses, line 2723: You probably want a different name from reverse to avoid inadvertant name collision.

enzyme/Enzyme/FunctionUtils.cpp

enzyme/Enzyme/GradientUtils.cpp
tgymnich: done. also renamed reverse

wsmoses (approved):
Enable some of the ode tests if possible as well since I presume this should enable many of those.

PR #400, opened 2021-12-15
Adding Julia and Rust install instructions

ZuseZ4:
I refactored it a bit to make it easier to add new languages.
I didn't touch the C/C++ instructions.

### DIFF
--- a/content/Installation/C++.md
+++ b/content/Installation/C++.md
@@ -1,0 +1,90 @@
+---
+title: "C and C++"
+date: 2019-11-29T15:26:15Z
+draft: false
+weight: 20
+---
+
+You can install Enzyme using [Homebrew](https://brew.sh), or build it manually from source. To install using Homebrew, run
+
+```
+brew install enzyme
+```
+
+Currently, Homebrew has a pre-built binary (bottle) on macOS, but will build from source on Linux. To request a bottle for Linux, [file an issue](https://github.com/Homebrew/linuxbrew-core/issues) at Homebrew.
+
+The rest of these instructions explain how to build Enzyme from source.
+
+## Downloading Enzyme
+To start you should download Enzyme's code [Github](https://github.com/wsmoses/Enzyme).
+
+```sh
+git clone https://github.com/wsmoses/Enzyme
+cd Enzyme
+```
+
+
+## Building LLVM
+
+Enzyme is a plugin for LLVM and consequently needs an existing build of LLVM to function.
+
+Enzyme is designed to work with a wide range of LLVM versions and is currently tested against LLVM 7, 8, 9, 10, 11, 12, and mainline. LLVM's plugin infrastructure can sometimes be flakey or not built by default. If loading Enzyme into an existing LLVM installation results in segfaults, we recommend building LLVM from source.
+
+Details on building LLVM can be found in for building LLVM can be found in the [LLVM Getting Started](https://llvm.org/docs/GettingStarted.html). A simple build command using Ninja is shown below:
+
+```sh
+cd /path/to/llvm/source/
+mkdir build && cd build
+cmake -G Ninja ../llvm -DLLVM_TARGETS_TO_BUILD="host" -DLLVM_ENABLE_PROJECTS="clang" -DLLVM_ENABLE_PLUGINS=ON -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON
+ninja
+```
+
+## Building Enzyme
+
+First enter the enzyme project directory inside the monorepo and create a build directory. Note that the source is inside a subdirectory of the repo called enzyme. If you get a CMakeLists.txt not found error, make sure you're pointing at the enzyme subdirectory.
+
+```sh
+cd /path/to/Enzyme/enzyme
+mkdir build && cd build
+```
+
+From here, we can configure Enzyme via cmake, then build. Again, for ease we use Ninja (requiring the ninja build to to be installed). One can use make by omitting `-G Ninja` and running make instead of ninja.
+
+```sh
+cmake -G Ninja .. -DLLVM_DIR=/path/to/llvm/lib/cmake/llvm
+ninja
+```
+
+This should create a file `Enzyme/LLVMEnzyme-<VERSION>.so` inside the build directory, which contains the LLVM analysis and optimization passes that implement Enzyme.
+
+
+
+## Verifying installation
+
+We can run Enzyme's unit tests by running the following command. They should run in less than a minute and verify that your build of Enzyme and LLVM interoperate properly.
+
+```sh
+ninja check-enzyme
+```
+
+We can also run Enzyme's C/C++ integration tests. These tests require an existing installation of the Adept AD Engine (to compare against) and Eigen. Running these tests will take a moderate amount of time (about 6 minutes on a recent multicore).
+
+```sh
+ninja check-enzyme-integration
+```
+
+Finally, we can also run Enzyme's benchmarking suite, which is composed of the reverse mode tests from Microsoft's [ADBench suite](https://github.com/microsoft/ADBench) and other interesting cases. Running these tests will potentially take a long time (about an hour on a recent multicore).
+
+```sh
+ninja bench-enzyme
+````
+
+If you run Enzyme tests and get an error like `/bin/sh: 1: ../../: Permission denied` or ` ../../ not found`, it's likely that cmake wasn't able to find your version of llvm-lit, LLVM's unit tester. This often happens if you use the default Ubuntu install of LLVM as they stopped including it in their packaging. To remedy, find lit.py or lit or llvm-lit on your system and add the following flag to cmake:
+```sh
+cmake .. -DLLVM_EXTERNAL_LIT=/path/to/lit/lit.py
+```
+
+## Developing inside a Container
+
+For debugging and testing purposes we have created [Docker images](https://github.com/tgymnich/enzyme-dev-docker), which closely resemble all of our CI environments. If you are using Visual Studio Code you can build and test Enzyme inside of a [Dev Container](https://code.visualstudio.com/docs/remote/containers). To change either the Ubuntu or the LLVM version in Visual Studio Code just edit the file at `.devcontainer/devcontainer.json` accordingly.
+
--- a/content/Installation/Julia.md
+++ b/content/Installation/Julia.md
@@ -1,0 +1,17 @@
+---
+title: "Rust"
+date: 2019-11-29T15:26:15Z
+draft: false
+weight: 40
+---
+
+Enzyme.jl can be installed in the usual way Julia packages are installed:
+
+```sh
+] add Enzyme
+```
+
+The Enzyme binary dependencies will be installed automatically via Julia's binary actifact system.
+
+The Enzyme.jl API revolves around the function `autodiff`, see it's [documentation](https://enzyme.mit.edu/julia/api/#Enzyme.autodiff-Union{Tuple{A},%20Tuple{F},%20Tuple{F,%20Type{A},%20Vararg{Any}}}%20where%20{F,%20A%3C:Enzyme.Annotation}) for details and a usage example. 
+Also see [Implementing pullbacks](https://enzyme.mit.edu/julia/pullbacks/) on how to use Enzyme.jl to implement back-propagation for functions with non-scalar results.
--- a/content/Installation/Julia.md
+++ b/content/Installation/Julia.md
@@ -1,5 +1,5 @@
 ---
-title: "Rust"
+title: "Julia"
 date: 2019-11-29T15:26:15Z
 draft: false
 weight: 40
--- a/content/Installation/Rust.md
+++ b/content/Installation/Rust.md
@@ -1,0 +1,63 @@
+---
+title: "Rust"
+date: 2019-11-29T15:26:15Z
+draft: false
+weight: 30
+---
+
+Getting started in Rust is easy, if you are using Linux. Otherwise, you probably need to wait for our next implementation, which will support Windows and Mac too.  
+We have published `enzyme` on [crates.io](https://crates.io/crates/enzyme).
+
+So after getting the following dependencies:
+
+- git
+- ninja
+- cmake
+- libssl-dev
+- libclang-dev
+- Rust (rustup) and any nightly toolchain 
+- ~10GB free storage in $HOME/.cache
+
+you can run
+```sh
+cargo install enzyme
+```
+
+This should only take a minute and give you an enzyme-install binary which you can use to build rustc/llvm/enzyme and a bunch of extra things, which might come helpful.  Now you can start the build process by calling
+```bash
+enzyme-install
+```
+It will build everything inside of `$HOME/.cache`, so please have around 10GB available there.
+Also take care, the build process takes around one hour on a recent 8 core cpu.
+
+## Verifying installation
+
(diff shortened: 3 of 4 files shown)


content/Installation/Rust.md
```diff
@@ -0,0 +1,63 @@
...
+```sh
+cargo install enzyme
+```
+
+This should only take a minute and give you an enzyme-install binary which you can use to build rustc/llvm/enzyme and a bunch of extra things, which might come helpful.  Now you can start the build process by calling
+```bash
+enzyme-install
+```
+It will build everything inside of `$HOME/.cache`, so please have around 10GB available there.
+Also take care, the build process takes around one hour on a recent 8 core cpu.
```
jedbrown, line 31: If I already have a `LLVMEnzyme-13.so` built, is there a way to use it from Rust without this time-consuming step? Besides compilation cost, I'm interested in using Enzyme from C and Rust within the same application.

wsmoses: @ZuseZ4, if I recall the issue here is not compiling Enzyme, but compiling Rust's LLVM, yes?

Also note that these instructions are aimed at the current work in progress nicer language integration (e.g. fewer compiler flag hacks + syntactic sugar) of Enzyme within Rust (see https://github.com/rust-ml/oxide-enzyme).

Using an existing Enzyme build (for a rust-compatible LLVM version) presumably would still work by either manually using the LLVM output of rust, or say using LLD as here (https://internals.rust-lang.org/t/automatic-differentiation-differential-programming-via-llvm/13188/7)

wsmoses:
A couple of higher-level comments, the separation you have of the Enzyme build into the "C/C++" section, isn't really a proper description. That's how Enzyme itself is built, it's able to be easily used within clang (a C/C++ compiler), but also Fortran, Swift, and is how one would build Enzyme from scratch as is needed by say Julia or Rust.

The difference here is that Julia and Rust also have language-specific package managers where we also vend binaries automatically.

As for julia, I think it would make more sense to redirect to enzyme.mit.edu/julia rather than have the separate page.

I would also be inclined to serve the rust bindings (currently at https://rust-ml.github.io/oxide-enzyme) at enzyme.mit.edu/rust if that sounds good to you. For Julia, the docs are still generated by (https://github.com/wsmoses/Enzyme.jl) and the webserver does appropriate forwarding.

ZuseZ4:
Yes, I just left the Enzyme build instructions in the C/C++ section, since the Rust and Julia solutions rely on the package manager. Also, you build clang along with LLVM, so I wasn't sure how to split that up.
`cmake -G Ninja ../llvm -DLLVM_TARGETS_TO_BUILD="host" -DLLVM_ENABLE_PROJECTS="clang" -DLLVM_ENABLE_PLUGINS=ON -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON`

I'm fine with having the Rust code at an extra website and would also prefer it. It makes it a little bit easier to align the layout with other rust / rust-ml websites. People will probably also be less confused about which page belongs to which language.
I can ping you once I have decided which documentation should be served at enzyme.mit.edu/rust.

Which place shall we then use to forward Rust and Julia users? I think something directly visible on the front page would be nice, to make sure that user's don't get lost in C++ specific instructions.

content/Installation/Rust.md
ZuseZ4: Fair question. I was already wondering if I should add a note about just that while writing. The short version is maybe. 

As mentioned by @wsmoses, the main reason of having the current build procedure is stability. Each Rustc is only guaranteed to be compile-able against one specific LLVM commit (not even an LLVM release). So we could invest time to speed up the getting started process, but at the cost of introducing more bugs due to incompatibilities. Also, our main goal is a proper rustc integration, which is required due to some other issues. Once that's achieved we will archive the entire enzyme_build archive, so all optimizations here are only of limited duration. If you care about using C and rust together, 
I would just recommend you to use the clang version which I build along with rustc. This way Enzyme, Clang, rustc are
all based on the same LLVM version, which is probably the most reliable setup that you can get. That's what I'm using right now for some C++ code too. You will find clang in `/home/$USER/.cache/enzyme/rustc-1.57.0-src/build/x86_64-unknown-linux-gnu/llvm/bin/clang`. It might cost you two cpu-hrs, but it will probably safe you more hours of debugging/developing.

However, if you prefer to test your existing Enzyme version together with an official rustc, here is some background /history:
I added a flag to rustc, which allows loading LLVM plugins. This flag will only be available if you build rustc from source, as it is done by my script. Having this plugin interface has two purposes:
1) we can run some specific Enzyme optimizations before AD
2) We could use the __enzyme_ notation, as it is done in the C++ examples.

However: we moved away from the __enzyme notation and we haven't added the usage of the enzyme optimizations yet,
so you should theoretically be able to use some other rustc. You can do that by replacing `+enzyme` with `+nightly`
when building a crate. I just tested it an it fails due to having `__rust_probestack` defined more than once. I handle this [here](https://github.com/rust-ml/oxide-enzyme/blob/68a1f1eec636eeb1a07f273660abd72a9d5f7eba/src/lib.rs#L548), 
so you could try to debug why it's working with the enzyme, but not with the nightly toolchain.
Finally you could try to build rustc based on your existing LLVM Version, but as mentioned, rustc devs don't offer any
guarantees about that working. Also, building rustc-stage2 even without llvm takes a descent amount of time.

So yes, you might be able to sip this time-consuming step, I just really don't know if it's recommendable atm.
Either way, just let me know what you prefer, and I can try to help with that.

wsmoses:
Perhaps right before the homebrew instructions, we could add something like: 

For languages with their own package manager, we have pre-built Enzyme binaries that one can install in the usual way. Click here (link to enzyme.mit.edu/julia) for information on installing Enzyme in Julia and  here (link to enzyme.mit.edu/rust) for Rust.

If you're working in a language other than those above, or otherwise want to compile Enzyme from source, read on.

content/Installation/Rust.md
jedbrown: Thanks for all these details. I'll just follow the procedure for now, though I hope it'll be possible to upstream what is necessary.

I'd be okay using one build of Enzyme with release clang for differentiating C and another built internally for rustc. Enzyme builds pretty fast and distribution will be a hassle if one has to always match. (For example, users may build our libraries on an HPC machine with vendor MPI set up to use their version of clang. Building rustc+llvm from source and then using environment variables to make the vendor MPI use our special clang will be an error-prone user experience.)

One follow-up: what's the process for incremental rebuilds instead of the released crate, say upon pulling Enzyme or if one is hacking on Enzyme?

content/Installation/Rust.md
```diff
@@ -0,0 +1,63 @@
...
+git clone https://github.com/rust-ml/oxide-enzyme
+cd oxide-enzyme/example
+cargo enzyme
+```
+You will find your binary slightly more hidden than usual in `target/$TARGET/debug/example`.
+
+## Using Enzyme
+
+In order to play nicely with cargo, the Rust workflow differs somewhat from the [C++](https://enzyme.mit.edu/getting_started/UsingEnzyme/) using Enzyme section.
+First, you need to specify in your `build.rs` file how which function should be differentiated.
```
jedbrown, line 53:
```suggestion
First, you need to specify in your `build.rs` file how each function should be differentiated. 
```
would be useful to link to an example. Also, is this `build.rs` approach a short-term solution or is it something you envision supporting longer term? I was hoping for a macro that could go in the source, ideally either in the module with the definition or in another module.

content/Installation/Rust.md
```diff
@@ -0,0 +1,63 @@
...
+cargo enzyme
+```
+You will find your binary slightly more hidden than usual in `target/$TARGET/debug/example`.
+
+## Using Enzyme
+
+In order to play nicely with cargo, the Rust workflow differs somewhat from the [C++](https://enzyme.mit.edu/getting_started/UsingEnzyme/) using Enzyme section.
+First, you need to specify in your `build.rs` file how which function should be differentiated. 
+Afterwards, you can run `cargo enzyme` without additional arguments, to build your crate.
+If you want to have some more controll, you can also modify the following build command, which `cargo enzyme` is using:
```
jedbrown, line 55:
```suggestion
If you want to have some more control, you can also modify the following build command that `cargo enzyme` uses:
```

jedbrown (approved):

content/Installation/Rust.md
ZuseZ4: So I just pushed a new version to crates.io yesterday, which moved to the latest enzyme release 0.0.25. I've added some minimal logic to recognize whether enzyme or rustc are outdated, and only build what is needed.

If you made some changes to enzyme locally and want to test them, I recommend to just build it against the llvm from the rustc-1.57 build.
If you place your enzyme repo folder inside of '$HOME/.cache/enzyme' you can then update the ENZYME_VER in oxide-enzyme/src/lib.rs to whatever name you used for your folder.
You also need to adjust the LD_LIBRARY_PATH, but then you should be ready. I will try to add this also to the README before publishing the next enzyme version on crates.io.

content/Installation/Rust.md
```diff
@@ -0,0 +1,63 @@
...
+```sh
+git clone https://github.com/rust-ml/oxide-enzyme
+cd oxide-enzyme/example
+cargo enzyme
+```
+You will find your binary slightly more hidden than usual in `target/$TARGET/debug/example`.
+
+## Using Enzyme
+
+In order to play nicely with cargo, the Rust workflow differs somewhat from the [C++](https://enzyme.mit.edu/getting_started/UsingEnzyme/) using Enzyme section.
```
wsmoses, line 52: Change C++ to generic?

content/Installation/Rust.md
```diff
@@ -0,0 +1,63 @@
...
+First, you need to specify in your `build.rs` file how which function should be differentiated. 
+Afterwards, you can run `cargo enzyme` without additional arguments, to build your crate.
+If you want to have some more controll, you can also modify the following build command, which `cargo enzyme` is using:
+```sh
+RUSTFLAGS="--emit=llvm-bc" cargo +enzyme -Z build-std rustc --target x86_64-unknown-linux-gnu -- --emit=llvm-bc -g -C opt-level=3 -Zno-link && RUSTFLAGS="--emit=llvm-bc" cargo +enzyme -Z build-std rustc --target x86_64-unknown-linux-gnu -- --emit=llvm-bc -g -C opt-level=3
+```
+
+Please be aware that Enzyme currently can't be used in dependencies. However you can use Enzyme to differentiate functions and types defined in your dependencies. 
+We still have quite a few issues open and are lacking some documentation. So you might find [this issue](https://github.com/rust-ml/oxide-enzyme/issues/6) and 
+[this C++ documentation](https://enzyme.mit.edu/getting_started/CallingConvention/#types) useful to understand the Rust frontend.
```
wsmoses, line 62: Change "this C++ documentation" to "Enzyme argument convention documentation"?

content/Installation/Rust.md
ZuseZ4: So, you still think we should merge this (in an updated manner)?
I thought we would just leave the installation instruction as they are and just add the links to the julia and rust website, as mentioned above?

content/Installation/Rust.md
```diff
@@ -0,0 +1,63 @@
...
+Afterwards, you can run `cargo enzyme` without additional arguments, to build your crate.
+If you want to have some more controll, you can also modify the following build command, which `cargo enzyme` is using:
+```sh
+RUSTFLAGS="--emit=llvm-bc" cargo +enzyme -Z build-std rustc --target x86_64-unknown-linux-gnu -- --emit=llvm-bc -g -C opt-level=3 -Zno-link && RUSTFLAGS="--emit=llvm-bc" cargo +enzyme -Z build-std rustc --target x86_64-unknown-linux-gnu -- --emit=llvm-bc -g -C opt-level=3
+```
+
+Please be aware that Enzyme currently can't be used in dependencies. However you can use Enzyme to differentiate functions and types defined in your dependencies. 
+We still have quite a few issues open and are lacking some documentation. So you might find [this issue](https://github.com/rust-ml/oxide-enzyme/issues/6) and 
+[this C++ documentation](https://enzyme.mit.edu/getting_started/CallingConvention/#types) useful to understand the Rust frontend. 
+If you have specific question please open an issue and if you have some time, we are always happy to review PR's which add documentation or examples.
```
wsmoses, line 63: Can you add a `for developers` section which describes how one can use the Rust frontend with a custom build/commit of Enzyme (for the purpose of development/debugging). 

See https://enzyme.mit.edu/julia/dev_docs/ for a Julia equivalent.

content/Installation/_index.md
```diff
@@ -2,88 +2,17 @@
...
+Enzyme can be used to differentiate any langauge which compiles to LLVM (C/C++, Rust, Swift, Julia, Fortran, Tensorflow, etc).
+We currently offer extra support (documentation, nice wrappers, package-manger integration, etc) for some of them.
 
-```
-brew install enzyme
-```
+Plase see the [C/C++](/Installation/C++) guide for information on how to set everything up for C and C++.
 
-Currently, Homebrew has a pre-built binary (bottle) on macOS, but will build from source on Linux. To request a bottle for Linux, [file an issue](https://github.com/Homebrew/linuxbrew-core/issues) at Homebrew.
+Plase see the [Julia](/Installation/Julia) guide for information on how to set everything up for Julia.
```
wsmoses, line 13: Change link for Julia to /julia

content/Installation/Julia.md
```diff
@@ -0,0 +1,17 @@
+---
+title: "Julia"
+date: 2019-11-29T15:26:15Z
```
wsmoses, line 3: Remove this file since superseded by enzyme.mit.edu/julia

content/Installation/_index.md
```diff
@@ -2,88 +2,17 @@
...
 ---
 
-You can install Enzyme using [Homebrew](https://brew.sh), or build it manually from source. To install using Homebrew, run
+Enzyme can be used to differentiate any langauge which compiles to LLVM (C/C++, Rust, Swift, Julia, Fortran, Tensorflow, etc).
+We currently offer extra support (documentation, nice wrappers, package-manger integration, etc) for some of them.
 
-```
-brew install enzyme
-```
+Plase see the [C/C++](/Installation/C++) guide for information on how to set everything up for C and C++.
```
wsmoses, line 11: Change this to "From source" rather than C/C++?

content/Installation/C++.md
```diff
@@ -0,0 +1,90 @@
+---
+title: "C and C++"
+date: 2019-11-29T15:26:15Z
+draft: false
+weight: 20
+---
+
+You can install Enzyme using [Homebrew](https://brew.sh), or build it manually from source. To install using Homebrew, run
```
wsmoses, line 8: Move both the homebrew and docker/dev container section to the main installation file (making this the "from source" file)

content/Installation/Rust.md
wsmoses: At minimum right now more docs are better than none, so having rust instructions is definitely a plus.

I do think the more elegant solution is that this part of the Rust documentation is at enzyme.mit.edu/rust, which we can still move to after the fact.

I'd say for now here this would still be useful as a way to better organize the installation page into a top level (link to from source / homebrew package manager / docker container / rust+julia package manager sections), and corresponding sections with docs for.

wsmoses: If it's okay with you, do you mind pinging me again when some of the changes have been made for another look before merging?

In particular, the generic instructions are used by folks who play with other languages [like Swift/Fortran], and I don't want them getting lost/confused when there's no section for their language.

tgymnich:
Need to be moved to EnzymeAD/www

PR #412, opened 2021-12-25
Include instructions for one-line call to clang w Enzyme as a plugin

ludgerpaehler:
Trying to point out "Getting Started"'s nature of breaking down the individual steps involved in compilation, and being able to summarise these steps into a single compilation call.

Questions left open for the moment are:
- Are we able to highlight the single-line call with a highlighting box or similar, e.g. by injecting HTML code?
- Should this have its own separate sub-header?

### DIFF
--- a/content/getting_started/UsingEnzyme.md
+++ b/content/getting_started/UsingEnzyme.md
@@ -5,13 +5,6 @@ draft: false
 weight: 10
 ---
 
----
-title: "Using Enzyme"
-date: 2019-11-29T15:26:15Z
-draft: false
-weight: 10
----
-
 ## Generating LLVM
 
 To begin, let's create a simple code `test.c` we want to differentiate. Enzyme will replace any calls to functions whose names contain "\_\_enzyme\_autodiff" with calls to the corresponding For now, let's ignore the details of Enzyme's calling convention/ABI which are described in detail [here](/getting_started/CallingConvention)
--- a/content/getting_started/UsingEnzyme.md
+++ b/content/getting_started/UsingEnzyme.md
@@ -5,6 +5,13 @@ draft: false
 weight: 10
 ---
 
+---
+title: "Using Enzyme"
+date: 2019-11-29T15:26:15Z
+draft: false
+weight: 10
+---
+
 ## Generating LLVM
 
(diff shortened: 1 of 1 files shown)


content/getting_started/UsingEnzyme.md
```diff
@@ -5,6 +5,13 @@ draft: false
 weight: 10
 ---
 
+---
```
wsmoses, line 8: This shouldn't be repeated?

ludgerpaehler: Yes, this should absolutely not get repeated!

content/getting_started/UsingEnzyme.md
```diff
@@ -120,9 +127,15 @@ We can then compile this to a final binary as follows:
...
+For ease, we could combine the final optimization and binary execution into one command as follows.
+```sh
+clang output.ll -O2 -o a.exe
+```
+
+These individual steps can be summarised into a single compilation step, where we load Enzyme as a plugin to our compiler clang, and perform the differentiation on our test code.
+
 ```sh
-clang output.ll -O3 -o a.exe
+clang test.c -O2 -Xclang -load -Xclang /path/to/Enzyme/enzyme/build/Enzyme/ClangEnzyme-<VERSION>.so -fno-vectorize -fno-slp-vectorize -fno-unroll-loops -o a.exe
```
wsmoses, line 138: I might rotate this and start by describing the clang path, and then discussing the internals as what's happening

content/getting_started/UsingEnzyme.md
ludgerpaehler: Solved it with the patch below

content/getting_started/UsingEnzyme.md
ludgerpaehler: Yes, that would definitely be an option. I think we should just make it more apparent that the steps explained as of now are basically a breakdown of the compilation steps when Enzyme is used as a compiler plugin, as in the call on line 138.

ludgerpaehler:
Closed due to being included in the docs pull-request, with the above comments worked into the new ``first-derivative`` page.

PR #425, opened 2022-01-18
Implement vector mode utils

tgymnich:
Minimal working example for cast inst + unwrapAndApply family of utility functions and getTypeForVectorMode.

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -1005,19 +1005,24 @@ public:
       break;
     }
     case DerivativeMode::ForwardMode: {
-      Value *orig_op0 = I.getOperand(0);
-
       IRBuilder<> Builder2(&I);
       getForwardBuilder(Builder2);
 
-      if (!gutils->isConstantValue(orig_op0)) {
-        Value *dif = diffe(orig_op0, Builder2);
-        setDiffe(&I, Builder2.CreateCast(I.getOpcode(), dif, I.getType()),
-                 Builder2);
-      } else {
-        setDiffe(&I, Constant::getNullValue(I.getType()), Builder2);
-      }
+      Value *orig_op0 = I.getOperand(0);
+      Type *diffTy = gutils->getShadowType(I.getType());
 
+      auto rule = [&](Value *dif) {
+        return Builder2.CreateCast(I.getOpcode(), dif, I.getType());
+      };
+
+      if (!gutils->isConstantValue(orig_op0)) {
+        Value *dop0 = diffe(orig_op0, Builder2);
+        Value *diff = applyChainRule(I.getType(), Builder2, rule, dop0);
+
+        setDiffe(&I, diff, Builder2);
+      } else {
+        setDiffe(&I, Constant::getNullValue(diffTy), Builder2);
+      }
       break;
     }
     }
@@ -1612,6 +1617,22 @@ public:
   void setDiffe(Value *val, Value *dif, IRBuilder<> &Builder) {
     assert(Mode != DerivativeMode::ReverseModePrimal);
     ((DiffeGradientUtils *)gutils)->setDiffe(val, dif, Builder);
+  }
+
+  /// Unwraps a vector derivative from its internal representation and applies a
+  /// function f to each element. Return values of f are collected and wrapped.
+  template <typename Func, typename... Args>
+  Value *applyChainRule(Type *diffType, IRBuilder<> &Builder, Func rule,
+                        Args... args) {
+    return ((DiffeGradientUtils *)gutils)
+        ->applyChainRule(diffType, Builder, rule, args...);
+  }
+
+  /// Unwraps a vector derivative from its internal representation and applies a
+  /// function f to each element.
+  template <typename Func, typename... Args>
+  void applyChainRule(IRBuilder<> &Builder, Func rule, Args... args) {
+    ((DiffeGradientUtils *)gutils)->applyChainRule(Builder, rule, args...);
   }
 
   bool shouldFree() {
--- a/enzyme/Enzyme/EnzymeLogic.cpp
+++ b/enzyme/Enzyme/EnzymeLogic.cpp
@@ -2413,7 +2413,8 @@ void createTerminator(TypeResults &TR, DiffeGradientUtils *gutils,
       } else if (!gutils->isConstantValue(ret)) {
         toret = gutils->diffe(ret, nBuilder);
       } else {
-        toret = Constant::getNullValue(ret->getType());
+        Type *retTy = gutils->getShadowType(ret->getType());
+        toret = Constant::getNullValue(retTy);
       }
 
       break;
--- a/enzyme/Enzyme/FunctionUtils.cpp
+++ b/enzyme/Enzyme/FunctionUtils.cpp
@@ -1633,13 +1633,23 @@ FunctionType *getFunctionTypeForClone(
     bool diffeReturnArg, ReturnType returnValue, DIFFE_TYPE returnType) {
   std::vector<Type *> RetTypes;
   if (returnValue == ReturnType::ArgsWithReturn ||
-      returnValue == ReturnType::ArgsWithTwoReturns ||
-      returnValue == ReturnType::Return ||
-      returnValue == ReturnType::TwoReturns)
+      returnValue == ReturnType::Return) {
+    if (returnType != DIFFE_TYPE::CONSTANT) {
+      RetTypes.push_back(
+          GradientUtils::getShadowType(FTy->getReturnType(), width));
+    } else {
+      RetTypes.push_back(FTy->getReturnType());
+    }
+  } else if (returnValue == ReturnType::ArgsWithTwoReturns ||
+             returnValue == ReturnType::TwoReturns) {
     RetTypes.push_back(FTy->getReturnType());
-  if (returnValue == ReturnType::ArgsWithTwoReturns ||
-      returnValue == ReturnType::TwoReturns)
-    RetTypes.push_back(FTy->getReturnType());
+    if (returnType != DIFFE_TYPE::CONSTANT) {
+      RetTypes.push_back(
+          GradientUtils::getShadowType(FTy->getReturnType(), width));
+    } else {
+      RetTypes.push_back(FTy->getReturnType());
+    }
+  }
   std::vector<Type *> ArgTypes;
 
   // The user might be deleting arguments to the function by specifying them in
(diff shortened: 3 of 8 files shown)


enzyme/Enzyme/GradientUtils.h
```diff
@@ -1531,6 +1531,217 @@ class GradientUtils : public CacheUtility {
...
+
+  /// Returns the type for the internal representation of a derivative vector.
+  Type *getTypeForVectorMode(Type *ty) {
+    return getTypeForVectorMode(ty, width);
+  }
+
+  /// Returns the type for the internal representation of a derivative vector of
+  /// size `width`
+  static Type *getTypeForVectorMode(Type *ty, unsigned int width) {
+    if (ty->isIntegerTy() || ty->isFloatingPointTy()) {
```
wsmoses, line 1543: I'm not sure we want to actually use llvm vector types here as only certain widths are permitted per platform. Thus could we always use the arraytype?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -980,19 +980,32 @@ class AdjointGenerator
...
+
       if (!gutils->isConstantValue(orig_op0)) {
-        Value *dif = diffe(orig_op0, Builder2);
-        setDiffe(&I, Builder2.CreateCast(I.getOpcode(), dif, I.getType()),
-                 Builder2);
+        Value *dop0 = diffe(orig_op0, Builder2);
+
+        Value *diff;
+        if (gutils->getWidth() > 1) {
+          diff = unwrapAndApply(I.getType(), dop0, Builder2, rule);
```
wsmoses, line 1000: It would be nicer as an API here if unwrapAndApply would internally do the width check and decision.

enzyme/Enzyme/GradientUtils.h
tgymnich: The comment is outdated. This does not concern the internal representation anymore, which is always an Array.  It is still used when we want to work with those unwrapped Arrays.
> only certain widths are permitted per platform
Isn't that code gens problem\?


On another note should we even vectorise add ourself or rely on llvm to vectors our n add instructions?

enzyme/Enzyme/AdjointGenerator.h
tgymnich: sounds reasonable

enzyme/Enzyme/GradientUtils.h
wsmoses: Is this code even used then? If so can we remove it.

I would argue this is not code gen's problem yet since us creating an invalid vector (say 3) would not be something we can expect will be fixed. However upgrading to a vector from an array is something we should be able to rely on.

tgymnich: So we should basically never emit any vectors and rely on the llvm vectoriser and instselct to do the right thing.

wsmoses: For the moment that’s what I’d say. I think in the future we can enable using a vector if the width is available on the machine after doing a check.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: done

enzyme/Enzyme/GradientUtils.h
tgymnich: done

wsmoses: I would still keep this function and use it everywhere (e.g no get width>1 array type otherwise regular.

However I might rename it get shadowtype

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -1589,6 +1596,77 @@ class AdjointGenerator
     ((DiffeGradientUtils *)gutils)->setDiffe(val, dif, Builder);
   }
 
+  /// Unwraps a vector derivative from its internal representation and applies a
+  /// function f to each element. Return values of f are collected and wrapped.
+  template <typename Func>
+  Value *applyChainRule(Type *diffType, Value *diff1, Value *diff2,
```
wsmoses, line 1602: Could this be renamed something like `vectorizeFwdRule` since this looks like it can only be used in forward mode?

Can the number of Values themselves be templated (rather than separate functions)?

Also, some of these functions aren't needed for bitcast, can you wait to propose them until a PR which uses them

tgymnich: The function can also be used in reverse mode. It is used in erf for example.

wsmoses: I'm not seeing this. Do you still need to push or am I missing something?

tgymnich: Its in the vector-mode branch.

enzyme/Enzyme/GradientUtils.h
```diff
@@ -1527,24 +1616,26 @@ class DiffeGradientUtils : public GradientUtils {
     if (auto inst = dyn_cast<Instruction>(val))
       assert(inst->getParent()->getParent() == oldFunc);
     assert(inversionAllocs);
+
+    Type *type = getWidth() > 1 ? ArrayType::get(val->getType(), getWidth())
```
wsmoses, line 1620: Replace with getshadow type?

enzyme/Enzyme/GradientUtils.h
```diff
@@ -1233,9 +1239,11 @@ class GradientUtils : public CacheUtility {
 
         IRBuilder<> BuilderZ(inst);
         getForwardBuilder(BuilderZ);
-
+        Type *antiTy = getWidth() > 1
+                           ? ArrayType::get(inst->getType(), getWidth())
```
wsmoses, line 1243: Replace with getshadowtype?

enzyme/Enzyme/GradientUtils.h
```diff
@@ -1204,9 +1209,10 @@ class GradientUtils : public CacheUtility {
         if (isa<LoadInst>(inst)) {
           IRBuilder<> BuilderZ(inst);
           getForwardBuilder(BuilderZ);
-
-          PHINode *anti = BuilderZ.CreatePHI(inst->getType(), 1,
-                                             inst->getName() + "'il_phi");
+          Type *antiTy = width > 1 ? ArrayType::get(inst->getType(), width)
```
wsmoses, line 1212: And here

enzyme/Enzyme/FunctionUtils.cpp
```diff
@@ -1601,7 +1610,11 @@ FunctionType *getFunctionTypeForClone(
     ArgTypes.push_back(I);
     if (constant_args[argno] == DIFFE_TYPE::DUP_ARG ||
         constant_args[argno] == DIFFE_TYPE::DUP_NONEED) {
-      ArgTypes.push_back(I);
+      if (width > 1) {
```
wsmoses, line 1613: Replace getshadow type?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2413,7 +2413,10 @@ void createTerminator(TypeResults &TR, DiffeGradientUtils *gutils,
       } else if (!gutils->isConstantValue(ret)) {
         toret = gutils->diffe(ret, nBuilder);
       } else {
-        toret = Constant::getNullValue(ret->getType());
+        Type *retTy = gutils->getWidth() > 1
+                          ? ArrayType::get(ret->getType(), gutils->getWidth())
```
wsmoses, line 2417: getShadowType

enzyme/Enzyme/GradientUtils.h
tgymnich: done

enzyme/Enzyme/GradientUtils.h
tgymnich: done

enzyme/Enzyme/FunctionUtils.cpp

enzyme/Enzyme/FunctionUtils.cpp
```diff
@@ -1584,13 +1584,22 @@ FunctionType *getFunctionTypeForClone(
...
-      returnValue == ReturnType::ArgsWithTwoReturns ||
-      returnValue == ReturnType::Return ||
-      returnValue == ReturnType::TwoReturns)
-    RetTypes.push_back(FTy->getReturnType());
-  if (returnValue == ReturnType::ArgsWithTwoReturns ||
-      returnValue == ReturnType::TwoReturns)
-    RetTypes.push_back(FTy->getReturnType());
+      returnValue == ReturnType::Return) {
+    if (width > 1 && returnType != DIFFE_TYPE::CONSTANT) {
+      RetTypes.push_back(ArrayType::get(FTy->getReturnType(), width));
```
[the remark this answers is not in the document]

enzyme/Enzyme/EnzymeLogic.cpp
tgymnich: done

enzyme/Enzyme/GradientUtils.h
```diff
@@ -1010,37 +1010,42 @@ class GradientUtils : public CacheUtility {
...
-    if (!isOriginalBlock(*BuilderM.GetInsertBlock()))
-      ptr = lookupM(ptr, BuilderM);
 
-    if (!mask) {
-      auto ts = BuilderM.CreateStore(newval, ptr);
-      if (align)
+    auto rule = [&](Value *ptr, Value *newval) {
+      if (!isOriginalBlock(*BuilderM.GetInsertBlock()))
+        ptr = lookupM(ptr, BuilderM);
+
```
wsmoses, line 1017: The lookup should be hoisted so it's not done multiple times.

Moreover, (with the exception of the mask) should this not be okay to still perform the regular (now extra wide) store?

Also need to ponder if we need a helper alignment for shadow function

enzyme/test/Enzyme/ForwardModeVector/bitcast.ll
```diff
@@ -0,0 +1,24 @@
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -mem2reg -simplifycfg -adce -instsimplify -S | FileCheck %s
```
wsmoses, line 1: Per above, if the applyChainRule is removed and the differential load/store is done directly, can we still get this without instsimplify?

enzyme/Enzyme/GradientUtils.h
tgymnich: An extra wide store is not possible since we get n different pointers (which may not point to the same region in memory).

wsmoses: Ah misunderstood this as setDiffe rather than setPtrDiffe, you're absolutely right.

wsmoses (approved):

enzyme/Enzyme/GradientUtils.h
```diff
@@ -1010,37 +1010,42 @@ class GradientUtils : public CacheUtility {
...
-      Type *tys[] = {newval->getType(), ptr->getType()};
-      auto F = Intrinsic::getDeclaration(oldFunc->getParent(),
-                                         Intrinsic::masked_store, tys);
-      assert(align);
+        ts->setVolatile(isVolatile);
+        ts->setOrdering(ordering);
+        ts->setSyncScopeID(syncScope);
+      } else {
+        if (!isOriginalBlock(*BuilderM.GetInsertBlock()))
+          mask = lookupM(mask, BuilderM);
```
wsmoses, line 1031: Also can you hoist this mask as well

PR #432, opened 2022-01-25
Fix OpenMP bound calculation

wsmoses:


enzyme/Enzyme/CacheUtility.cpp
```diff
@@ -628,93 +626,70 @@ bool CacheUtility::getContext(BasicBlock *BB, LoopContext &loopContext,
...
+                    dyn_cast<AllocaInst>(omp_lb_post->getPointerOperand());
+                if (AI) {
+                  for (auto u : AI->users()) {
+                    CallInst *call = dyn_cast<CallInst>(u);
+                    if (!call)
+                      continue;
+                    Function *F = call->getCalledFunction();
+                    if (!F)
+                      continue;
+                    if (F->getName() == "__kmpc_for_static_init_4" ||
```
vchuravy, line 657: Can we just do  a prefix match?

vchuravy (approved):

PR #459, opened 2022-01-30
fix handling of const return activity

ZuseZ4:
Fixes the following testcase, where the return value is set to Const.
Once we exposed that for __enzyme we should add a testcase.
Still have to fix the formating.

   17 #[differentiate(d_test, Reverse, All(Active), Constant, false)]
   18 fn test(x: f64) -> f64 {
   19     2.0 * x
   20 }

### DIFF
--- a/enzyme/Enzyme/EnzymeLogic.cpp
+++ b/enzyme/Enzyme/EnzymeLogic.cpp
@@ -3421,7 +3421,7 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
           assert(differetval);
           if (!gutils->isConstantValue(orig->getReturnValue())) {
             IRBuilder<> reverseB(gutils->reverseBlocks[BB].back());
-            gutils->setDiffe(orig->getReturnValue(), Constant::getNullValue(differetval->getType()), reverseB);
+            gutils->setDiffe(orig->getReturnValue(), differetval, reverseB);
           }
         } else {
           assert(dretAlloca == nullptr);
--- a/enzyme/Enzyme/EnzymeLogic.cpp
+++ b/enzyme/Enzyme/EnzymeLogic.cpp
@@ -3421,7 +3421,7 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
           assert(differetval);
           if (!gutils->isConstantValue(orig->getReturnValue())) {
             IRBuilder<> reverseB(gutils->reverseBlocks[BB].back());
-            gutils->setDiffe(orig->getReturnValue(), differetval, reverseB);
+            gutils->setDiffe(orig->getReturnValue(), Constant::getNullValue(differetval->getType()), reverseB);
           }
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -3410,19 +3410,21 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
...
+        	if (dretAlloca) {
+          		rb.CreateStore(gutils->invertPointerM(orig->getReturnValue(), rb),
                          dretAlloca);
-        }
-        if (key.retType == DIFFE_TYPE::OUT_DIFF) {
+		}
+        } else if (key.retType == DIFFE_TYPE::OUT_DIFF) {
           assert(orig->getReturnValue());
           assert(differetval);
           if (!gutils->isConstantValue(orig->getReturnValue())) {
```
wsmoses, line 3422: If it is constant, this should be set to zero.

wsmoses: Not quite, the if the return is active this should still be set to differetval, but if it is a constant value then it should be set to zero

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -3410,19 +3410,21 @@ Function *EnzymeLogic::CreatePrimalAndGradient(
...
             IRBuilder<> reverseB(gutils->reverseBlocks[BB].back());
-            gutils->setDiffe(orig->getReturnValue(), differetval, reverseB);
+            gutils->setDiffe(orig->getReturnValue(), Constant::getNullValue(differetval->getType()), reverseB);
```
wsmoses, line 3424: ``
```suggestion
      gutils->setDiffe(orig->getReturnValue(), differetval, reverseB);
    }
      else
            gutils->setDiffe(orig->getReturnValue(), Constant::getNullValue(differetval->getType()), reverseB);
```

ZuseZ4: The issue with that is that it fails 9 of our tests since orig->getReturnValue() will give us a ConstantValue, which triggers an assertion in setDiffe:
https://github.com/wsmoses/Enzyme/blob/1f6f58d5bc1c1c78f0a1cd8aae2e89e9f6f17036/enzyme/Enzyme/GradientUtils.h#L1919
I didn't found out yet what's the best way to fix that, do you have a recommendation?

wsmoses: Oh sorry, I thought we were setting the return gradient, not the gradient of the return value. In which case the original `gutils->setDiffe(orig->getReturnValue(), differetval, reverseB);` is fine.

wsmoses (approved):

ZuseZ4:
Closed in favor of #469

PR #464, opened 2022-01-31
updated 3 functions / loops to continue / return early

ZuseZ4:


enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2407,68 +2407,68 @@ void createTerminator(TypeResults &TR, DiffeGradientUtils *gutils,
   IRBuilder<> nBuilder(nBB);
   nBuilder.setFastMathFlags(getFast());
```
wsmoses, line 2409: Add a comment that in forward mode only need to update the return

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2550,125 +2550,71 @@ void createInvertedTerminator(TypeResults &TR, DiffeGradientUtils *gutils,
 
   // Ensure phi values have their derivatives propagated
   for (auto I = oBB->begin(), E = oBB->end(); I != E; ++I) {
-    if (PHINode *orig = dyn_cast<PHINode>(&*I)) {
-      if (gutils->isConstantInstruction(orig))
-        continue;
+    PHINode *orig = dyn_cast<PHINode>(&*I);
+    if (orig == nullptr) break;
```
wsmoses, line 2554: Continue, rather than break.

ZuseZ4: Are you sure? All testcases passed.
Also, before you had that break in the following else clause on line 2755, which I moved to the beginning: 
https://github.com/wsmoses/Enzyme/blob/1f6f58d5bc1c1c78f0a1cd8aae2e89e9f6f17036/enzyme/Enzyme/EnzymeLogic.cpp#L2755

The gh diff looks generally a bit strange, probably because the largest if affected 200 lines. 
The continue shown there is not removed, it's still there: https://github.com/wsmoses/Enzyme/blob/914ac8c88e4f4a5925f9616833b21d66706acfce/enzyme/Enzyme/EnzymeLogic.cpp#L2555

wsmoses: Actually, you're right since PHIs guaranteed to be at the top we can early exit if we don't find a PHI.

enzyme/Enzyme/EnzymeLogic.cpp
ZuseZ4: done

wsmoses: Mind removing "so we might return early", otherwise lgtm

wsmoses (approved):

ZuseZ4:
@wsmoses I missed that the required formating check only runs on pushes, not if I create a PR from my own fork.
Should I close this and push on a branch directly on Enzyme instead of my fork, or can you trigger the formating check?

ZuseZ4:
Closed in favor of #470

PR #471, opened 2022-02-02
Use original function instead of new intrinsic

vchuravy:
- use original powi function
- use original pow function


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3218,7 +3218,11 @@ class AdjointGenerator
             orig_ops[1]->getType()
 #endif
           };
-          Function *PowF = Intrinsic::getDeclaration(M, Intrinsic::powi, tys);
+          auto &CI = cast<CallInst>(I));
+          Function *PowF = CI.getCalledFunction();
```
wsmoses, line 3222: You should use getCalledValue (there's an llvm version invariant helper)

wsmoses (approved):

PR #473, opened 2022-02-02
Vector mode: call

tgymnich:
call and other remaining instructions and tests

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -9882,16 +9882,19 @@ public:
           auto newfree = gutils->getNewFromOriginal(orig->getArgOperand(0));
           auto tofree = gutils->invertPointerM(origfree, Builder2);
 
-          gutils->ComparePrimalAndShadow(Builder2, newfree, tofree);
+          Function *free = getOrInsertCheckedFree(
+              *orig->getModule(), orig->getFunctionType(),
+              orig->getCalledFunction(), newfree->getType(),
+              orig->getAttributes(), gutils->getWidth());
 
-          auto rule = [&](Value *tofree) {
-            SmallVector<Value *, 2> args = {tofree};
-            CallInst *CI = Builder2.CreateCall(orig->getFunctionType(),
-                                               orig->getCalledFunction(), args);
-            CI->setAttributes(orig->getAttributes());
-          };
+          SmallVector<Value *, 3> args;
+          args.push_back(newfree);
 
+          auto rule = [&args](Value *tofree) { args.push_back(tofree); };
           applyChainRule(Builder2, rule, tofree);
+
+          Builder2.CreateCall(free->getFunctionType(), free, args);
+
           return;
         }
       }
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -9883,9 +9883,7 @@ public:
           auto tofree = gutils->invertPointerM(origfree, Builder2);
 
           Function *free = getOrInsertCheckedFree(
-              *orig->getModule(), orig->getFunctionType(),
-              orig->getCalledFunction(), newfree->getType(),
-              orig->getAttributes(), gutils->getWidth());
+              *orig->getModule(), orig, newfree->getType(), gutils->getWidth());
 
           SmallVector<Value *, 3> args;
           args.push_back(newfree);
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -29,6 +29,7 @@
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/DerivedTypes.h"
 #include "llvm/IR/Value.h"
+#include "llvm/Transforms/Utils/BasicBlockUtils.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
 #include "DifferentialUseAnalysis.h"
@@ -440,13 +441,13 @@ public:
     constantval |= gutils->isConstantValue(&I);
 
     BasicBlock *parent = I.getParent();
-    Type *type = I.getType();
+    Type *type = gutils->getShadowType(I.getType());
 
     auto *newi = dyn_cast<Instruction>(gutils->getNewFromOriginal(&I));
 
     //! Store inverted pointer loads that need to be cached for use in reverse
     //! pass
-    if (!type->isEmptyTy() && !type->isFPOrFPVectorTy() &&
+    if (!I.getType()->isEmptyTy() && !I.getType()->isFPOrFPVectorTy() &&
         TR.query(&I).Inner0().isPossiblePointer()) {
       auto found = gutils->invertedPointers.find(&I);
       assert(found != gutils->invertedPointers.end());
@@ -1234,19 +1235,21 @@ public:
     IRBuilder<> Builder2(&SI);
     getForwardBuilder(Builder2);
 
+    Type *type = gutils->getShadowType(SI.getType());
+
     Value *dif1;
     Value *dif2;
 
     if (!constantval0) {
       dif1 = diffe(op1, Builder2);
     } else {
-      dif1 = Constant::getNullValue(SI.getType());
+      dif1 = Constant::getNullValue(type);
     }
 
     if (!constantval1) {
       dif2 = diffe(op2, Builder2);
     } else {
-      dif2 = Constant::getNullValue(SI.getType());
+      dif2 = Constant::getNullValue(type);
     }
 
     Value *diffe = Builder2.CreateSelect(cond, dif1, dif2);
@@ -8104,8 +8107,11 @@ public:
                                         tys),
               args));
 
-          Value *dif0 =
-              Builder2.CreateFDiv(diffe(orig->getArgOperand(0), Builder2), cal);
+          Value *op = diffe(orig->getArgOperand(0), Builder2);
+
+          auto rule = [&](Value *op) { return Builder2.CreateFDiv(op, cal); };
+
+          Value *dif0 = applyChainRule(call.getType(), Builder2, rule, op);
           setDiffe(orig, dif0, Builder2);
           return;
         }
@@ -8156,8 +8162,14 @@ public:
           Value *x = gutils->getNewFromOriginal(orig->getArgOperand(0));
           Value *onePx2 = Builder2.CreateFAdd(
               ConstantFP::get(x->getType(), 1.0), Builder2.CreateFMul(x, x));
-          Value *dif0 = Builder2.CreateFDiv(
-              diffe(orig->getArgOperand(0), Builder2), onePx2);
+
+          Value *op = diffe(orig->getArgOperand(0), Builder2);
+
+          auto rule = [&](Value *op) {
+            return Builder2.CreateFDiv(op, onePx2);
+          };
+
+          Value *dif0 = applyChainRule(call.getType(), Builder2, rule, op);
           setDiffe(orig, dif0, Builder2);
           return;
         }
@@ -8208,10 +8220,15 @@ public:
               Builder2.CreateCall(orig->getFunctionType(), callval, args));
           cubcall->setDebugLoc(gutils->getNewFromOriginal(orig->getDebugLoc()));
           cubcall->setCallingConv(orig->getCallingConv());
-          Value *dif0 = Builder2.CreateFDiv(
-              Builder2.CreateFMul(diffe(orig->getArgOperand(0), Builder2),
-                                  cubcall),
-              Builder2.CreateFMul(ConstantFP::get(x->getType(), 3), x));
+
+          Value *m = Builder2.CreateFMul(ConstantFP::get(x->getType(), 3), x);
+          Value *op = diffe(orig->getArgOperand(0), Builder2);
+
+          auto rule = [&](Value *op) {
+            return Builder2.CreateFDiv(Builder2.CreateFMul(op, cubcall), m);
+          };
+
+          Value *dif0 = applyChainRule(call.getType(), Builder2, rule, op);
           setDiffe(orig, dif0, Builder2);
           return;
         }
@@ -8268,9 +8285,13 @@ public:
               (funcName == "tanh") ? "cosh" : "coshf",
               called->getFunctionType(), called->getAttributes());
           auto cal = cast<CallInst>(Builder2.CreateCall(coshf, args));
-          Value *dif0 =
-              Builder2.CreateFDiv(diffe(orig->getArgOperand(0), Builder2),
-                                  Builder2.CreateFMul(cal, cal));
+
+          Value *m = Builder2.CreateFMul(cal, cal);
+          Value *op = diffe(orig->getArgOperand(0), Builder2);
+
+          auto rule = [&](Value *op) { return Builder2.CreateFDiv(op, m); };
+
+          Value *dif0 = applyChainRule(call.getType(), Builder2, rule, op);
           setDiffe(orig, dif0, Builder2);
           return;
         }
@@ -8321,9 +8342,12 @@ public:
           auto sinhf = gutils->oldFunc->getParent()->getOrInsertFunction(
               (funcName == "cosh") ? "sinh" : "sinhf",
               called->getFunctionType(), called->getAttributes());
-          auto cal = cast<CallInst>(Builder2.CreateCall(sinhf, args));
-          Value *dif0 =
-              Builder2.CreateFMul(diffe(orig->getArgOperand(0), Builder2), cal);
+          Value *cal = Builder2.CreateCall(sinhf, args);
+          Value *op = diffe(orig->getArgOperand(0), Builder2);
+
+          auto rule = [&](Value *op) { return Builder2.CreateFMul(op, cal); };
+
+          Value *dif0 = applyChainRule(call.getType(), Builder2, rule, op);
           setDiffe(orig, dif0, Builder2);
           return;
         }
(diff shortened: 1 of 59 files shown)


wsmoses:
I also noticed memset was missing for regular forward as well

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -7590,30 +7618,40 @@ class AdjointGenerator
               cal = Builder2.CreateFMul(cal, factor);
             }
 
-            Value *dfactor = (Mode == DerivativeMode::ForwardMode)
-                                 ? diffe(orig->getArgOperand(0), Builder2)
-                                 : diffe(orig, Builder2);
+            Value *dfactor = Mode != DerivativeMode::ForwardMode
```
wsmoses, line 7621: Can you make and != fwdsplit?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -8068,14 +8127,34 @@ class AdjointGenerator
...
-            Value *args[] = {vdiff, exponent};
+            auto rule = [&](Value *diff) {
+              if (auto vty = dyn_cast<VectorType>(diff->getType())) {
+#if LLVM_VERSION_MAJOR >= 12
+                unsigned ec = vty->getElementCount().getKnownMinValue();
+#else
+                unsigned ec = vty->getNumElements();
+#endif
+                Value *vdiff = UndefValue::get(vty);
+                for (int i = 0; i < ec; i++) {
```
wsmoses, line 8142: Is there a reason why this isn't already handled by the applyChainRule?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -8160,18 +8239,25 @@ class AdjointGenerator
...
 
           SmallVector<Value *, 2> args;
-#if LLVM_VERSION_MAJOR >= 14
-          for (unsigned i = 0; i < orig->arg_size(); ++i)
-#else
-          for (unsigned i = 0; i < orig->getNumArgOperands(); ++i)
-#endif
-          {
+
+          for (unsigned i = 0; i < orig->getNumArgOperands(); ++i) {
```
wsmoses, line 8243: Please restore the LLVM14 check (it hasn't been released yet / packed for Ubuntu in CI, but let's not break it until it does if we can).

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -8160,18 +8239,25 @@ class AdjointGenerator
...
             auto arg = orig->getArgOperand(i);
             args.push_back(gutils->getNewFromOriginal(arg));
           }
-          CallInst *CI = Builder2.CreateCall(orig->getFunctionType(),
-                                             orig->getCalledFunction(), args);
-          CI->setAttributes(orig->getAttributes());
+
+          Value *CI;
+          if (gutils->getWidth() > 1) {
+            CI = applyChainRule(call.getType(), Builder2, [&]() {
```
wsmoses, line 8250: Can this be done without the width check?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -8440,14 +8526,19 @@ class AdjointGenerator
...
           auto tofree = gutils->invertPointerM(origfree, Builder2);
-          if (tofree != origfree) {
-            SmallVector<Value *, 2> args = {tofree};
-            CallInst *CI = Builder2.CreateCall(orig->getFunctionType(),
-                                               orig->getCalledFunction(), args);
-            CI->setAttributes(orig->getAttributes());
-          }
+
+          auto rule = [&](Value *tofree) {
+            if (tofree != origfree) {
```
wsmoses, line 8531: Have the whole rule / applyChain inside the if?

wsmoses: Also the more I think about this code, I think this would be better as a runtime rather than compile time check. In essence the issue here is that we permit the primal version of a value to be used in a shadow (e.g. stored into shadow data structures/etc). This is necessary to maintain some of the same structure as the original code. However, this may result in multiple copies (and in the case of vector mode width extra many copies). A value check helps but does not completel resolve the problem since the LLVM values may still be distinct even if the runtime data is the same from this.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -8656,8 +8751,10 @@ class AdjointGenerator
       if (subretused && subretType != DIFFE_TYPE::CONSTANT) {
         primal = Builder2.CreateExtractValue(diffes, 0);
         diffe = Builder2.CreateExtractValue(diffes, 1);
-      } else if (!FT->getReturnType()->isVoidTy()) {
+      } else if (subretType != DIFFE_TYPE::CONSTANT) {
         diffe = diffes;
+      } else if (!FT->getReturnType()->isVoidTy()) {
+        primal = diffes;
```
wsmoses, line 8757: It's odd we haven't hit a bug here before.

enzyme/test/Enzyme/ForwardMode/Faddeeva_erfi.ll
```diff
@@ -1,4 +1,4 @@
-; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -mem2reg -sroa -instsimplify -simplifycfg -adce -S | FileCheck %s
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -mem2reg -instsimplify -instcombine -simplifycfg -adce -S | FileCheck %s
```
wsmoses, line 1: Do we need instcombine here, if possible we should avoid because its impact is very diferent from version to version (which makes it a great optimization, but bad for a test).

wsmoses (approved):
Address comments from review, then looks good to me.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: yes

enzyme/Enzyme/AdjointGenerator.h
tgymnich: merge error ...

enzyme/Enzyme/AdjointGenerator.h
tgymnich: no

enzyme/Enzyme/AdjointGenerator.h
tgymnich: inverted and added fwdsplit

enzyme/Enzyme/AdjointGenerator.h
tgymnich: fixed

enzyme/test/Enzyme/ForwardMode/Faddeeva_erfi.ll
tgymnich: fixed

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -8569,14 +8638,20 @@ class AdjointGenerator
           getForwardBuilder(Builder2);
           auto origfree = orig->getArgOperand(0);
           auto tofree = gutils->invertPointerM(origfree, Builder2);
+
           if (tofree != origfree) {
```
wsmoses, line 8642: Okay I thought about this even more and the reason this is unsafe for regular AD makes this incredibly danger for split. Specifically suppose we allocate memory for one reverse, which we then splat into a width number vector. We would end up calling free on that same one width number of times. I do really think we need to add a runtime check.

Also as a note the origfree should be getNewFromOriginal

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -8813,15 +8884,40 @@ class AdjointGenerator
           IRBuilder<> Builder2(&call);
           getForwardBuilder(Builder2);
           auto origfree = orig->getArgOperand(0);
+          auto newfree = gutils->getNewFromOriginal(orig->getArgOperand(0));
           auto tofree = gutils->invertPointerM(origfree, Builder2);
-          if (tofree != origfree) {
+          auto newcall = gutils->getNewFromOriginal(orig);
+
+          auto rule = [&](Value *tofree) {
+            Value *ptrDiff = Builder2.CreatePtrDiff(tofree, newfree);
```
wsmoses, line 8892: Why not an icmp ne?

tgymnich: > Why not an icmp ne?
`CreateIsNotNull` felt shorter...


```
  Value *CreateIsNotNull(Value *Arg, const Twine &Name = "") {
    return CreateICmpNE(Arg, Constant::getNullValue(Arg->getType()),
                        Name);
  }
```

wsmoses: https://llvm.org/docs/LangRef.html#icmp-instruction

It should accept pointer types

tgymnich: you are right.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2365,16 +2367,17 @@ const AugmentedReturn &EnzymeLogic::CreateAugmentedPrimal(
 
 void createTerminator(TypeResults &TR, DiffeGradientUtils *gutils,
                       BasicBlock *oBB, DIFFE_TYPE retType, ReturnType retVal) {
+  ReturnInst *inst = dyn_cast_or_null<ReturnInst>(oBB->getTerminator());
+  // In forward mode we only need to update the return value
+  if (inst == nullptr)
+    return;
 
-  BasicBlock *nBB = cast<BasicBlock>(gutils->getNewFromOriginal(oBB));
+  ReturnInst *newInst = dyn_cast<ReturnInst>(gutils->getNewFromOriginal(inst));
```
wsmoses, line 2375: We should just do a cast rather than dyn_cast here.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2365,16 +2367,17 @@ const AugmentedReturn &EnzymeLogic::CreateAugmentedPrimal(
 
 void createTerminator(TypeResults &TR, DiffeGradientUtils *gutils,
                       BasicBlock *oBB, DIFFE_TYPE retType, ReturnType retVal) {
+  ReturnInst *inst = dyn_cast_or_null<ReturnInst>(oBB->getTerminator());
```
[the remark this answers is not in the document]
tgymnich: Yes you are right, every basic block has to have a terminator. 👍

enzyme/Enzyme/EnzymeLogic.cpp
wsmoses: Yes, but we already checked the original instruction as a return. Should we not have the return here as well?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -9785,15 +9872,24 @@ class AdjointGenerator
...
           auto tofree = gutils->invertPointerM(origfree, Builder2);
-          if (tofree != origfree) {
-            SmallVector<Value *, 2> args = {tofree};
-            CallInst *CI = Builder2.CreateCall(orig->getFunctionType(),
-                                               orig->getCalledFunction(), args);
-            CI->setAttributes(orig->getAttributes());
-          }
+
+          Function *free = getOrInsertCheckedFree(
+              *orig->getModule(), orig->getFunctionType(),
```
wsmoses, line 9879: Can you instead pass a width argument to the insert checked free, and then pass all of the unwrapped as arguments?

That way we can check if a vector width shadow is the same as another vector width shadow?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -9785,15 +9872,24 @@ class AdjointGenerator
...
+              orig->getCalledFunction(), newfree->getType(),
+              orig->getAttributes(), gutils->getWidth());
+
+          SmallVector<Value *, 3> args;
+          args.push_back(newfree);
+
+          auto rule = [&args](Value *tofree) { args.push_back(tofree); };
+          applyChainRule(Builder2, rule, tofree);
+
+          Builder2.CreateCall(free->getFunctionType(), free, args);
```
wsmoses, line 9889: Can you add the debug info of the original function, and also add the calling convention?

enzyme/Enzyme/Utils.cpp
```diff
@@ -298,6 +298,71 @@ Function *getOrInsertDifferentialFloatMemmove(Module &M, Type *T,
...
+
+  BasicBlock *entry = BasicBlock::Create(M.getContext(), "entry", F);
+  BasicBlock *body = BasicBlock::Create(M.getContext(), "body", F);
+  BasicBlock *end = BasicBlock::Create(M.getContext(), "end", F);
+
+  IRBuilder<> EntryBuilder(entry);
+  IRBuilder<> BodyBuilder(body);
+  IRBuilder<> EndBuilder(end);
+
+  Value *checkResult = EntryBuilder.getTrue();
```
wsmoses, line 336: Are you freeing the original?

enzyme/Enzyme/Utils.cpp
```diff
@@ -298,6 +298,71 @@ Function *getOrInsertDifferentialFloatMemmove(Module &M, Type *T,
...
+    checkResult = EntryBuilder.CreateAnd(isNotEqual, checkResult);
+
+    if (i < width - 1) {
+      Argument *nextShadow = F->arg_begin() + i + 2;
+      Value *isNotEqual = EntryBuilder.CreateICmpNE(shadow, nextShadow);
+      checkResult = EntryBuilder.CreateAnd(isNotEqual, checkResult);
+    }
+
+    Value *args[] = {shadow};
+    CallInst *CI = BodyBuilder.CreateCall(FreeTy, Free, args);
```
wsmoses, line 355: Doesn't the call need to be protected by the condition?

tgymnich: it should be.

enzyme/Enzyme/AdjointGenerator.h
tgymnich: done

enzyme/Enzyme/Utils.cpp
tgymnich: I am not removing the original instruction, so ... yes.

enzyme/test/Enzyme/ForwardModeVector/ptr-eq.ll
```diff
@@ -0,0 +1,83 @@
...
+; CHECK-NEXT:   %6 = and i1 %5, %4
+; CHECK-NEXT:   %7 = icmp ne i8* %0, %2
+; CHECK-NEXT:   %8 = and i1 %7, %6
+; CHECK-NEXT:   %9 = icmp ne i8* %2, %3
+; CHECK-NEXT:   %10 = and i1 %9, %8
+; CHECK-NEXT:   %11 = icmp ne i8* %0, %3
+; CHECK-NEXT:   %12 = and i1 %11, %10
+; CHECK-NEXT:   br i1 %12, label %end, label %body
+
+; CHECK: body:                                             ; preds = %entry
```
wsmoses, line 75: The structure here is still a bit off.

I think we need something like:
```
br primal == shadow0, return, free0

free0
  free(shadow0)
  br shadow0 == shadow1, return, free1

free1:
  free(shadow1)
  free(shadow2)
  ....
  br return
return:
  ret void
``

File: enzyme/test/Enzyme/ForwardModeVector/ptr-eq.ll [hunk 17]
```diff
@@ -0,0 +1,77 @@
...
+; CHECK-NEXT:   ret void
+; CHECK-NEXT: }
+
+; CHECK: define internal void @__enzyme_checked_free_3(i8* nocapture %0, i8* nocapture %1, i8* nocapture %2, i8* nocapture %3)
+; CHECK-NEXT: entry:
+; CHECK-NEXT:   %4 = icmp ne i8* %0, %1
+; CHECK-NEXT:   br i1 %4, label %end, label %free0
+
+; CHECK: free0:                                            ; preds = %entry
+; CHECK-NEXT:   %5 = icmp ne i8* %2, %3
```
wsmoses, line 68: Isn't this missing freeing (conditionally) %1 and %3?

enzyme/test/Enzyme/ForwardModeVector/ptr-eq.ll
```diff
@@ -65,6 +65,7 @@ entry:
 ; CHECK-NEXT:   br i1 %4, label %end, label %free0
```
wsmoses, line 65: The comparison is the wrong way (it should free if they arent equal)

enzyme/test/Enzyme/ForwardModeVector/ptr-eq.ll
```diff
@@ -0,0 +1,79 @@
...
+
+; CHECK: define internal void @__enzyme_checked_free_3(i8* nocapture %0, i8* nocapture %1, i8* nocapture %2, i8* nocapture %3)
+; CHECK-NEXT: entry:
+; CHECK-NEXT:   %4 = icmp ne i8* %0, %1
+; CHECK-NEXT:   br i1 %4, label %free0, label %end
+
+; CHECK: free0:                                            ; preds = %entry
+; CHECK-NEXT:   call void @free(i8* %1)
+; CHECK-NEXT:   %5 = icmp ne i8* %2, %3
+; CHECK-NEXT:   br i1 %5, label %free1, label %end
```
wsmoses, line 70: Shouldn't this comparison check %1 (first shadow) against %2? E.g. what if you had vector size 2?

wsmoses:
Fix the C++/LLVM tests then good to merge!

PR #490, opened 2022-02-05
Forward mode: memset

tgymnich:


wsmoses (approved):
Looks good, add a test?

And add a vector mode assertion here (or add support for)?

PR #495, opened 2022-02-06
Document the most important functions

ZuseZ4:
Just to get started, should be improved further.
Feel free to update this branch directly.

### DIFF
--- a/enzyme/Enzyme/CApi.h
+++ b/enzyme/Enzyme/CApi.h
@@ -117,6 +117,20 @@ typedef enum {
   DEM_ForwardModeSplit = 4,
 } CDerivativeMode;
 
+
+/// Generates a new function based on the input, which uses forward-mode AD.
+///
+/// @param mode: DEM_ForwardMode and DEM_ForwardModeSplit
+/// @param retType: DFT_DUP_ARG, DFT_CONSTANT 
+/// @param constant_args: pointing to combinations of DFT_DUP_ARG and DFT_CONSTANT
+/// @param width: integer n >= 1. The generated functions expects n additional inputs
+///      for each input marked as DFT_DUP_ARG.
+/// @param returnValue: Whether to return the primary return value.
+///      dret_used: Not available here / implicitely set to true, as it is 
+///      the only place where we return gradient information.
+/// @param uncacheable_args_size: Should be set to 0, as long as using DEM_ReverseModeCombined.
+///      Otherwise, it should be set to the amount of input params of `todiff`, which might 
+///      change between the forward and the reverse pass.
 LLVMValueRef EnzymeCreateForwardDiff(
     EnzymeLogicRef, LLVMValueRef todiff, CDIFFE_TYPE retType,
     CDIFFE_TYPE *constant_args, size_t constant_args_size,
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/CApi.h
```diff
@@ -117,13 +117,42 @@ typedef enum {
   DEM_ForwardModeSplit = 4,
 } CDerivativeMode;
 
+/// Generates a new function based on the input, which uses forward-mode AD.
+///
```
wsmoses, line 121: We also need to document the formal function: https://enzyme.mit.edu/doxygen/classEnzymeLogic.html#a8e6ba5c54950f1ebcdfcbd7ce2a16083

I'm slightly inclined to have the C API documentation be light, and instead forward one to see the function which it wraps.

enzyme/Enzyme/CApi.h
```diff
@@ -117,13 +117,42 @@ typedef enum {
...
     EnzymeTypeAnalysisRef TA, uint8_t returnValue, CDerivativeMode mode,
     unsigned width, LLVMTypeRef additionalArg, struct CFnTypeInfo typeInfo,
     uint8_t *_uncacheable_args, size_t uncacheable_args_size);
 
+/// Generates a new function based on the input, which uses reverse-mode AD.
+///
+/// Based on the
+/// @param retType: When returning f32/f64 we might use DFT_OUT_DIFF.
+///      When returning anything else, one should use DFT_CONSTANT
+/// @param width: currently only supporting width=1 here
```
wsmoses, line 146: I wouldn't add this as a comment here, and instead describe what width should do.

enzyme/Enzyme/CApi.h
```diff
@@ -117,13 +117,42 @@ typedef enum {
...
     EnzymeLogicRef, LLVMValueRef todiff, CDIFFE_TYPE retType,
     CDIFFE_TYPE *constant_args, size_t constant_args_size,
     EnzymeTypeAnalysisRef TA, uint8_t returnValue, CDerivativeMode mode,
     unsigned width, LLVMTypeRef additionalArg, struct CFnTypeInfo typeInfo,
     uint8_t *_uncacheable_args, size_t uncacheable_args_size);
 
+/// Generates a new function based on the input, which uses reverse-mode AD.
+///
+/// Based on the
+/// @param retType: When returning f32/f64 we might use DFT_OUT_DIFF.
```
wsmoses, line 144: I would use llvm type names here: float/double. We also would want to state the general rule that out_diff is used when the value being differentiated is being passed as a register (examples being float double, a vector of these, an llvm.struct of these, etc) rather than through memory (e.g. a pointer)

enzyme/Enzyme/CApi.h
```diff
@@ -133,6 +162,17 @@ LLVMValueRef EnzymeCreatePrimalAndGradient(
...
 
+/// Generates an augmented forward function based on the input.
+///
+/// Cached information will be stored on a Tape.
+/// Will usually be used in combination with EnzymeCreatePrimalAndGradient for
+/// reverse-mode-split AD.
+/// @param retType: DFT_DUP_ARG, DFT_CONSTANT
+/// @param constant_args: pointing to combinations of DFT_DUP_ARG and
+/// DFT_CONSTANT
+/// @param forceAnonymousTape: TODO
```
wsmoses, line 173: Per https://enzyme.mit.edu/doxygen/classEnzymeLogic.html#a8e6ba5c54950f1ebcdfcbd7ce2a16083, `forceAnonymousTape forces the tape to be an i8* rather than the true tape`.

PR #502, opened 2022-02-07
Cancel benchmarks on push

tgymnich:


wsmoses:
This inadvertently cancels main, can you exclude main from the cancellation

tgymnich:
> This inadvertently cancels main, can you exclude main from the cancellation

#512

PR #515, opened 2022-02-09
Fix JLL buildscript

vchuravy:
Mirror https://github.com/JuliaPackaging/Yggdrasil/pull/4417


wsmoses:
The macos julia nightly segfault should be investigated

PR #519, opened 2022-02-11
Merge rust build system

ZuseZ4:
Adds the rust build repo which is published as crate under https://crates.io/crates/enzyme
Should allow further CI automation.


ZuseZ4:
I think when squashing during merge the commits from @bytesnake will show up as my own, I guess this can be fixed via using a merge-commit? I can squash my lastest 35 commits into one then, to reduce noise. 
Also, I would expect that the current Apache2/MIT license already covers Apache License v2.0 with LLVM Exceptions, 
but I'm not sure, so let's also wait to get that formally right.

.packaging/rust/src/bin/main.rs
```diff
@@ -0,0 +1,39 @@
...
+fn get_enzyme() -> Result<(), String> {
+    let mut cmd_line = std::env::args();
+    if cmd_line.len() == 2 {
+        let arg = cmd_line
+            .nth(1)
+            .expect("failed reading")
+            .parse::<Repo>()
+            .expect("failed reading first argument!");
+
+        enzyme::download(arg.clone())?;
```
tgymnich, line 18: there should probably be an option to just get enzyme from ../../../../enzyme.

wsmoses: +1

.packaging/rust/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+authors = [
+	"Manuel Drehwald",
+	"Lorenz Schmidt"
+]
+repository = "https://github.com/rust-ml/enzyme_build"
+keywords = ["enzyme", "auto-diff", "llvm", "machine-learning", "scientific-computing"]
+categories = ["development-tools::cargo-plugins", "mathematics", "science", "simulation"]
+readme = "README.md"
+documentation = "https://docs.rs/enzyme/"
+homepage = "https://github.com/rust-ml/enzyme_build"
```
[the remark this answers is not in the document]
ZuseZ4: Left the docs site, since this one is auto-generated by the rust infra and we want the crate specific instructions.
Changed the homepage for enzyme.mit.edu, since the /rust isn't active yet.
I can adjust that later once we finished all the transitions.

.packaging/rust/src/bin/main.rs
ZuseZ4: Added.

ZuseZ4:
```
enzyme-install
error: The following required arguments were not provided:
    <--rust-stable|--rust-head|--rust-local <RUST_LOCAL>>
    <--enzyme-stable|--enzyme-head|--enzyme-local <ENZYME_LOCAL>>

USAGE:
    enzyme-install <--rust-stable|--rust-head|--rust-local <RUST_LOCAL>> <--enzyme-stable|--enzyme-head|--enzyme-local <ENZYME_LOCAL>>
```

.packaging/rust/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "enzyme"
+version = "0.4.0"
+edition = "2021"
+license = "Apache-2.0 WITH LLVM-Exceptions"
+authors = [
+	"Manuel Drehwald",
+	"Lorenz Schmidt"
+]
+repository = "https://github.com/rust-ml/enzyme_build"
```
wsmoses, line 10: Is the URL still right?

.packaging/rust/src/code/generate_api.rs
```diff
@@ -0,0 +1,94 @@
...
+        .allowlist_function("EnzymeSetCLInteger")
+        .allowlist_function("CreateTypeAnalysis")
+        .allowlist_function("ClearTypeAnalysis")
+        .allowlist_function("FreeTypeAnalysis")
+        .allowlist_function("CreateEnzymeLogic")
+        .allowlist_function("ClearEnzymeLogic")
+        .allowlist_function("FreeEnzymeLogic")
+        .allowlist_type("LLVMOpaqueModule")
+        .allowlist_function("EnzymeCreatePrimalAndGradient")
+        .allowlist_function("EnzymeCreateAugmentedPrimal")
```
wsmoses, line 62: add the CreateForward while we're at it?

wsmoses: Also could this just parse the whole header, or does it need the whitelist?

ZuseZ4: I needed it in the past when we introduced C++ classes to the file, but now I handle them fine, so I removed it.

Without a whitelist bindgen creates bindings for everything it sees, so instead of a 300 LoC file it now creates a 9k LoC file covering all kind of LLVM functions like "LLVMIntrinsicIsOverloaded" and globals like "pub const __glibc_c99_flexarr_available: u32 = 1;" beside of the Enzyme bindings. I already use a specific crate offering nicer bindings to LLVM, so I don't need those generated ones. 

There isn't a real damage in having (and ignoring) those generated LLVM bindings and without the whitelist it should be slightly less maintenance work, so I guess that's a small win.

ZuseZ4:
I guess it makes sense to merge the release CI updates with some build tests which I should also add to CI, so I prefer to bundle those all into a follow-up PR.

PR #531, opened 2022-02-22
Use multiple threads for Julia ci

tgymnich:


wsmoses:
I don't think this will necessarily run the tests in parallel so much as enable two threads in the JULIA runtime (for example, in the Threads test).

@vchuravy

tgymnich:
I just checked and what I saw was a feature of CUDA.jl and not of Test.
```
┌ Warning: You are running the CUDA.jl test suite with only a single thread; this will take a long time.
│ Consider launching Julia with `--threads auto` to run tests in parallel.
└ @ Main ~/.julia/packages/CUDA/KnJGx/test/runtests.jl:52
```

PR #543, opened 2022-03-04
Prepare folder structure for forward benchmarks

tgymnich:
Creare subfolders for forward and reverse mode benchmarks.


wsmoses:
A lot of the benchmarks that are moved appear to be recreated files/deleted, can you use git mv to preserve history?

PR #554, opened 2022-03-06
Forward bench

tgymnich:


.github/workflows/benchmark.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         llvm: ["7",  "8", "9", "10", "11", "12", "13"] #, "14"]
         build: ["Release", "Debug"] # "RelWithDebInfo"
         os: [self-hosted]
-    timeout-minutes: 120 
+    timeout-minutes: 300
```
wsmoses, line 26: We can probably split bench forward/reverse as two separate things (and I should start more CI)

tgymnich: I would disable the fwd tests by default. At some point we need to decide on a subset of test cases that run in reasonable time.

enzyme/benchmarks/CMakeLists.txt
```diff
@@ -46,3 +46,5 @@ add_lit_testsuite(bench-enzyme "Running enzyme benchmarks tests"
 set_target_properties(bench-enzyme PROPERTIES FOLDER "bench Tests")
```
wsmoses, line 47: Can the file structure here be changed to have only one copy of the benchmark data?

wsmoses (commented):
Can you remove the intermediate ll files

PR #564, opened 2022-03-14
Lld enzyme

ZuseZ4:
As discussed, these changes should fix the usage of Enzyme as an lld linker plugin.
Beside of CI I have tested it compiling with 
```
Running: `"cmake" "-G" "Ninja" ".." "-DCMAKE_BUILD_TYPE=Release" "-DENZYME_EXTERNAL_SHARED_LIB=ON" "-DLLVM_DIR=/home/zusez4/.cache/enzyme/rustc-1.59.0-src/build/x86_64-unknown-linux-gnu/llvm/build/lib/cmake/llvm" "-DLLVM_EXTERNAL_LIT=/home/zusez4/.cache/enzyme/rustc-1.59.0-src/src/llvm-project/llvm/utils/lit/lit.py"
```
It fixed the undefined symbol error for me.


tgymnich (commented):
It would probably be nice if you could add LLDEnzyme to `EnzymeConfig.cmake.in`, so that it can be found when somebody is using the Enzyme CMake package.

PR #594, opened 2022-03-25
Add context

vchuravy:


vchuravy:
For #590 we would need the improvement that we pass a call stack so that we can give a proper backtrace instead of a truncated one.

PR #598, opened 2022-04-07
Update rust

ZuseZ4:
- Updating Enzyme version
- Re-introducing specific bindings. C Enums aren't handled nicely by default so it's worth to maintain the allowlist.


wsmoses:
Worth double checking the ABI hasn't changed.

wsmoses (approved):

ZuseZ4:
For some time it's useful to have both rust branches, but no intention to merge this, so closing.

PR #600, opened 2022-04-07
Don't overflow bit size

wsmoses:


vchuravy (approved):
Doea fix the immediate issue, but now leads to a type analysis limit

wsmoses:
Explain a bit more?

PR #606, opened 2022-04-11
handle more IO

ZuseZ4:
With this PR we should (mostly) recognize the functions from the following modules as inactive.
I ran tests with the rust-build clang, as well as the compiler explorer / CI. 
A few functions are currently blocked on two probably minor bugs and I probably missed a few ones.

- [x] istream [std::cin](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#afdcd7ecbf1544ef3f79b89f0dd06c3b7)

- [x] ostream [std::cout](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#aaf93fdf0812752e0e02c501dea1b38f0)

- [x] ostream [std::cerr](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#a7431d56d1e8cd7a9b854171294bd71c7)

- [x] wostream [std::wcout](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#a2f5fc307ed84f0ecfbc36d2cda322040)

- [ ] wistream [std::wcin](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#ab1807d3145162e06150b063da4da2707)

- [ ] wostream [std::wcerr](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#ae50854b2a6629e6504846f8aff472e7e)

- [ ] ostream [std::clog](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#a7e2a2fc4b5924b7292c0566ca4c95463)

- [ ] wostream [std::wclog](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-api-4.5/a01118.html#a09e3edb1609f2a7fb18370eab59ba8dc)

I further marked a few functions from
locale [std::locale](https://en.cppreference.com/w/cpp/locale/locale)

### DIFF
--- a/enzyme/Enzyme/ActivityAnalysis.cpp
+++ b/enzyme/Enzyme/ActivityAnalysis.cpp
@@ -1187,10 +1187,12 @@ bool ActivityAnalyzer::isConstantValue(TypeResults &TR, Value *Val) {
 
           // If requesting empty unknown functions to be considered inactive,
           // abide by those rules
-          if (!isCertainCacheable(called) && called->empty() &&
+          if (EnzymeEmptyFnInactive && called->empty() &&
               !hasMetadata(called, "enzyme_gradient") &&
               !hasMetadata(called, "enzyme_derivative") &&
-              !isa<IntrinsicInst>(op) && EnzymeEmptyFnInactive) {
+              !isAllocationFunction(*called, TLI) &&
+              !isDeallocationFunction(*called, TLI) &&
+              !isa<IntrinsicInst>(op)) {
             InsertConstantValue(TR, Val);
             insertConstantsFrom(TR, *UpHypothesis);
             return true;
@@ -1867,7 +1869,7 @@ bool ActivityAnalyzer::isInstructionInactiveFromOrigin(TypeResults &TR,
 
       // If requesting empty unknown functions to be considered inactive, abide
       // by those rules
-      if (!isCertainCacheable(called) && called->empty() &&
+      if (!isCertainPrintMallocOrFree(called) && called->empty() &&
           !hasMetadata(called, "enzyme_gradient") &&
           !hasMetadata(called, "enzyme_derivative") &&
           !isa<IntrinsicInst>(op) && EnzymeEmptyFnInactive) {
@@ -1950,7 +1952,7 @@ bool ActivityAnalyzer::isInstructionInactiveFromOrigin(TypeResults &TR,
       if (!ci->onlyAccessesArgMemory() && !ci->doesNotAccessMemory()) {
 
         Function *called = getFunctionFromCall(ci);
-        if (!called || (!isCertainCacheable(called) &&
+        if (!called || (!isCertainPrintMallocOrFree(called) &&
                         !isMemFreeLibMFunction(called->getName()))) {
           if (EnzymePrintActivity)
             llvm::errs() << "nonconstant(" << (int)directions << ")  up-global "
--- a/enzyme/Enzyme/ActivityAnalysis.cpp
+++ b/enzyme/Enzyme/ActivityAnalysis.cpp
@@ -82,18 +82,21 @@ const char *KnownInactiveFunctionsStartingWith[] = {
     "_ZN4core3fmt", "_ZN3std2io5stdio6_print", "f90io", "$ss5print",
     "_ZNSt7__cxx1112basic_string", "_ZNSt7__cxx1118basic_string",
     // ostream generic <<
-    "_ZStlsISt11char_traitsIcEERSt13basic_ostreamIcT_ES5_",
-    "_ZSt16__ostream_insert", "_ZNSo9_M_insert",
+    "_ZStlsISt11char_traitsIcEERSt13basic_ostream", "_ZSt16__ostream_insert",
+    "_ZNSo9_M_insert",
+    // ostream wchar
+    "_ZNSt13basic_ostreamIwSt11char_traits",
+    "_ZStlsIwSt11char_traitsIwEERSt13basic_ostream",
     // ostream put
     "_ZNSo3put",
-    // std::istream init, widen_init, get, getline, >>, sync, ignore
+    // std::istream: init, widen_init, get, getline, >>, sync, ignore
     "_ZNSt8ios_base4InitC1Ev", "_ZNKSt5ctypeIcE13_M_widen_initEv", "_ZNSi3get",
     "_ZNSi7getline", "_ZNSirsER", "_ZNSt7__cxx1115basic_stringbuf",
     "_ZNSi6ignoreEv",
-    // std::ios_base::ios_base() ==  "_ZNSt8ios_baseC2Ev",
+    // std::ios_base
     "_ZNSt8ios_base", "_ZNSt9basic_ios",
-    // std::local::local() ==   "_ZNSt6localeC1Ev",
-    "_ZNSt6locale",
+    // std::local
+    "_ZNSt6locale", "_ZNKSt6locale4name",
     // init
     "_ZNSt9basic_iosIcSt11char_traitsIcEE4init",
     // std::cout
@@ -118,6 +121,7 @@ const std::set<std::string> InactiveGlobals = {
     "stdin",
     "_ZSt3cin",
     "_ZSt4cout",
+    "_ZSt5wcout",
     "_ZSt4cerr",
     "_ZTVNSt7__cxx1115basic_stringbufIcSt11char_traitsIcESaIcEEE",
     "_ZTVSt15basic_streambufIcSt11char_traitsIcEE",
--- a/enzyme/Enzyme/ActivityAnalysis.cpp
+++ b/enzyme/Enzyme/ActivityAnalysis.cpp
@@ -79,35 +79,53 @@ cl::opt<bool>
 #include <unordered_map>
 
 const char *KnownInactiveFunctionsStartingWith[] = {
-    "_ZN4core3fmt", "_ZN3std2io5stdio6_print", "f90io", "$ss5print",
-    "_ZNSt7__cxx1112basic_string", "_ZNSt7__cxx1118basic_string",
+    "_ZN4core3fmt",
+    "_ZN3std2io5stdio6_print",
+    "f90io",
+    "$ss5print",
+    "_ZNSt7__cxx1112basic_string",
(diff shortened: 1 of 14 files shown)


enzyme/test/Integration/ReverseMode/cin.cpp
```diff
@@ -0,0 +1,40 @@
+// RUN: %clang++ -std=c++11 -fno-exceptions -ffast-math -O0 %s -S -emit-llvm -o - | %opt - %loadEnzyme -enzyme -S
+// RUN: %clang++ -std=c++11 -fno-exceptions -ffast-math -O1 %s -S -emit-llvm -o - | %opt - %loadEnzyme -enzyme -S
```
wsmoses, line 2: You may be able to use `https://en.cppreference.com/w/cpp/io/basic_istream/swap` to swap cin with a stringbuffer stream which you have the input you want already prepared

ZuseZ4: done

ZuseZ4:
It seems like CI mangles names differently from my local clang (shipped with Rust).
Rust afaik usually applies a few minor patches, does it make sense that they affect the mangling?
In that case I would need to re-run all my experiments with the unmodified clang..

enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -104,6 +116,20 @@ const std::set<std::string> InactiveGlobals = {
     "stderr",
     "stdout",
     "stdin",
+    "_ZSt3cin",
+    "_ZSt4cout",
```
wsmoses, line 120: Add cerr?

ZuseZ4: Seem to already work out of the box with these settings, but added a test.

ZuseZ4: done

wsmoses (approved):

ZuseZ4:
renaming `isCertainPrintMallocOrFree` to `isCertainPrintReadMallocOrFree` sounded like a bit to much, so renamed to `isCertainCacheable`. Also added `_ZNSaIcE` to cover _ZNSaIcED1Ev and _ZNSaIcEC1Ev, std::allocator<char>, as required by CI.

enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -1136,7 +1187,7 @@ bool ActivityAnalyzer::isConstantValue(TypeResults &TR, Value *Val) {
 
           // If requesting empty unknown functions to be considered inactive,
           // abide by those rules
```
wsmoses, line 1189: This no longer appears to be correct since a cacheable function doesn't necessarily mean it is (in)active and vice versa.

wsmoses (changes requested):

enzyme/Enzyme/ActivityAnalysis.cpp
ZuseZ4: Rename from `isCertainCacheable` to `isCertainInactiveAndCacheable`, or do you have another suggestion?
At least with the function I added here I think it currently still seems to hold and we can add another function / logic later, once we consider more functions to cover all 2*2 combinations?

wsmoses: I mean this location solely cares about activity, nothing about whether or not it is cacheable.

Regardless, here it should be safe to just change this to:

```
          if (EnzymeEmptyFnInactive && called->empty() &&
              !hasMetadata(called, "enzyme_gradient") &&
              !hasMetadata(called, "enzyme_derivative") &&
              !isAllocationFunction(called) && !isDeallocationFunction(called) &&
              !isa<IntrinsicInst>(op)) {
 ```
 
 In essence what this code is saying is if one chooses to enable the "all empty functions are inactive by assumption" flag, we still should mark allocation/free functions as being potentially able to create active memory.

File: enzyme/Enzyme/ActivityAnalysis.cpp [hunk 4]
```diff
@@ -1816,7 +1867,7 @@ bool ActivityAnalyzer::isInstructionInactiveFromOrigin(TypeResults &TR,
 
       // If requesting empty unknown functions to be considered inactive, abide
       // by those rules
-      if (!isCertainPrintMallocOrFree(called) && called->empty() &&
+      if (!isCertainCacheable(called) && called->empty() &&
```
wsmoses, line 1870: Same applies here.

enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -1899,7 +1950,7 @@ bool ActivityAnalyzer::isInstructionInactiveFromOrigin(TypeResults &TR,
       if (!ci->onlyAccessesArgMemory() && !ci->doesNotAccessMemory()) {
 
         Function *called = getFunctionFromCall(ci);
-        if (!called || (!isCertainPrintMallocOrFree(called) &&
+        if (!called || (!isCertainCacheable(called) &&
```
wsmoses, line 1953: This code says that a call to a memory free libm function, allocation, free, or print, doesn't change the activity of a global.

This code doesn't have anything to do with cacheability.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -396,7 +396,7 @@ struct CacheAnalysis {
       return {};
     }
 
-    if (isCertainPrintMallocOrFree(Fn) || isAllocationFunction(*Fn, TLI)) {
+    if (isCertainCacheable(Fn) || isAllocationFunction(*Fn, TLI)) {
```
wsmoses, line 399: Again this code says that a print/malloc/free function doesn't overwrite a potential load/call/etc and isn't a question about whether the function is cacheable.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -460,7 +460,7 @@ struct CacheAnalysis {
       // Don't consider modref from malloc/free as a need to cache
       if (auto obj_op = dyn_cast<CallInst>(inst2)) {
         Function *called = getFunctionFromCall(obj_op);
-        if (called && isCertainPrintMallocOrFree(called)) {
+        if (called && isCertainCacheable(called)) {
```
wsmoses, line 463: Same comment as above.

enzyme/Enzyme/Utils.cpp
```diff
@@ -1214,7 +1214,7 @@ bool writesToMemoryReadBy(llvm::AAResults &AA, llvm::Instruction *maybeReader,
   using namespace llvm;
   if (auto call = dyn_cast<CallInst>(maybeWriter)) {
     Function *called = getFunctionFromCall(call);
-    if (called && isCertainPrintMallocOrFree(called)) {
+    if (called && isCertainCacheable(called)) {
```
wsmoses, line 1217: Same comment as above.

enzyme/Enzyme/ActivityAnalysis.cpp
wsmoses: Most uses of the function are actually as the name implies, a check whether it is a free, allocation, or print, and not a question of cacheability.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -10990,6 +10990,8 @@ class AdjointGenerator
         llvm::errs() << *gutils->newFunc->getParent() << "\n";
         llvm::errs() << " orig: " << *orig << " callval: " << *callval << "\n";
       }
+      llvm::errs() << callval->getName() << "\n";
```
wsmoses, line 10993: Remove this print?

enzyme/Enzyme/ActivityAnalysis.cpp
ZuseZ4: Updated this if statement and undone the renaming on other locations.

enzyme/Enzyme/ActivityAnalysis.cpp
ZuseZ4: Undone

enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -1867,7 +1869,7 @@ bool ActivityAnalyzer::isInstructionInactiveFromOrigin(TypeResults &TR,
 
       // If requesting empty unknown functions to be considered inactive, abide
       // by those rules
-      if (!isCertainCacheable(called) && called->empty() &&
+      if (!isCertainPrintMallocOrFree(called) && called->empty() &&
```
wsmoses, line 1872: I would also do same here

```
EnzymeEmptyFnInactive && called->empty() &&
              !hasMetadata(called, "enzyme_gradient") &&
              !hasMetadata(called, "enzyme_derivative") &&
              !isAllocationFunction(*called, TLI) &&
              !isDeallocationFunction(*called, TLI) &&
              !isa<IntrinsicInst>(op)
              ```

wsmoses:
Can the additional strings inside of `isCertainPrintMallocOrFree` be deferred to a later PR.  I'd be inclined to try to eliminate that function entirely (and have/use just relevant aliasing properties/etc defined).

Nevertheless it shouldn't impact the ability to differentiate code and should only make things not be cached (and thus faster), yes?

ZuseZ4:
I would surely prefer both. Having two almost identical functions seems strange and getting this PR merged would help me to simplify the buggy testcases. For those I'll also need a follow up PR either way, so I'll strip the string changes.
They also didn't affect the outcome of any of my use cases or the added tests.

enzyme/Enzyme/ActivityAnalysis.cpp
ZuseZ4: done

enzyme/Enzyme/ActivityAnalysis.cpp
ZuseZ4: I think (?) you need to mark this change request as solved or directly merge this PR. Github won't let me do that.

wsmoses (approved):

PR #608, opened 2022-04-12
Sample Tablegen for non-memory math-like, non-calling code

wsmoses:
The aim here is to get this to a point we can use it for BLAS and let GSOC handle more general usage/etc.

Tentative test with atan/atan2.


ZuseZ4:
@wsmoses All tests should pass now. Any objections to merging it?

ZuseZ4:
julia builds actually seems to fail based on not finding the generated tablegen .inc file. Does it have extra requirements?

wsmoses:
@vchuravy

vchuravy:
We do an out-of-tree build that is the only thing special? You can reproduce this by just running the build_local.jl script in deps in Enzyme.jl

ZuseZ4:
> We do an out-of-tree build that is the only thing special? 

out-of-tree builds work, tested it locally in a tmp dir.
However I have been able to reproduce the julia failures locally too.

There are two things that seem broken:
First, the tablegen binary isn't build correctly when using the llvm that julia ships, even --help gives an error (saying I should try --help). Using instead a locally build llvm-14.04 helps.
The next thing is that the target section doesn't seem to work nicely with tablegen. When setting -t Enzyme-X EnzymeBCLoader-X compilation fails with both llvms. Removing those flags however fixes the compilation.
I can look on how to fix the second issue, but I don't know what / how you need to change the llvm version which you download in order to play nicely with tablegen. Probably just a less stripped version, since llvm-12 on it's own works with this patch (see CI success)?

vchuravy:
This seems to not have fully fixed the issue? There is still a CMAKE dependent issue

vchuravy:
What does `enzyme-tblgen --help` yield?

vchuravy:
Also where is that error from I am seeing:

https://github.com/EnzymeAD/Enzyme/runs/7044395426?check_suite_focus=true

Which is just CMAKE order

ZuseZ4:
> What does `enzyme-tblgen --help` yield?

For the one based on julia-llvm nothing useful, unfortunately.
```
/home/drehwald/.julia/scratchspaces/7cc45869-7501-5eee-bdea-0790c847d4ef/build/tools/enzyme-tblgen/enzyme-tblgen --gen-derivatives --help
enzyme-tblgen: Unknown command line argument '--help'.  Try: '/home/drehwald/.julia/scratchspaces/7cc45869-7501-5eee-bdea-0790c847d4ef/build/tools/enzyme-tblgen/enzyme-tblgen --help'
enzyme-tblgen: Did you mean '--gen-derivatives'?
``` 

My llvm-14.04 local based one gives me

```
./tools/enzyme-tblgen/enzyme-tblgen --help
USAGE: enzyme-tblgen [options] <input file>

OPTIONS:

Color Options:

  --color                           - Use colors in output (default=autodetect)

General options:

  -D=<macro name>                   - Name of the macro to be defined
  -I=<directory>                    - Directory of include files
  -d=<filename>                     - Dependency filename
  Action to perform:
      --gen-derivatives                - Generate instruction derivative
  --no-warn-on-unused-template-args - Disable unused template argument warnings.
  -o=<filename>                     - Output filename
  --time-phases                     - Time phases of parser and backend
  --write-if-changed                - Only write output if it changed

Generic Options:

  --help                            - Display available options (--help-hidden for more)
  --help-list                       - Display list of available options (--help-list-hidden for more)
  --version                         - Display the version of this program
```

This hasn't been affected by the last commits here, they both had that behaviour before.

ZuseZ4:
I just feel like the enzyme-tg binary produced with julia-llvm is simply broken.

`tools/enzyme-tblgen/enzyme-tblgen -I ../../../llvm-14/llvm/include ../Enzyme/InstructionDerivatives.td`
This tg was generated using llvm-14.04 and will print / generate all the things as expected.

Trying the same comand with the tg which julia build leave in the scratchpad just fails with the same errors as shown in CI.

ZuseZ4:
@vchuravy time for another quick meeting to check again?
Apparently the cmake also breaks the compiler explorer.

wsmoses:
Hopeful fix: https://github.com/EnzymeAD/Enzyme/commit/1f7356cd552c8ffac9c95f06316dbcbdc9159ada

PR #611, opened 2022-04-13
Reverse vector mode

tgymnich:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -1735,6 +1786,14 @@ class AdjointGenerator
     ((DiffeGradientUtils *)gutils)->applyChainRule(Builder, rule, args...);
   }
 
+  /// Unwraps an collection of constant vector derivatives from their internal
+  /// representations and applies a function f to each element.
+  template <typename Func>
```
wsmoses, line 1791: Covered already?

enzyme/Enzyme/GradientUtils.h
```diff
@@ -2098,7 +2098,7 @@ class DiffeGradientUtils : public GradientUtils {
         idx2.push_back(v);
         auto selects = addToDiffe(
             val, BuilderM.CreateExtractValue(dif, ArrayRef<unsigned>(i)),
-            BuilderM, nullptr, idx2);
+            BuilderM, addingType, idx2);
```
wsmoses, line 2101: Save for later? This seems like the wrong way to handle (and worth explicitly testing)

enzyme/Enzyme/GradientUtils.h
```diff
@@ -2117,7 +2117,7 @@ class DiffeGradientUtils : public GradientUtils {
...
```
wsmoses, line 2120: Same comment

wsmoses: E.g. I would've expected addingType to also be an array and needing to extract the inner type as the inner addingType

wsmoses: That said usually the result here comes directly from TA and that probably gives a scalar -- but this is likely why this was nullptr.

wsmoses (approved):
LGTM, but make sure Julia passes before merging.

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -970,11 +970,19 @@ class Enzyme : public ModulePass {
...
+      else if (cast<Function>(fn)->getReturnType()->isFPOrFPVectorTy()) {
+        Constant *seed =
+            ConstantFP::get(cast<Function>(fn)->getReturnType(), 1.0);
+        if (width == 1) {
+          args.push_back(seed);
+        } else {
+          ArrayType *arrayType =
+              ArrayType::get(cast<Function>(fn)->getReturnType(), width);
+          args.push_back(
+              ConstantArray::get(arrayType, SmallVector(width, seed)));
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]

PR #630, opened 2022-04-22
Fix LLVM 14 build

wsmoses:


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -792,11 +792,19 @@ class Enzyme : public ModulePass {
               element = Builder.CreateBitCast(
                   element, PointerType::get(Type::getInt8Ty(CI->getContext()),
                                             elementPtrTy->getAddressSpace()));
+#if LLVM_VERSION_MAJOR >= 7
```
tgymnich, line 795: are we even supporting llvm < 7?

tgymnich (approved):

enzyme/Enzyme/Enzyme.cpp
wsmoses: It remains untested and not explicitly supported, but at one point we were, and we might as well not make it harder.

PR #641, opened 2022-04-29
Fix reverse vector mode malloc

tgymnich:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -10177,8 +10191,12 @@ class AdjointGenerator
...
                   (Mode == DerivativeMode::ForwardModeSplit &&
                    backwardsShadow)) {
-                if (!inLoop)
-                  zeroKnownAllocation(bb, anti, args, *called, gutils->TLI);
+                if (!inLoop) {
+                  auto rule = [&](Value *anti) {
+                    zeroKnownAllocation(bb, anti, args, *called, gutils->TLI);
+                  };
+                  applyChainRule(bb, rule, anti);
+                }
```
wsmoses, line 10199: The `  !isa<AllocaInst>(anti)) {` part of the check below needs to be modified, since presumably in vector mode that won't be the case for the equivalent condition.

tgymnich: done

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -10075,69 +10075,77 @@ class AdjointGenerator
...
                   called->getName() == "_Znwm") {
                 if (auto ci = dyn_cast<ConstantInt>(args[0])) {
                   unsigned derefBytes = ci->getLimitedValue();
                   CallInst *cal =
                       cast<CallInst>(gutils->getNewFromOriginal(orig));
+                  auto rule = [&](Value *anti) {
 #if LLVM_VERSION_MAJOR >= 14
-                  cast<CallInst>(anti)->addDereferenceableRetAttr(derefBytes);
-                  cal->addDereferenceableRetAttr(derefBytes);
+                    cast<CallInst>(anti)->addDereferenceableRetAttr(derefBytes);
```
wsmoses, line 10116: This is not something you want to do an applychainrules for, as you actually need this to apply to the actual call itself. In retrospect the extractMeta I added, might actually give you the correct behavior on this, but I would still prefer this functionality be inside of the first applyChainRule call all together.

enzyme/Enzyme/GradientUtils.h
```diff
@@ -1300,8 +1300,12 @@ class GradientUtils : public CacheUtility {
           if (auto I = dyn_cast<Instruction>(U))
             todo.push_back(I);
       }
+
       for (auto v : llvm::reverse(seen)) {
-        assert(v->getNumUses() == 0);
+        for (auto &use : v->uses())
```
wsmoses, line 1305: What case is this needed for?

tgymnich: before we were checking if the value is unused. I modified this to check if the values is unused or if the uses only happen inside seen (which are due to be removed anyways). This happens a lot with our unwraps in vector mode.

wsmoses: In theory this should already occur (at least for primal instructions) by an analysis beforehand that deduces which values are needed -- should this not also apply to the shadow (though perhaps the vector complicates).

In any case a minimal code snippet for this would be useful to understand.

tgymnich: Consider looprematmalloc.ll:

`rematerializedShadowPHIs` contains:  `%"i1'mi" = phi [3 x i8*] `


seen contains:
```
<begin dump>
  %0 = phi [3 x i8*] 
  %1 = extractvalue [3 x i8*] %0, 0
  %"i'ipc" = bitcast i8* %1 to [30 x double]*
  %2 = insertvalue [3 x [30 x double]*] undef, [30 x double]* %"i'ipc", 0
  %4 = insertvalue [3 x [30 x double]*] %2, [30 x double]* %"i'ipc7", 1
  %6 = insertvalue [3 x [30 x double]*] %4, [30 x double]* %"i'ipc8", 2
  %"i20'ipg" = getelementptr inbounds [30 x double], [30 x double]* %"i'ipc", i64 0, i64 %iv1
  %7 = insertvalue [3 x double*] undef, double* %"i20'ipg", 0
  %8 = insertvalue [3 x double*] %7, double* %"i20'ipg9", 1
  %9 = insertvalue [3 x double*] %8, double* %"i20'ipg10", 2
  %"i7'ipg" = getelementptr inbounds [30 x double], [30 x double]* %"i'ipc", i64 0, i64 %iv
  %10 = insertvalue [3 x double*] undef, double* %"i7'ipg", 0
  %11 = insertvalue [3 x double*] %10, double* %"i7'ipg23", 1
  %12 = insertvalue [3 x double*] %11, double* %"i7'ipg24", 2
  %3 = extractvalue [3 x i8*] %0, 1
  %"i'ipc7" = bitcast i8* %3 to [30 x double]*
  %"i20'ipg9" = getelementptr inbounds [30 x double], [30 x double]* %"i'ipc7", i64 0, i64 %iv1
  %"i7'ipg23" = getelementptr inbounds [30 x double], [30 x double]* %"i'ipc7", i64 0, i64 %iv
  %5 = extractvalue [3 x i8*] %0, 2
  %"i'ipc8" = bitcast i8* %5 to [30 x double]*
  %"i20'ipg10" = getelementptr inbounds [30 x double], [30 x double]* %"i'ipc8", i64 0, i64 %iv1
  %"i7'ipg24" = getelementptr inbounds [30 x double], [30 x double]* %"i'ipc8", i64 0, i64 %iv
</end dump>
```

wsmoses: The extract shouldn't depend on each other (and don't here).

I don't understand why the insertvalue should exist here, since we should just make a ficticious phi (e.g. shadow of value placeholder) with all elements all at once, rather than inserting into it?

tgymnich: The insert values are here because they are using the first phi we erased.

wsmoses: Okay, I still think this area of the code is likely to contain bugs (not necessarily due to, but possibly accentuated from this).

I'll approve this for now, but we should do a deeper dive on this.

wsmoses (approved):

tgymnich:
@wsmoses agreed.

Issue #643, opened 2022-04-30
Reverse vector mode stack overflow


tgymnich:
Reverse vector mode causes a stack overflow when looking up / unwrapping values for a large vector width.

This issue seems related to the stack overflow we have seen in Julia.

https://fwd.gymni.ch/apJzvT

**note**: The above example works perfectly fine with lower vector widths.

wsmoses:
So the problem is that when unwrapping vector'd values in the reverse pass, we will call an unwrap for each instruction -- which in this case is "vector width" * "call chain length", where each occurance is a call to a (actually a few) co-recursive functions.

In my mind the immediate solution here is to modify the insertvalue in unwrapM (https://github.com/EnzymeAD/Enzyme/blob/e9084808993c94661f33ec29955499b51a8ada03/enzyme/Enzyme/GradientUtils.cpp#L445) to iterate through aggregate of the insertion for all operands which are themselves insertvalue and not marked with an explicit `knownRecomputeHeuristic` to be looked up rather than unwrapped.

wsmoses:
Something related that I think we may want to explicitly do:

When doing the extractelement for the next instruction of X. If the aggregate operand we would use for the extract is itself an insertelement, use its operand if the offset != the current index we're extracting (at which point we can just use the inserted value directly).

PR #644, opened 2022-05-01
Simplify vector extraction

wsmoses:


tgymnich:
Does this fix #643?

wsmoses:
Haven't tested, but at least am partially hoping.

tgymnich:
Just tested. It doesn't fix it.

wsmoses:
Alas, well this is still useful for both performance and reducing the amount of LLVM we produce.

tgymnich (approved):

PR #647, opened 2022-05-03
Fix unwrap

### LINKED ISSUE #643: Reverse vector mode stack overflow

tgymnich:
fixes #643


enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -442,22 +442,36 @@ Value *GradientUtils::unwrapM(Value *const val, IRBuilder<> &BuilderM,
...
-      unwrappedLoads[newi] = val;
-      if (newi->getParent()->getParent() != op->getParent()->getParent())
-        newi->setDebugLoc(nullptr);
+    SmallVector<std::pair<Value *, unsigned>, 0> insertElements;
+    Value *agg = op->getAggregateOperand();
+    while (auto op1 = dyn_cast<InsertValueInst>(agg)) {
+      if (knownRecomputeHeuristic.count(op1) &&
+          knownRecomputeHeuristic[op1] == true)
+        break;
+      insertElements.push_back({op1->getInsertedValueOperand(),
```
wsmoses, line 451: You should check that there's only 1 such index?

tgymnich: or maybe just assert?

wsmoses: No because there are plenty of multi-index insertions, and we would then have compilation fail when unwrapping them (ignorant of vector mode even)

tgymnich: added check

wsmoses: I don't see the check?

tgymnich: forgot to commit and push

wsmoses: Still dont see this `if (op1->getNumIndices() != 1) break` check

wsmoses: The count doesn't need to check against 1, if that's what you misunderstood this as.

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -442,22 +442,36 @@ Value *GradientUtils::unwrapM(Value *const val, IRBuilder<> &BuilderM,
...
+                                op1->getInsertedValueOperandIndex()});
+      agg = op1->getAggregateOperand();
     }
+
+    Value *toreturn = agg;
+    for (auto &&[valOp, idcs] : reverse(insertElements)) {
+      auto op1 = getOp(valOp);
```
wsmoses, line 458: Can you also add a regular reverse mode test where the knownrecompute is somewhere in the chain, and confirm we get the intended behavior from the cache analysis?

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -442,22 +442,38 @@ Value *GradientUtils::unwrapM(Value *const val, IRBuilder<> &BuilderM,
...
-    auto toreturn = BuilderM.CreateInsertValue(op0, op1, op->getIndices(),
-                                               op->getName() + "_unwrap");
-    if (permitCache)
-      unwrap_cache[BuilderM.GetInsertBlock()][idx.first][idx.second] = toreturn;
-    if (auto newi = dyn_cast<Instruction>(toreturn)) {
-      newi->copyIRFlags(op);
-      unwrappedLoads[newi] = val;
-      if (newi->getParent()->getParent() != op->getParent()->getParent())
-        newi->setDebugLoc(nullptr);
+    SmallVector<std::pair<Value *, unsigned>, 0> insertElements;
```
tgymnich, line 445:
```suggestion
    SmallVector<std::pair<Value *, unsigned>, 4> insertElements;
```

wsmoses (approved):

PR #652, opened 2022-05-08
Batch Mode

tgymnich:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -1724,6 +1724,8 @@ class AdjointGenerator
       break;
     case DerivativeMode::ReverseModePrimal:
       return;
+    case DerivativeMode::BatchMode:
```
wsmoses, line 1727: I don't think we should make this a separate derivative mode. Namely, you can make an instruction visitor that for every instruction, excluding a call to a function, we will extract all operands, then insert all the results.

PR #655, opened 2022-05-10
Batch mode [2]

tgymnich:


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -506,6 +616,171 @@ class Enzyme : public ModulePass {
...
+      BATCH_TYPE ty = BATCH_TYPE::SCALAR;
+      Optional<StringRef> metaString = getMetadataName(res);
+
+      // handle metadata
+      if (metaString && metaString.getValue().startswith("enzyme_")) {
+        if (*metaString == "enzyme_scalar") {
+          ty = BATCH_TYPE::SCALAR;
+        } else if (*metaString == "enzyme_vector") {
+          ty = BATCH_TYPE::VECTOR;
+        } else if (*metaString == "enzyme_buffer") {
```
wsmoses, line 674: Name?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -4226,6 +4227,146 @@ Function *EnzymeLogic::CreateForwardDiff(
...
+    }
+  }
+
+  InstructionBatcher *batcher =
+      new InstructionBatcher(tobatch, NewF, width, vectorizedValues,
+                             originalToNewFn, toVectorize, *this);
+
+  for (BasicBlock &BB : *tobatch) {
+    for (Instruction &I : BB) {
+      if (!isa<PHINode>(I))
```
wsmoses, line 4348: See different block order case from whiteboard.

enzyme/Enzyme/InstructionBatcher.h
```diff
@@ -0,0 +1,242 @@
...
+
+    for (unsigned i = 1; i < width; ++i) {
+      ValueToValueMapTy vmap;
+      Instruction *new_inst = placeholder->clone();
+      vmap[placeholder] = new_inst;
+
+      for (unsigned j = 0; j < inst.getNumOperands(); ++j) {
+        Value *op = inst.getOperand(j);
+        Value *new_op;
+        if (isa<Constant>(op)) {
```
tgymnich, line 78: Can we extract this into a function to evaluate a condition? If it's a constant data I concur, but if this is a global, you might need a batch global or something?

enzyme/Enzyme/InstructionBatcher.h
```diff
@@ -0,0 +1,242 @@
...
+          new_op = op;
+        } else if (toVectorize.count(op) != 0) {
+          new_op = vectorizedValues[op][i];
+        } else {
+          new_op = originalToNewFn[op];
+        }
+        vmap[placeholder->getOperand(j)] = new_op;
+      }
+
+      if (placeholders.size() == width) {
```
tgymnich, line 88: // For instructions which return a value

enzyme/Enzyme/InstructionBatcher.h
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  void visitSwitchInst(llvm::SwitchInst &inst) {
+    EmitFailure("SwitchConditionCannotBeVectorized", inst.getDebugLoc(), &inst,
+                "switch conditions have to be scalar values", inst);
+    llvm_unreachable("vectorized control flow is not allowed");
+  }
+
+  void visitBranchInst(llvm::BranchInst &branch) {
+    EmitFailure("BranchConditionCannotBeVectorized", branch.getDebugLoc(),
```
tgymnich, line 148: At one point it would be cool to have as a fallback, check if these are all the same -- and if not go to an error.

enzyme/Enzyme/InstructionBatcher.h
```diff
@@ -0,0 +1,242 @@
...
+    auto found = vectorizedValues.find(&call);
+    assert(found != vectorizedValues.end());
+    auto placeholders = found->second;
+    CallInst *placeholder = cast<CallInst>(placeholders[0]);
+    IRBuilder<> Builder2(placeholder);
+    Builder2.SetCurrentDebugLocation(DebugLoc());
+    Function *orig_func = getFunctionFromCall(&call);
+
+    auto isDefined = !orig_func->isDeclaration();
+    if (isDefined) {
```
tgymnich, line 192: Style: I'd restructure this slightly to:
if (!isDefined) {
  visitInstruction();
  return;
}

...

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -4236,6 +4237,256 @@ Function *EnzymeLogic::CreateForwardDiff(
   return nf;
 }
 
+llvm::Function *EnzymeLogic::CreateBatch(Function *tobatch, unsigned width,
+                                         ArrayRef<BATCH_TYPE> arg_types) {
```
tgymnich, line 4241: I think you may want a BATCH_TYPE for the return?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -4236,6 +4237,256 @@ Function *EnzymeLogic::CreateForwardDiff(
...
+      Argument *arg = tobatch->arg_begin() + i;
+      toVectorize.insert(arg);
+
+      for (auto user : arg->users()) {
+        worklist.insert(user);
+      }
+    }
+  }
+
+  if (!NewF->getReturnType()->isVoidTy())
```
tgymnich, line 4315: This would then be if vector.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -4236,6 +4237,256 @@ Function *EnzymeLogic::CreateForwardDiff(
...
+        worklist.insert(user);
+      }
+      for (auto &user : inst->operands()) {
+        worklist.insert(user);
+      }
+    }
+  }
+
+  // find instructions to vectorize (going down / refine)
+  SetVector<llvm::Value *, std::deque<llvm::Value *>> refinelist;
```
tgymnich, line 4347: You should also add all instructions here, not just the arg users.

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -4236,6 +4237,256 @@ Function *EnzymeLogic::CreateForwardDiff(
...
+    if (isa<CallInst>(todo))
+      continue;
+
+    if (auto todo_inst = dyn_cast<Instruction>(todo)) {
+
+      if (todo_inst->mayReadOrWriteMemory())
+        continue;
+
+      bool br = false;
+      for (auto &arg : todo_inst->operands()) {
```
tgymnich, line 4375: Let's also have this work on phi nodes.

The algorithm for this would be as follows:
toCheck = todo_inst->operands();
safe = set();
legal = true;
while (!toCheck.empty()) {
  cur = todo.pop_front();
  if (safe.count(cur)) continue;
  safe.insert(cur);
  if (cur is scalar) continue;
  if (cur is not call and doesn't write memory) {
    add all cur's operands.
    continue
  }
  legal = false; break;
}

enzyme/Enzyme/EnzymeLogic.cpp
tgymnich: @wsmoses how would this be expressed in our calling conversation?

PR #659, opened 2022-05-18
closes #657

### LINKED ISSUE #657: f != originalToNewFn.end()
I have a library which crashes when using Enzyme. After the IO updates it now failes at the following assertion:
`f != originalToNewFn.end()`

I have the errormessage / stacktrace here
[err.txt](https://github.com/EnzymeAD/Enzyme/files/8712129/err.txt)

Is there anything obviously wrong around this assertion?
I will work on providing a minimized src code example.

Also, how would I pass the extra enzyme-print flags, when using LLDEnzyme?

ZuseZ4:
@wsmoses let me know if I took your comment too literal. Seems to fix at least my Bug.


enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -3821,6 +3821,10 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
       }
     }
 
+    if (auto arg = dyn_cast<ConstantExpr>(oval)) {
+      return oval;
```
tgymnich, line 3825: You might also want to handle the vector mode case here. E.g. wrap this inside of applyChainRule.

tgymnich, line 3825: You might also want to handle the vector mode case here. E.g. wrap this inside of applyChainRule.

wsmoses (approved):
LGTM, but address Tim's comment and please add a test.

enzyme/Enzyme/GradientUtils.cpp
ZuseZ4: done.

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -3821,6 +3821,11 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
       }
     }
 
+    if (auto arg = dyn_cast<ConstantExpr>(oval)) {
+      auto rule = [&oval]() { return Constant::getNullValue(oval->getType()); };
```
wsmoses, line 3825: Oh you don't want to return the null, here but arg (null case is already taken care of above.

ZuseZ4: Thanks, I was already wondering about the test-case changes.

ZuseZ4:
@wsmoses Do you mind if I revert the test for now?
It not only triggers this, but also a second bug, I don't have any example which just triggers the first bug.
Once the second bug is resolved, I can add the testcase, without causing CI failures.

wsmoses:
I'm fine with that, but open an issue for the new bug, and reference this PR to make sure we remember that the test for it should also include the test for this.

Issue #656, opened 2022-05-13
Regression / `pp->getNumUses() == 0' failed


ZuseZ4:
After some time I came back to an old Benchmark. Using the latest Enzyme release and a comparably new LLVM i now get a compilation failure:

```
 pp:   %47 = phi double  of   %29 = fmul double %27, %27
ld.lld: /home/drehwald/prog/Rust/src/tools/enzyme/enzyme/Enzyme/GradientUtils.h:1326: void GradientUtils::eraseFictiousPHIs(): Assertion `pp->getNumUses() == 0' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: /usr/bin/ld.lld -z relro --hash-style=gnu --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o rb_t /lib/x86_64-linux-gnu/crt1.o /lib/x86_64-linux-gnu/crti.o /usr/lib/gcc/x86_64-linux-gnu/11/crtbegin.o -L../../../external/lib -L/usr/lib/gcc/x86_64-linux-gnu/11 -L/usr/lib/gcc/x86_64-linux-gnu/11/../../../../lib64 -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -L/home/drehwald/prog/Rust/build/x86_64-unknown-linux-gnu/llvm/bin/../lib -L/lib -L/usr/lib -plugin-opt=mcpu=cascadelake -plugin-opt=O3 rb_t.o -lz -ltinyxml --lto-legacy-pass-manager -mllvm=-load=/home/drehwald/prog/Rust/src/tools/enzyme/enzyme/build/Enzyme/LLDEnzyme-14.so -lstdc++ -lm -lgcc_s -lgcc -lc -lgcc_s -lgcc /usr/lib/gcc/x86_64-linux-gnu/11/crtend.o /lib/x86_64-linux-gnu/crtn.o
1.	Running pass 'Enzyme Pass' on module 'ld-temp.o'.
Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
[...]

PR #660, opened 2022-05-20
Move TypeResults into GradientUtils

tgymnich:


wsmoses (approved):
I'm supportive of this PR, but regardless -- is there a specific impetus which caused this to be necessary recently?

tgymnich:
@wsmoses I don't think fixing #656 is a good use of my time. Fixing this will require lots of extra checks in AdjointGenerator to work around the work around for #637. I'd much rather fix #637, and have all mode share the same code path.
In order to fix #637 we might need `TypeResults` inside of `invertPointerM`.

wsmoses:
Am fine with that, but let's continue to do the parts one at a time (easier to test)

PR #662, opened 2022-05-24
mark unnecessary mallocs

wsmoses:
Solves https://github.com/EnzymeAD/Enzyme.jl/issues/311 until the jl_reshape needing a custom reverse


vchuravy:
Add a test for mallformed malloc?

wsmoses:
This is distinct from the malformed malloc from Julia, but just a different part of malloc analysis (e.g. preserving a memcpy from initialized memory is worthless)

PR #667, opened 2022-05-27
Add log1p

wsmoses:
cc @jedbrown @LeilaGhaffari


jedbrown (approved):
Thanks. This works in our test.

PR #669, opened 2022-05-28
Fix ROCM

wsmoses:


enzyme/CMakeLists.txt
```diff
@@ -128,6 +128,11 @@ if (ENZYME_FLANG)
   add_definitions(-DFLANG=1)
 endif()
 
+option(ENZYME_ROCM "Build for non-version compliant ROCM" OFF)
```
vchuravy, line 131: grrml

vchuravy: Can we document which version this is? Is this ROCM 4.X or 5?

vchuravy (approved):

jedbrown:
Thanks. If I've successfully managed my environment variables/build paths, then this (on CPU) fails when built with ROCm's LLVM, but not when build with mainline LLVM. And when I add `-ceed /gpu/hip`, I get that `stddef.h` error. (I don't remember if you figured out the cause of that today.)

```console
$ ./elasticity -problem FSInitial-NH-AD -E 1 -nu .3 -degree 2 -dm_plex_box_faces 1,1,1 -num_steps 1 -bc_clamp 6 -bc_traction 5 -bc_traction_5 0,0,.01 -snes_monitor -snes_converged_reason -outer_ksp_monitor_true_residual -pc_type none
```

wsmoses:
The stddef.h error was fixed by rebuilding AMD comgr to point to the new ROCM installation (5.1.1). Specifically that involves setting the LD_LIBRARY_PATH to /home/wsmoses/rocm2/... instead of rocm.

What's the error you get when building using ROCm's LLVM? (and what are the precise CMake flags and ROCM path's)?

enzyme/CMakeLists.txt
wsmoses: 5.1.1

PR #683, opened 2022-06-07
move inactive fn-matching to demangled names

ZuseZ4:


ZuseZ4:
ss5print might be swift? It probably should be translated then later.

enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -266,6 +200,34 @@ const std::set<std::string> KnownInactiveFunctions = {
     "logbl",
 };
 
+const char *DemangledKnownInactiveFunctionsStartingWith[] = {
+    "f90io",
```
wsmoses, line 204: Can this be a separate category. In particular the fortran f90io as an example doesn't belong here.

enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -1438,8 +1412,9 @@ bool ActivityAnalyzer::isConstantValue(TypeResults const &TR, Value *Val) {
               F->getName() == "__fd_sincos_1") {
             return false;
           }
-          for (auto FuncName : KnownInactiveFunctionsStartingWith) {
-            if (F->getName().startswith(FuncName)) {
+          auto dName = demangle(F->getName().str());
```
wsmoses, line 1415: Please make sure this works on LLVM 7/8

ZuseZ4: LLVM-7 demangle support seems too basic: https://github.com/llvm/llvm-project/blob/00b161b8971bc6d3cb55f13502288b8fe0dbaa42/llvm/include/llvm/Demangle/Demangle.h

So we now have the demangled support for C++ for >= 9 and the unmangled support for f90io, ss5print and (if <=8) for most C++ IO.

enzyme/Enzyme/ActivityAnalysis.cpp
ZuseZ4: afaik only ss5print and f90io are non-c++ names? So I moved them back to check unconditionally for those unmangled.

enzyme/Enzyme/ActivityAnalysis.cpp
ZuseZ4: diff looks like I missed one place

ZuseZ4: fixed

tgymnich (approved):

PR #684, opened 2022-06-09
Ubuntu22

tgymnich:


ZuseZ4:
Would it make sense to set up a kind of support policy in order to limit what we test (and thus especially CI time)?
18.04 is more than 4 yrs old and EOL in less than a year, so I wonder if we can't just replace it by 22.04?

tgymnich:
Supporting ubuntu 18.04 until its EOL and replacing it with 22.04 sound like a good plan.

tgymnich:
Failures in the integration tests for LLVM 11,12 and 13 in `Integration/ReverseMode/eigensumsqdyn-notmp.cpp`.
```
warning: <unknown>:0:0: failed to deduce type of copy   call void @llvm.memcpy.p0i8.p0i8.i64(i8* nonnull align 8 dereferenceable(32) %5, i8* nonnull align 8 dereferenceable(32) %6, i64 32, i1 false) #18
[599](https://github.com/EnzymeAD/Enzyme/runs/6832529792?check_suite_focus=true#step:7:600)
```

wsmoses:
We should investigate the Type analysis failures to see if we can make them detectable/fixable.

Otherwise lgtm

wsmoses:
The segfault from `Enzyme :: Integration/ReverseMode/integrateexp.cpp` needs to be fixed before merging. Anyone have time to investigate?

tgymnich:
Mostly addressed in #2074

PR #696, opened 2022-06-16
More global typeinfo

wsmoses:


ZuseZ4 (approved):
but maybe check formating. was broken in ci after your last addition

PR #703, opened 2022-06-24
move asin + log1p to tg

ZuseZ4:


enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -68,6 +70,17 @@ def : CallPattern<(Op $y, $x),
                   (FNeg (FDiv (FMul (DiffeRet<"">), $y), (FAdd (FMul $x, $x), (FMul $y, $y))))
                   ]
                   >;
+
+def : CallPattern<(Op $x),
+                  ["asin", "asinf", "asinl"],
+                  [(FDiv (DiffeRet<"">), (SameSqrt (FSub (ConstantFP<"1.0"> $x), (FMul $x, $x)))  )]
```
wsmoses, line 76: Can you add support for intrinsics here and use the llvm sqrt intrinsic rather than the std libc function?

ZuseZ4: done

PR #704, opened 2022-06-24
Fix recursive global

wsmoses:


ZuseZ4 (approved):

vchuravy (approved):

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -4103,14 +4103,11 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
...
-        auto ip = arg->getInitializer() ? invertPointerM(arg->getInitializer(),
-                                                         B, /*nullShadow*/ true)
-                                        : Constant::getNullValue(type);
 
-        auto rule = [&](Value *ip) {
+        auto rule = [&]() {
           auto shadow = new GlobalVariable(
               *arg->getParent(), elemTy, arg->isConstant(), arg->getLinkage(),
-              cast<Constant>(ip), arg->getName() + "_shadow", arg,
+              Constant::getNullValue(type), arg->getName() + "_shadow", arg,
```
tgymnich, line 4110: this should be `elemTy` instead of `type`, because `type` is the vectorised type in the case of vector mode.

PR #712, opened 2022-06-25
Vectorize at leaf nodes

tgymnich:


tgymnich:
Closed in favor of #747

PR #713, opened 2022-06-25
Fix type mismatch in GlobalVariable initialiser in the case of vector mod

tgymnich:
Use scalar type instead of vector type for the GlobalVariable initialiser.


tgymnich:
for future reference #704

wsmoses (approved):
LGTM, add a test?

tgymnich:
There is already a test

PR #719, opened 2022-06-27
handle old cmake with newer llvm

ZuseZ4:
Users might want to use llvm with an older, officially unsupported cmake (see our CI).
In this case we don't want to use the tablegen function from llvm but rather our own replacement which doesn't use the TRANSFORM command.


wsmoses (approved):
Check this passes benchmark CI, then go ahead and merge.

ZuseZ4:
> Check this passes benchmark CI, then go ahead and merge.

Already passed CI Benchmarking a few times before merging main, so I'll just let CI merge it.

PR #732, opened 2022-07-09
Fix wrapped value used in chain rule

tgymnich:


wsmoses (approved):
Add a test?

wsmoses:
@tgymnich to preserve the issue. Add a test?

tgymnich:
LOL auto merge. I do not think a test is necessary here, since Tablegen will hopefully get rid of these kind of errors.

PR #733, opened 2022-07-09
Cast gutils to correct subclass for applyChainRule

tgymnich:


ludgerpaehler (approved):
Looks sensible to me! Would recommend to better pronounce the difference between `DiffeGradientUtils`, and `GradientUtils` in `GradientUtils.h` for better dev-documentation at that point.

PR #743, opened 2022-07-20
fix ConstantExpr handling in CreateAugmentedPrimal

### LINKED ISSUE #721: Assertion `VMap[invertedRetPs[ri]]' failed
Another (somewhat larger) variation of #720 and #677 triggered the assertion above.
Passed you the stacktrace on Slack.

ZuseZ4:
llvm's copyFunctionInto won't add constantexpr to the vmap, since It can be shared, similar to constantData.

fixes #721

### DIFF
--- a/enzyme/test/Enzyme/ReverseMode/constexpr.ll
+++ b/enzyme/test/Enzyme/ReverseMode/constexpr.ll
@@ -1,0 +1,26 @@
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -mem2reg -instsimplify -simplifycfg -S | FileCheck %s
+
+@_ZTId = external dso_local constant i8*
+
+define i8* @_ZNK4implIdE4typeEv()  {
+  ret i8* bitcast (i8** @_ZTId to i8*)
+}
+
+declare void @_Z17__enzyme_virtualreverse(i8*)
+
+define void @_Z18wrapper_1body_intsv()  {
+  call void @_Z17__enzyme_virtualreverse(i8* bitcast (i8* ()* @_ZNK4implIdE4typeEv to i8*))
+  ret void
+}
+
+; CHECK: define internal { i8*, i8*, i8* } @augmented__ZNK4implIdE4typeEv() #1 {
+; CHECK-NEXT:   %1 = alloca { i8*, i8*, i8* }, align 8
+; CHECK-NEXT:   %2 = getelementptr inbounds { i8*, i8*, i8* }, { i8*, i8*, i8* }* %1, i32 0, i32 0
+; CHECK-NEXT:   store i8* null, i8** %2, align 8
+; CHECK-NEXT:   %3 = getelementptr inbounds { i8*, i8*, i8* }, { i8*, i8*, i8* }* %1, i32 0, i32 1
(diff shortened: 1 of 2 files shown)


wsmoses (approved):

enzyme/test/Enzyme/ReverseMode/constexpr.ll
```diff
@@ -0,0 +1,26 @@
...
+}
+
+declare void @_Z17__enzyme_virtualreverse(i8*)
+
+define void @_Z18wrapper_1body_intsv()  {
+  call void @_Z17__enzyme_virtualreverse(i8* bitcast (i8* ()* @_ZNK4implIdE4typeEv to i8*))
+  ret void
+}
+
+; CHECK: define internal { i8*, i8*, i8* } @augmented__ZNK4implIdE4typeEv() #1 {
```
wsmoses, line 16: This currently fails on LLVM7, please fix

ZuseZ4:
@wsmoses MacOs fails, but not sure what the error should tell me?

ZuseZ4:
I also had an issue on Linux on an earlier commit here: https://github.com/EnzymeAD/Enzyme/runs/7436868577?check_suite_focus=true
Maybe it's just the infrastructure having a bad day?
Can someone working on MacOS test it? I took the run command from the Reverse/constant.ll example, so I see no reason why my new test should fail on macOS.

tgymnich:
I was able to reproduce the issue on my machine. I will look into it soon.

tgymnich:
It seems like the `inversionAllocs` are marked unreachable and successively deleted (EnzymeLogic.cpp:2235). The crash then happens when we want to access `inversionAllocs` the next time at EnzymeLogic.cpp:2301.

wsmoses:
@ZuseZ4 the solution then is to move the invertpointer calls (https://github.com/EnzymeAD/Enzyme/blob/e8ed87c549f07621f591d90ff91a445a6873f34c/enzyme/Enzyme/EnzymeLogic.cpp#L2295) prior to the deletion/move of inversionAllocas

ZuseZ4:
Thanks @tgymnich @wsmoses.

tgymnich:
Your added test still triggers assertions. They just don't show up in our CI.

```
    for (auto pair : toVirtual) {
      auto CI = pair.first;
      Constant *fn = dyn_cast<Constant>(CI->getArgOperand(0));
      if (!fn) {
        EmitFailure("IllegalVirtual", CI->getDebugLoc(), CI,
                    "Cannot create virtual version of non-constant value ", *CI,
                    *CI->getArgOperand(0));
        return false;
      }
      TypeAnalysis TA(Logic.PPC.FAM);

      auto Arch =
          llvm::Triple(
              CI->getParent()->getParent()->getParent()->getTargetTriple())
              .getArch();

      bool AtomicAdd = Arch == Triple::nvptx || Arch == Triple::nvptx64 ||
                       Arch == Triple::amdgcn;

      auto val = GradientUtils::GetOrCreateShadowConstant(
          Logic, TLI, TA, fn, pair.second, /*width*/ 1, AtomicAdd);
      CI->replaceAllUsesWith(ConstantExpr::getPointerCast(val, CI->getType()));
      CI->eraseFromParent();
      Changed = true;
    }
```

CI->getType() => void
val->getType() => i8*

PR #754, opened 2022-07-29
Julia malloc handling

wsmoses:


vchuravy:
What is the motivation? While Julia does provide Libc.malloc it shouldn't be used by most programs?

wsmoses:
```
No augmented forward pass found for malloc
declare i64 @malloc(i64) local_unnamed_addr
```

I do agree with you, but I'm actually myself using it right now to minimize a test (https://github.com/EnzymeAD/Enzyme.jl/issues/388) and we might as well handle it.

ZuseZ4 (approved):

PR #757, opened 2022-07-29
Fallback pthread [incorrect]

wsmoses:
@ZuseZ4


wsmoses:
@ZuseZ4 if you still need this, it can be done more easily at the user level now with `__attribute__((enzyme_inactive))`, so I'm going to close

PR #767, opened 2022-08-01
Enable custom allocation

wsmoses:


vchuravy (approved):

wsmoses:
@jedbrown while the immediate motivation for this custom tape allocation mechanism is to better inter operate with Julia GC, I remember a discussion about using a custom allocation function from your end as well that we should make sure can work with this.

jedbrown:
Thanks @wsmoses. Will there be an example (test or docs) of how to access this feature from C/C++? It's hard for me to interpret from the implementation code.

wsmoses:
Yeah this isn't yet accessible from a user facing API (currently designing pieces and trying to see what various requirements there are).

Presently the 2-3 functions desired are:
* typed allocation (Eg allocate X of type Y) which takes the type, alignment, and element count
* free
* optional enlarge existing allocation (which will fall back to the amortized power of two), which takes the current and next size, alignment, and previous allocation.

The type requirements from Julia are because we need to determine which objects need to be marked by the GC. (The corresponding malloc staging example of this at an internal API level is here: https://github.com/EnzymeAD/Enzyme.jl/pull/396)

My question for you presently is what information / context would be necessary or useful.

jedbrown:
How does this compose with vectorization? Like if you write a pointwise function and expect it to get inlined and vectorized, would you get a callback to allocate enough items for the vector length, with lanes interlaced for SIMD load/stores?

```c
static inline double f(double a, double b) {
    return ...; // expensive
}
for (int i=0; i<n; i++) { // we expect this loop to vectorize
    c[i] = f(a[i], b[i]);
}
```

And on GPU, is there any way that this interface would help the host do the allocation, or does it have to be allocated within a kernel?

wsmoses:
Regarding the vectorization comment, that's related though probably not going to be addressed through this interface. The vectorization part is more of a "should this part of the cache be a struct of arrays/array of structs", which we should add separately.

As for the GPU part -- yes. As an example one could preallocate a large chunk of GPU memory on the host, and then define the "allocator" function within the GPU to use memory from that block.

The GPU allocation would also benefit from allocation hoisting at a interprocedural level (currently this is done within a function across loops) -- which would be a separate but related part.

jedbrown:
Cool, so I think to make use of that, one would need a context (could hold the start of the buffer that was preallocated on the host) and an index (like a thread id, but I think it's better to allow the caller to compute an application-relevant index rather than having the allocator directly checking `threadIdx` and `blockIdx`, which requires making assumptions). Is that what you're thinking?

PR #769, opened 2022-08-02
Custom function-like registration

wsmoses:


LeilaGhaffari (approved):
This is working with our application code. Thanks for your help, @wsmoses !

PR #775, opened 2022-08-05
Expm1

wsmoses:
The derivative of expm1 behaves exactly like exp


ZuseZ4 (approved):

enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -107,3 +107,9 @@ def : CallPattern<(Op $x),
                   ["exp10"],
                   [(FMul (FMul (DiffeRet<"">), (Call<(SameFunc), [ReadNone,NoUnwind]> $x) ), (ConstantFP<"2.30258509299404568401799145468"> $x))]
                   >;
+
+// TODO replace with intrinsic call to exp
```
wsmoses, line 111: @ZuseZ4 do you have an ETA on intrinsic call support in tablegen. This is a direct use case where calling an intrinsic (exp) would be useful. Without it we fall back to calling expm1 and add 1.

ZuseZ4: done

PR #788, opened 2022-08-15
Update AdjointGenerator.h

ZuseZ4:


ZuseZ4:
Just comparing / minimizing the diff between my generated code and the existing one, that seems off.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -5791,7 +5791,7 @@ class AdjointGenerator
                        (!ycache &&
                         !gutils->isConstantValue(call.getArgOperand(1)))) {
               if (Mode != DerivativeMode::ForwardModeSplit)
-                trueXinc = lookup(trueXinc, Builder2);
+                trueYinc = lookup(trueYinc, Builder2);
```
wsmoses, line 5794: Your patch here is correct. This is a bug in the current code.

wsmoses (approved):
Add a test if possible, please.

ZuseZ4:
@wsmoses 
Testing this would require adding fblas_ddot.ll tests, rn. we only have cblas tests, so I prefer to do all the fblas tests later together, once I have tg-blas2 rdy (possibly semi-scripted). 
Also, is there some enzyme_x metadata to set an arg as uncacheable? I only know how to do it via c-abi, I couldn't find any metadata that sounded suitable.

PR #802, opened 2022-08-26
add remainder

ZuseZ4:

PR #807, opened 2022-08-30
Intrinsics

ZuseZ4:
fixes #703
fixes #802

@wsmoses I'll dislike the type part and will most likely adjust it once I start moving the mem related intrinsics and see what works better. For the meantime it should work. 
Also need to add some lines to automatically collect the useful attributes from the intrinsics and apply them to the outer call,
but I'd add that later if you don't mind.
Will add a test for the remainder in a sec.


enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -110,9 +115,22 @@ def : CallPattern<(Op $x),
...
+                  [(FDiv (DiffeRet<"">), (Intrinsic<"sqrt", ["args[0]->getType()"]> (FSub (ConstantFP<"1.0"> $x), (FMul $x, $x)))  )]                  
+                  >;
+def : CallPattern<(Op $x),
+                  ["log1p", "log1pf", "log1pl"],
+                  [(FDiv (DiffeRet<"">), (FAdd (ConstantFP<"1.0"> $x), $x))]
+                  >;
 def : CallPattern<(Op $x),
                   ["expm1"],
-                  [(FMul (FAdd (ConstantFP<"1.0"> $x), (Call<(SameFunc), [ReadNone,NoUnwind]> $x) ), (DiffeRet<"">))]
+                  [(FMul (Intrinsic<"exp", ["args[0]->getType()"]> $x), (DiffeRet<"">))]
```
wsmoses, line 135: Can you make a typeof expr, like maybe (Typeof $x)

enzyme/test/Enzyme/ForwardMode/remainder.ll
```diff
@@ -0,0 +1,39 @@
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -O3 -S | FileCheck %s
```
wsmoses, line 1: Can you change this to only do mem2reg simplifycfg and instsimplify?

We should check that the differeturn is handled properly

enzyme/Enzyme/InstructionDerivatives.td
wsmoses: Alternatively maybe something like [0,2] denotes the types of the 0th op and the 2nd op

ZuseZ4: I think I'll go for the TypeOf version for now, imho more readable.

enzyme/test/Enzyme/ForwardMode/remainder.ll
ZuseZ4: Done, output looks correct to me.

enzyme/test/Enzyme/ReverseMode/remainder.ll
```diff
@@ -0,0 +1,38 @@
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -O3 -S | FileCheck %s
```
wsmoses, line 1: This and the above tests still don't seem updated?

enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -110,9 +115,22 @@ def : CallPattern<(Op $x),
 def : CallPattern<(Op $x),
                   ["tan", "tanf", "tanl"],
                   [(FMul (DiffeRet<"">), (FAdd (ConstantFP<"1.0"> $x), (FMul(Call<(SameFunc), [ReadNone,NoUnwind]> $x):$c, $c)))]>;
-
-// TODO replace with intrinsic call to exp
+def : CallPattern<(Op $x, $y),
+                  ["remainder"],
+                  [
+                    (DiffeRet<"">),
+                    (FMul (DiffeRet<"">), (Intrinsic<"round", ["orig->getOperand(0)->getType()"]> (FDiv $x, $y)))
```
wsmoses, line 122: I think this should be -dret * round(x/y).  that is to say -- shouldn't there be a minus sign here?

enzyme/test/Enzyme/ReverseMode/remainder.ll
ZuseZ4: Have it locally. I didn't finish the other update yet and I generally try to push multiple commits simultaneously to not wait so long for CI.

enzyme/Enzyme/InstructionDerivatives.td
ZuseZ4: Done. This way I can add something else impl TypeInfo for the mem things.

enzyme/Enzyme/InstructionDerivatives.td
ZuseZ4: good catch, lost it when replacing constantFP<"-1"> by differet

PR #810, opened 2022-09-01
improve Intrinsics

ZuseZ4:
Since we create a call using the intrinsics I guess we should still apply some attributes to that call.
ReadNone seems safe for all the math ones we look at, NoUnwind seems to also be ok for most since they often don't seem to trap, but probably should be more added case by case? https://llvm.org/docs/LangRef.html#llvm-sqrt-intrinsic
Anything else that could be performance relevant?


wsmoses:
We shouldn't need to add attributes to intrinsic functions since they should get those automatically (for external functions those, they are needed).

PR #824, opened 2022-09-05
Fix inverted pointer use

wsmoses:


ZuseZ4 (approved):
maybe add a test if possible?

wsmoses:
Unfortunately this is in the external api usage, so a test can only be added on say the julia side.

PR #831, opened 2022-09-06
Simplify inactive function code

wsmoses:


enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -318,114 +318,204 @@ const char *DemangledKnownInactiveFunctionsStartingWith[] = {
...
-
-  auto Name = F->getName();
-
-  // Allocations, deallocations, and c++ guards don't impact the activity
-  // of arguments
-  if (isAllocationFunction(Name, TLI) || isDeallocationFunction(Name, TLI))
-    return true;
-  if (Name == "posix_memalign")
+  if (funcName == "free" || funcName == "_ZdlPv" || funcName == "_ZdlPvm" ||
+      funcName == "munmap") {
```
ZuseZ4, line 331: You don't seem to return true for memalign anymore (and the other 4 are new). Is that intended?

wsmoses: Per the comment the memalign allocates and thus isn't in the common code, but is in the inst variant.

PR #835, opened 2022-09-08
Add Symbolizer Path to CI

ZuseZ4:


ZuseZ4:
afaik that should always be available?

tgymnich (changes requested):
I think this will only apply to that one step. To be sure you could add it to the environment of the whole job:
```
jobs:
  build-linux:
    name: Enzyme CI LLVM ${{ matrix.llvm }} ${{ matrix.build }} ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    env:
      LLVM_SYMBOLIZER_PATH: /usr/bin/llvm-symbolizer-${{ matrix.llvm }}
```

For reference: https://docs.github.com/en/actions/learn-github-actions/environment-variables

Issue #840, opened 2022-09-12
Invalid Diffe index


wsmoses:
https://fwd.gymni.ch/AtuaRY

wsmoses:
https://fwd.gymni.ch/VeaUIh

Looks like we just don't have a good error message for too few args.

PR #846, opened 2022-09-14
Simplify and speed up warning/error API

wsmoses:
Fixes: https://github.com/EnzymeAD/Enzyme/issues/840

@tgymnich we should add warning/error tests like below for all new warnings (and eventually old ones).


vchuravy (approved):

ZuseZ4:
@wsmoses That broke openLB due to missing `_ZNSt7__cxx1115basic_stringbufIcSt11char_traitsIcESaIcEE7_M_syncEPcmm`, `@_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE10_M_replaceEmmPKcm`, `@_ZNSo5flushEv` and possibly more, still fixing them. I moved those in the ActivityAnalysis to the demangled names exactly to avoid having to add every single mangled function version. Can I just do the same here?

wsmoses:
Yes, but you will need to check that the functions also don’t free memory
(the marking of allocator is as inactive is why this was needed).

PR #854, opened 2022-09-27
Enable error propagation in MLIR AD

ftynse:
This enables error propagation in MLIR Enzyme flow and approximates
activity analysis by only treating provable constants as inactive.

### DIFF
--- a/enzyme/Enzyme/MLIR/Interfaces/GradientUtils.cpp
+++ b/enzyme/Enzyme/MLIR/Interfaces/GradientUtils.cpp
@@ -16,6 +16,7 @@
 
 // TODO: this shouldn't depend on specific dialects except Enzyme.
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
 
 using namespace mlir;
 using namespace mlir::enzyme;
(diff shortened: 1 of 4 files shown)


enzyme/Enzyme/MLIR/Interfaces/GradientUtils.cpp
```diff
@@ -295,13 +296,22 @@ void mlir::enzyme::MGradientUtils::forceAugmentedReturns() {
 
 LogicalResult MGradientUtils::visitChild(Operation *op) {
   if (mode == DerivativeMode::ForwardMode) {
+    // In absence of a proper activity analysis, approximate it by treating any
+    // side effect-free operation producing constants as inactive.
+    if (auto iface = dyn_cast<MemoryEffectOpInterface>(op)) {
```
wsmoses, line 301: For ease, can we push this into a isConstantOperation function within GradientUtils?

And then this simply becomes isConstantOperation && all(isConstantValue), without the memory effect check?

enzyme/Enzyme/MLIR/Interfaces/GradientUtils.cpp
```diff
@@ -126,6 +127,15 @@ mlir::enzyme::MGradientUtils::getNewFromOriginal(Operation *originst) const {
   return found->second;
 }
 
+bool mlir::enzyme::MGradientUtils::isConstantOperation(Operation *op) const {
+  // In absence of proper activity analysis, approximate it by checking that the
+  // operation doesn't write into memory.
+  auto iface = dyn_cast<MemoryEffectOpInterface>(op);
```
wsmoses, line 133: Resolved later, but this isn't necessarily true (eg. adding two active floats)

wsmoses (approved):

wsmoses:
I believe this functionality has since been merged in elsewhere, closing

PR #857, opened 2022-09-27
add VectorType with note

Wimmerer:


enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -4254,6 +4254,24 @@ void TypeAnalyzer::visitCallInst(CallInst &call) {
             llvm::errs() << *T << " - " << call << "\n";
             llvm_unreachable("Unknown type for libm");
           }
+        } else if (auto VT = dyn_cast<VectorType>(T)) {
+          // AFAICT there is no equivalent for VectorType?
```
wsmoses, line 4258: You can remove the note.

However can you add a test?

See here for a sample (except just have it call the function you ran into with corresponding type signature): https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/test/TypeAnalysis/cpuid.ll

wsmoses: Also in your case I think the issue is not in the argument (here). But in the return ( a few lines below on say 4280)

wsmoses:
bump @Wimmerer

wsmoses:
bump @rayegun

Issue #687, opened 2022-06-10
Support Opaque Pointers


ZuseZ4:
Motivated by https://discourse.llvm.org/t/cast-x-is-broken-implications-and-proposal-to-address/63033
I just build enzyme on top of llvm-head. Enzyme would compile fine and ninja check-enzyme only has 6 extra tests failing (on top of the old 6 ones). However enzyme-check-integration for me fails almost all tests:
```
Testing Time: 0.62s
  Passed           :  3
  Expectedly Failed:  4
  Failed           : 87
```
llvm-14 integration test work fine (except of those where I'm missing the deps).

Edit, my bad. I missed the opaque pointers have been switched on by default for 15, that's causing 67 of the failures.

ZuseZ4:
@tgymnich Do you have a few minutes to verify if it's only a local issue on my side?

wsmoses:
Yeah opaque pointer support is the likely cause here. Feel free to take a stab at it!

ZuseZ4:
https://discourse.llvm.org/t/rfc-better-support-for-typed-pointers-in-an-opaque-pointer-world/63339/3

PR #861, opened 2022-09-29
Remove some calls to `getPointerElementType` in `Utils.cpp`

thewilsonator:
As part of the Opaque pointer changes this function is no longer available in recent LLVMs.

see also #687 

I'll go through and remove all the easy to remove calls (i.e. ones that don't require refactoring), before moving to ones that will require refactoring.
I've just finished doing this process for another project, so I'd like to get this done while its still fresh.


enzyme/Enzyme/Utils.cpp
```diff
@@ -193,7 +193,7 @@ Function *getOrInsertExponentialAllocator(Module &M, Function *newFunc,
 
     Value *margs[] = {
 #if LLVM_VERSION_MAJOR > 7
-      B.CreateInBoundsGEP(gVal->getType()->getPointerElementType(), gVal,
+      B.CreateInBoundsGEP(gVal->getValueType(), gVal,
```
thewilsonator, line 196: oops this is not a `GlobalVariable`.

PR #863, opened 2022-09-29
Remove call to `getPointerElementType` in `CApi.cpp`

thewilsonator:
see #687

### DIFF
--- a/enzyme/Enzyme/CApi.cpp
+++ b/enzyme/Enzyme/CApi.cpp
@@ -646,8 +646,7 @@ LLVMValueRef EnzymeGradientUtilsCallWithInvertedBundles(
   auto callval = unwrap(func);
 
 #if LLVM_VERSION_MAJOR > 7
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/CApi.cpp
```diff
@@ -646,8 +646,7 @@ LLVMValueRef EnzymeGradientUtilsCallWithInvertedBundles(
   auto callval = unwrap(func);
 
 #if LLVM_VERSION_MAJOR > 7
-  auto res = BR.CreateCall(
-      cast<FunctionType>(callval->getType()->getPointerElementType()), callval,
+  auto res = BR.CreateCall(isa<CallBase>(callval)->getFunctionType(), callval,
```
wsmoses, line 649: callval is a literal value being called, not a callbase.

thewilsonator: Oh, I see. This will require an API change to fix then.

PR #866, opened 2022-09-30
Fix tests for LLVM15 and some more opaque pointers

wsmoses:


thewilsonator:
Awesome!

wsmoses:
There's definitely still a lot more to go, but this does some more. If you run check-enzyme-integration on LLVM15+ you'll see some more of the critical path ones to handle.

PR #932, opened 2022-11-20
Addition of SC'22 citation to Readme

ludgerpaehler:
Adding the  SC'22 citation to the README, plus adding a preceding pointer to with regards to when one should cite said paper.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -42,7 +42,7 @@ spack install enzyme
 
 To get involved or if you have questions, please join our [mailing list](https://groups.google.com/d/forum/enzyme-dev).
 
-If using this code in an academic setting, please cite the following two papers (first for Enzyme as a whole, second for GPU+optimizations, and third for the AD of Parallel Paradigms):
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -42,7 +42,7 @@ spack install enzyme
 
 To get involved or if you have questions, please join our [mailing list](https://groups.google.com/d/forum/enzyme-dev).
 
-If using this code in an academic setting, please cite the following two papers (first for Enzyme as a whole, then for GPU+optimizations):
+If using this code in an academic setting, please cite the following two papers (first for Enzyme as a whole, second for GPU+optimizations, and third for the AD of Parallel Paradigms):
```
wsmoses, line 45: three papers?

Also say third for AD of all other parallel programs (OpenMP, MPI, Julia Tasks, etc)

ludgerpaehler: Just addressed it, and thank you for catching the the leftover "two papers"!

PR #942, opened 2022-12-09
Simplify out TLI from entry

wsmoses:


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -1705,7 +1704,7 @@ class Enzyme final : public ModulePass {
                                    : cast<StructType>(aug->fn->getReturnType())
                                          ->getElementType(tapeIdx);
         unsigned idxs[] = {(unsigned)tapeIdx};
-        Value *tapeRes = (tapeIdx == -1)
```
tgymnich, line 1708: ?

wsmoses: I have no idea how that occurred. Good catch.

wsmoses:
So far only updated the Forward tests. Still need to do:
* ForwardModeVector
* BatchMode
* ReverseModeVector
* ReverseMode
* ForwardModeSplit
* Integration (clang), etc (Type/Activity/BCPass)

PR #943, opened 2022-12-12
Bench unit test compile time

tgymnich:


wsmoses:
@tgymnich CI is broken since you also need to change the CI command run to be the json one, in the relevant case.

tgymnich:
I think it is the file path for the json that is  wrong here...

wsmoses (approved):

PR #956, opened 2022-12-20
Update github actions

tgymnich:
node 12 actions are deprecated


wsmoses (approved):

tgymnich:
@wmoses we cannot update the benchmark action, since the open stack benchmark runners are using an older git version (git version 2.17.1). Checkout v3 falls back to the github REST API for git versions less than 2.18. The Github API then does not include the benchmarks since they are marked as `export-ignore` in `.gitattributes`.

PR #959, opened 2022-12-25
Fix shadow remat bug

wsmoses:


wsmoses:
@tgymnich this seems to fix the hand benchmark

tgymnich (approved):

PR #966, opened 2023-01-17
move TA into rustc

ZuseZ4:
We can't get the Rust Type Information from Debug Information,
thus we have to use [rustc_middle](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/index.html)::[ty](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/index.html) for the Rust abi anyway.
It's easier to just handle the C types in rustc too and keep the code in one place.
Let's merge this once we verified that this enzyme branch + rustc can handle all test cases.


ZuseZ4:
Those rules were apparently inactive by default and we never set `enzyme-rust-type`.
We already have basic testcases passing based on rustc-enzyme, and we know how to 
get the Type Information for the EnzymeTypeTree out of the TyCtxt, so let's clean this up
and work from there.

ZuseZ4:
We already have written TA out on the rust side but should keep this as potential Debug improvements, so closing.

PR #978, opened 2023-02-03
More opaque pointer support

wsmoses:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -11780,8 +11782,7 @@ class AdjointGenerator
       }
 
       assert(newcalled);
-      FunctionType *FT =
-          cast<FunctionType>(newcalled->getType()->getPointerElementType());
+      assert(FT);
```
tgymnich, line 11785: this assertion will never fail.

wsmoses: True, but I'd still like to have it as an assertion.

tgymnich (approved):

PR #990, opened 2023-02-12
Trace interface improvements

tgymnich:


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -1842,17 +1843,15 @@ class EnzymeBase {
...
     // Interface
-
-    Function *sample = nullptr;
-    for (auto &&interface_func : F->getParent()->functions()) {
-      if (interface_func.getName().contains("__enzyme_sample")) {
-        assert(interface_func.getFunctionType()->getNumParams() >= 3);
-        sample = &interface_func;
-      }
+    bool has_dynamic_interface = dynamic_interface != nullptr;
+    TraceInterface *interface;
```
wsmoses, line 1847: Make this a uique_ptr and then it'll get auto freed on return, perhaps?

wsmoses (approved):

enzyme/Enzyme/Enzyme.cpp

PR #991, opened 2023-02-12
Use getCalledFunction

tgymnich:


wsmoses (approved):
Perhaps assert that the result is non null?

PR #993, opened 2023-02-13
Extract handleArguments

tgymnich:
for future reuse in prob prog. Diff looks a bit wild event though I only extracted that one function.

Also replace some instance of llvm::Optional with std::optional.
https://discourse.llvm.org/t/deprecating-llvm-optional-x-hasvalue-getvalue-getvalueor/63716/38

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -23,6 +23,8 @@
 // the function passed as the first argument.
 //
 //===----------------------------------------------------------------------===//
+#include <optional>
+
 #include "SCEV/ScalarEvolution.h"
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -23,6 +23,8 @@
 // the function passed as the first argument.
 //
 //===----------------------------------------------------------------------===//
+#include <optional>
```
wsmoses, line 26: We should use the LLVM optional

tgymnich: https://discourse.llvm.org/t/deprecating-llvm-optional-x-hasvalue-getvalue-getvalueor/63716/38

wsmoses (approved):

PR #1005, opened 2023-02-23
Move custom handler to preserve nvvm

wsmoses:


tgymnich:
@wsmoses why does this belong to PreserveNVVM?

wsmoses:
It needs to run as the first pass, since otherwise things could get inlined/etc.

tgymnich:
Can we have a dedicated header for custom derivatives?

wsmoses:
Yes, but in a separate PR as this fixes the currently broken existing
interface.

PR #1014, opened 2023-03-06
Remove std optional and rewrite to use llvm/std compliant apis

wsmoses:


tgymnich:
LLVM is changing over to `std::optional` Why are we switching to `llvm::Optional`?

wsmoses:
std optional breaks macos on Yggdrassil for building Julia. I also rewrote the API usage here (using {} and * rather than has_value, etc).

We should then be able to do using Optional=std::optional above LLVM 16, while also not breaking macos julia.

tgymnich (approved):

PR #1028, opened 2023-03-13
Rewrite ReplaceOriginalCall

tgymnich:
This needed to be cleaned up

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -345,14 +345,9 @@ static Optional<StringRef> getMetadataName(llvm::Value *res) {
   }
 }
 
-static Value *adaptReturnedVector(CallInst *CI, Value *diffret,
+static Value *adaptReturnedVector(Value *ret, Value *diffret,
                                   IRBuilder<> &Builder, unsigned width) {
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -377,102 +372,99 @@ static Value *adaptReturnedVector(CallInst *CI, Value *diffret,
...
-        EmitFailure("IllegalReturnCast", CI->getDebugLoc(), CI,
-                    "Cannot cast return type of gradient ", *diffret->getType(),
-                    *diffret, ", to desired type ", *CI->getType());
-        return false;
-      }
-    } else {
+    CI->replaceAllUsesWith(cload);
+    CI->eraseFromParent();
+    return true;
+  }
```
wsmoses, line 450: Why does this need the mode != primal?

tgymnich: This if has been there for a [while](https://github.com/EnzymeAD/Enzyme/blob/25bf34184088629743bc85bb7d48bec716a36a2a/enzyme/Enzyme/Enzyme.cpp#L577). I tried to keep this change a non-functional change.

wsmoses (approved):

PR #1037, opened 2023-03-21
handle sdiv

ZuseZ4:


wsmoses (approved):
Sure but also would be good to see why TA didn't get this

PR #1040, opened 2023-03-22
reduce clang-14 warnings

ZuseZ4:


ZuseZ4:
@tgymnich only actual change is in [InstructionBatcher.cpp](https://github.com/EnzymeAD/Enzyme/pull/1040/files#diff-8f474a5d6163109845a750e234c464997a806c2d502f13c395e59aeed3a2ae1a), before it was UB I guess (but maybe worked by coincidence), now it would panic.

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -4124,6 +4124,9 @@ Constant *GradientUtils::GetOrCreateShadowFunction(
     return ConstantExpr::getPointerCast(GV, fn->getType());
   }
   }
+  llvm::errs() << "unhandled mode\n";
```
tgymnich, line 4127: why is this necessary? We should already handle all the cases.

wsmoses: @ZuseZ4 remove the llvm errs warning and change the llvm_unreachable text to "Illegal state: unknown mode for GetOrCreateShadowFunction

@tgymnich my guess is that some compilers cannot prove that we handle all cases of the enum, and this is seen as a path without a return/unreachable.

wsmoses (approved):

enzyme/Enzyme/GradientUtils.cpp
ZuseZ4: Indeed, for clang 14.0.6 it's not enough to handle all valid enum states in the places I updated.
At least this fixed the warning for it.

I'll adjust the message and correct tablegen tmrw.

PR #1046, opened 2023-03-22
Wrap `AdjointGenerator` in namespaces

tgymnich:
Removes `using namespace llvm` from header

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -42,7 +42,12 @@
 #include "TypeAnalysis/TBAA.h"
 
 #define DEBUG_TYPE "enzyme"
+
+namespace EnzymeLLVM {
+
 using namespace llvm;
+
+namespace Enzyme {
 
 // Helper instruction visitor that generates adjoints
 template <class AugmentedReturnType = AugmentedReturn *>
@@ -12810,3 +12815,8 @@ public:
                                         subretused);
   }
 };
+
+}; // namespace Enzyme
+}; // namespace EnzymeLLVM
+
+using namespace EnzymeLLVM::Enzyme;


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -42,8 +42,13 @@
 #include "TypeAnalysis/TBAA.h"
 
 #define DEBUG_TYPE "enzyme"
+
+namespace ADGenLLVM {
+
 using namespace llvm;
```
wsmoses, line 48: Can you try to put the using namespace llvm inside of the `class AdjointGenerator` instead?

E.g.

```
template <class AugmentedReturnType = AugmentedReturn *>
class AdjointGenerator
    : public llvm::InstVisitor<AdjointGenerator<AugmentedReturnType>> {
private:
using namespace llvm;
```

or is that not legal?

tgymnich: using namespace is not allowed on  a class level

wsmoses: I feel like ADGen is too specific if it's just going to contain this.

What if we just had the enzyme namespace or an anonymous namespace

tgymnich: It cannot be anonymous since we need to refer to it in our outer `using`

tgymnich: change names to `Enzyme` and `EnzymeLLVM`

Other options I can see are:
- Go through the whole thing an namespace all llvm symbols.
- Put the whole thing in a .cpp implementation file and implicitly instantiate the instances of the template we will use
- `using llvm::%symbol%` all the llvm symbols we use on a class level.

wsmoses: Would this not work?

```
wmoses@beast:~/git/Enzyme/enzyme ((HEAD detached at origin/main)) $ git diff
diff --git a/enzyme/Enzyme/AdjointGenerator.h b/enzyme/Enzyme/AdjointGenerator.h
index 32018269ad2..7da33d37222 100644
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -42,6 +42,9 @@
 #include "TypeAnalysis/TBAA.h"
 
 #define DEBUG_TYPE "enzyme"
+
+namespace {
+
 using namespace llvm;
 
 // Helper instruction visitor that generates adjoints
@@ -12810,3 +12813,5 @@ public:
                                         subretused);
   }
 };
+
+}
```

PR #1060, opened 2023-03-24
Loose ta

ZuseZ4:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2876,6 +2882,13 @@ class AdjointGenerator
             }
           }
         }
+      if (looseTypeAnalysis) {
+        forwardModeInvertedPointerFallback(BO);
```
tgymnich, line 2886: this does not belong here right?

ZuseZ4: Not strictly necessary rn, since I currently just use reverse mode. But I guess it doesn't hurt either, in case I run fwd mode later? Shall I remove it?

tgymnich: No that's fine. I was for a moment confused about `forwardModeInvertedPointerFallback`, but that seems fine now.

tgymnich (approved):

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -2253,6 +2253,12 @@ class AdjointGenerator
           }
         }
       }
+      if (looseTypeAnalysis) {
```
tgymnich, line 2256: maybe add forward version as well .

ZuseZ4: Just checked, we don't even support that instruction in Fwd mode.
So I'd leave that for when I actually use it since it probably deserves a new testcase too.

PR #1062, opened 2023-03-25
check hasVal() before getVal()

ZuseZ4:
Apparently this isn't always true.
TODO: find reason for it being None in the first place.


wsmoses (approved):
Yeah -- do you have a test case for this?

ZuseZ4:
I found it along with the other bugs lately. I guess once the TA is fixed I'll be able to reproduce and minimize this one more reliably, I'll leave this PR open till then.

wsmoses:
@ZuseZ4 this still seems useful and looks like it was approved and just waiting on a test for you.

Any reason you closed?

PR #1066, opened 2023-03-27
Fix ignoretype

wsmoses:
Will be necessary when external rules have opaque pointers for tape.

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -2503,7 +2503,7 @@ Value *GradientUtils::fixLCSSA(Instruction *inst, BasicBlock *forwardBlock,
 }
 
 Value *GradientUtils::cacheForReverse(IRBuilder<> &BuilderQ, Value *malloc,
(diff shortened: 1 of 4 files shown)


enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -2547,17 +2547,15 @@ Value *GradientUtils::cacheForReverse(IRBuilder<> &BuilderQ, Value *malloc,
...
+          llvm::errs() << "oldFunc: " << *oldFunc << "\n";
+          llvm::errs() << "newFunc: " << *newFunc << "\n";
+          llvm::errs() << "inst==malloc: " << *inst << "\n";
+          llvm::errs() << "ret: " << *ret << "\n";
         }
+        assert(inst->getType() == ret->getType());
+        if (replace)
+          inst->replaceAllUsesWith(UndefValue::get(ret->getType()));
         if (replace)
           erase(inst);
```
tgymnich, line 2560:
```suggestion
        if (replace) {
          inst->replaceAllUsesWith(UndefValue::get(ret->getType()));
          erase(inst);
        }
```

PR #1072, opened 2023-03-29
Add infrastructure for Fortran testing

l90lpa:
Extending testing to cover testing Fortran, currently only under the Intel Fortran compiler.


.github/workflows/fortran.yml
```diff
@@ -0,0 +1,55 @@
...
+    timeout-minutes: 60 
+    steps:
+    - name: add llvm
+      run: |
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+        sudo apt-add-repository "deb http://apt.llvm.org/`lsb_release -c | cut -f2`/ llvm-toolchain-`lsb_release -c | cut -f2`-${{ matrix.llvm }} main" || true
+        sudo apt-get install -y autoconf cmake gcc g++ libtool gfortran llvm-${{ matrix.llvm }}-dev ninja-build pip
+        sudo python3 -m pip install --upgrade pip setuptools
+        sudo python3 -m pip install lit
+        sudo touch /usr/lib/llvm-${{ matrix.llvm }}/bin/yaml-bench
```
tgymnich, line 33: This LLVM should not be needed. We need to build Enzyme against the intel llvm

tgymnich: We might also be able to build against the fortran compiler.

wsmoses: @chelini do you know the ifx folks by chance?

enzyme/test/Fortran/ReverseMode/square.f90
```diff
@@ -0,0 +1,38 @@
+! RUN: ifx -flto -O0 -c %s | %opt - %loadEnzyme -enzyme
+! RUN: ifx -flto -O1 -c %s | %opt - %loadEnzyme -enzyme
+! RUN: ifx -flto -O2 -c %s | %opt - %loadEnzyme -enzyme
+! RUN: ifx -flto -O3 -c %s | %opt - %loadEnzyme -enzyme
+
```
wsmoses, line 5: Can you add a check line? Eg maybe through lli or run an executable?

wsmoses:
Yes, that would be useful. I think we need them to enable extensions and/or
allow us to build against their headers.

wsmoses:
We can also move this to an email convo with them if easier.

enzyme/test/Fortran/ReverseMode/square.f90
l90lpa: Yep I'll do that

l90lpa:
@wsmoses yes moving this to email with them sounds good. I'll reach out tomorrow.

.github/workflows/fortran.yml
chelini: > k a look at ifx and it does not seem like it supports llvm plugins. So we don't need to build it. We only need to build against the llvm used to compile opt or lld (whic

@wsmoses no but I can ask around. What do you need?

enzyme/test/Fortran/ReverseMode/square.f90
l90lpa: @wsmoses I think there is also a bug in what I'm doing above. So to fix that and add more checking I propose,
```
! RUN: ifx -flto -O0 -c  %s -o %t && %opt %t %loadEnzyme -enzyme -o %t1 && ifx -flto -O0 %t1 -o %t2 && %t2 | FileCheck %s
```
...
```
! CHECK: 9
! CHECK-NEXT: 6
``` 
Its more file heavy, but I can't find a way to have ifx compile from stdin. Also I went for running the exe since I'm expecting lli not to understand the intrinsics/runtime library functions later down the line when tests including those arise.

l90lpa:
@wsmoses I reached out to the person I know on the IFX team (regarding plugin support and headers etc), and so I'm waiting on hearing back from them once they're back in the office. 

Regarding this PR, do you want to wait until we hear back from them before we do anything further with it. Or, do you think whilst we wait we could get this PR merged in, and then I can come back later to modify the build/tests if they're able to provide us with what we are after?

.github/workflows/fortran.yml
l90lpa: I'm resolving this conversation as we have decided to use standard LLVM for now.

enzyme/test/Fortran/ReverseMode/square.f90
l90lpa: Good question. Yes, ifx can print to stdout either by using `-o /dev/stdout` or I can use process substitution `ifx ... -o >(opt ...)`. I need to check if the second is valid in LIT. So I could replace the first temporary file.

l90lpa: Also just to add to that, it doesn't support `-o - ` like in clang.

PR #1090, opened 2023-04-06
New PM clang

wsmoses:
The new pass manager pipeline has fewer available plugin points to inject an optimization pass (enzyme) at its previous location (right before vectorization). This means that existing optimizations will not run before Enzyme on the new pm.

Support for better plugin points has been improving over time (aka adding new plugin injection points). Unfortunately, they aren't available on all LLVM versions. This remedies this, by forcibly running the passes that would've occured in the pipeline before vectorization and after the point Enzyme can be injected in the New PM. As new LLVM's have more plugin point options, this manual list diminishes over LLVM version time.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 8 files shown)


LeilaGhaffari (approved):
Works for me, thanks!

tgymnich (approved):

PR #1103, opened 2023-04-10
add vector fmax

wsmoses:


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4142,6 +4142,27 @@ class AdjointGenerator
         }
         return;
       }
+#if LLVM_VERSION_MAJOR >= 12
+      case Intrinsic::vector_reduce_fmax: {
+        if (vdiff && !gutils->isConstantValue(orig_ops[0])) {
+          auto prev = lookup(gutils->getNewFromOriginal(orig_ops[0]), Builder2);
+          auto lhs = Builder2.CreateExtractElement(prev, (uint64_t)0);
```
tgymnich, line 4149: @wsmoses why are we only handling vectors of size 2? Can we add an assertion here at least?

wsmoses: That is a fair point and one I didn't realize that was possible, making a PR.

tgymnich: #1110

PR #1104, opened 2023-04-10
Fix/enable later llvm version integration tests

wsmoses:


.github/workflows/ccpp.yml
```diff
@@ -54,7 +54,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        llvm: ["7", "8", "9", "10", "11", "12", "13", "14", "15"]
+        llvm: ["7", "8", "9", "10", "11", "12", "13"]
```
tgymnich, line 57: why are u disabling tests?

tgymnich:
#1048

.github/workflows/ccpp.yml
wsmoses: It was mistakenly added last commit (automerged without realizing). These arent easily installable on ubuntu 18.04, so worth removing (since they are tested on 20.04)

tgymnich: I missed the fact that those changes only apply to 18

PR #1106, opened 2023-04-11
compiles with llvm-16

stepasite:
* port to llvm-16 (not backward-compatible with older llvms), WIP

* wip, but probably not correct on may places

* rule out old PM on llvm 16 and above

* #if LLVM_VERSION_MAJOR >= 16, WIP

* #if LLVM_VERSION_MAJOR >= 16, WIP

* #if LLVM_VERSION_MAJOR >= 16, WIP

* #if LLVM_VERSION_MAJOR >= 16, WIP

* #if LLVM_VERSION_MAJOR >= 16, WIP

* #if LLVM_VERSION_MAJOR >= 16


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -189,8 +189,13 @@ class AdjointGenerator
     AL =
         AL.addParamAttribute(DT->getContext(), 1, Attribute::AttrKind::NonNull);
 #if LLVM_VERSION_MAJOR >= 14
+  #if LLVM_VERSION_MAJOR >= 16
+    AL = AL.addAttributeAtIndex(DT->getContext(), AttributeList::FunctionIndex,
+                                Attribute::AttrKind::Memory);
```
wsmoses, line 194: is this equivalent to argmemonly?

wsmoses:
Forward mode has new pass manager tests (in addition to old pm ones). Inside the llvm lit config, at or above LLVM 16 can we get away with setting the loadEnzyme equivalent pieces to use the new pm (we do this for clangEnzyme tests)?

If not, we can version lock all the test on old pm / new pm (e.g. if [%llvmver < 16 ] oldpm lit config;

enzyme/Enzyme/AdjointGenerator.h
stepasite: trying to fix, please check

stepasite:
> Forward mode has new pass manager tests (in addition to old pm ones). Inside the llvm lit config, at or above LLVM 16 can we get away with setting the loadEnzyme equivalent pieces to use the new pm (we do this for clangEnzyme tests)?
> 
> If not, we can version lock all the test on old pm / new pm (e.g. if [%llvmver < 16 ] oldpm lit config;
 
unable to grasp your idea/question, sorry

wsmoses:
> > Forward mode has new pass manager tests (in addition to old pm ones). Inside the llvm lit config, at or above LLVM 16 can we get away with setting the loadEnzyme equivalent pieces to use the new pm (we do this for clangEnzyme tests)?
> > If not, we can version lock all the test on old pm / new pm (e.g. if [%llvmver < 16 ] oldpm lit config;
> 
> unable to grasp your idea/question, sorry

We can potentially version gate the old/new pass manager pipelines: https://github.com/EnzymeAD/Enzyme/blob/25715346011f7fb8953eb38d3bf1307a2088b256/enzyme/test/Integration/ReverseMode/omp.c#L2

Addtionally, forward mode already has both new and old PMs. https://github.com/EnzymeAD/Enzyme/blob/25715346011f7fb8953eb38d3bf1307a2088b256/enzyme/test/Enzyme/ForwardMode/add.ll

The reverse mode LLVM tests will require a bit more work to set up, though. But getting the LLVM tests running can be a separate PR.

After https://github.com/EnzymeAD/Enzyme/pull/1104 lands can we try to enable C++ integration test CI (which should already be set up for the new pm infrastructure).

stepasite:
> 

OK, thanks. Hopefully got the first idea to disable old PM tests with llvm 16.
--

Here:

https://github.com/EnzymeAD/Enzyme/blob/25715346011f7fb8953eb38d3bf1307a2088b256/enzyme/test/Enzyme/ForwardMode/add.ll

I found this `; RUN: %opt < %s %newLoadEnzyme -passes="enzyme" -enzyme-preopt=false -S | FileCheck %s` but not sure about `%newLoadEnzyme`. Using `-load=LLVMEnzyme-16.so` simply does not work and I am getting error: _opt: unknown pass name 'enzyme'_

wsmoses:
See https://github.com/EnzymeAD/Enzyme/blob/0edca6259ef3f2eb4b90292fdb99cff34cadf72f/enzyme/test/lit.site.cfg.py.in#L55

stepasite:
> See
> 
> https://github.com/EnzymeAD/Enzyme/blob/0edca6259ef3f2eb4b90292fdb99cff34cadf72f/enzyme/test/lit.site.cfg.py.in#L55

Thanks, I will give it a try.

Anyway, I conditioned all the tests where an alternative test invoked via new pm exists.

If you are ok with it, I can add the same condition also to tests where only old pm is used.

wsmoses:
That would work. However, if it's not too much, we really should add a new pass manager line to the other tests.

wsmoses:
Also we should add LLVM16 to the CI matrix here https://github.com/EnzymeAD/Enzyme/blob/main/.github/workflows/ccpp.yml and https://github.com/EnzymeAD/Enzyme/blob/main/.github/workflows/enzyme-ci.yml

enzyme/test/Enzyme/ForwardMode/div.ll
```diff
@@ -1,4 +1,4 @@
-; RUN: %opt < %s %loadEnzyme -enzyme -instsimplify -enzyme-preopt=false -S | FileCheck %s
+; RUN: %if [ %llvmver -lt 16 ]; then  < %s %loadEnzyme -enzyme -instsimplify -enzyme-preopt=false -S | FileCheck %s; fi
```
wsmoses, line 1: You accidentally removed the %opt here.

stepasite: thanks, fixed, resolving

stepasite:
> That would work. However, if it's not too much, we really should add a new pass manager line to the other tests.

OK, I see. But before doing this, I would like to find out why I get zeros instead of getting proper derivatives. No issues with llvm-15, but problems arise with llvm-16 build; there is something wrong within my changes.

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -1789,9 +1824,14 @@ class EnzymeBase {
 #endif
     }
 
+#if LLVM_VERSION_MAJOR >= 16
     bool status = HandleAutoDiff(
```
wsmoses, line 1828: You need to have the bool status = HandleAutodiff( as well in the other version.

stepasite: thanks, fixed

enzyme/Enzyme/Utils.h
```diff
@@ -1051,7 +1051,11 @@ template <typename T> static inline llvm::StringRef getFuncNameFromCall(T *op) {
 }
 
 template <typename T>
+#if LLVM_VERSION_MAJOR >= 16
+static inline std::optional<size_t> getAllocationIndexFromCall(T *op) {
```
wsmoses, line 1055: Can you put the curly braces after the ifdef? It confuses some editors otherwise.

stepasite: done

enzyme/test/Enzyme/ForwardMode/div.ll
```diff
@@ -1,4 +1,4 @@
-; RUN: %opt < %s %loadEnzyme -enzyme -instsimplify -enzyme-preopt=false -S | FileCheck %s
+; RUN: %if [ %llvmver -lt 16 ]; then %opt < %s %loadEnzyme -enzyme -instsimplify -enzyme-preopt=false -S | FileCheck %s; fi
```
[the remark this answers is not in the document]
stepasite: thanks, done

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -2824,7 +2937,10 @@ llvmGetPassPluginInfo() {
 #endif
               FunctionPassManager OptimizerPM;
               FunctionPassManager OptimizerPM2;
-#if LLVM_VERSION_MAJOR >= 14
+#if LLVM_VERSION_MAJOR >= 16
+              OptimizerPM.addPass(llvm::GVNPass());
+              OptimizerPM.addPass(llvm::SROAPass(llvm::SROAOptions::PreserveCFG));
```
stepasite, line 2942: @wsmoses Not sure what setting should be passed here: ModifyCFG/PreserveCFG, could you please check?

stepasite:
@wsmoses Before pushing it into this PR, could you please check the PR into my main?

https://github.com/stepasite/Enzyme/pull/11

I am enabling llvm-16 in gha. Also trying to format code with the following:

`docker run -it --rm --workdir /Enzyme -v $(pwd):/Enzyme clang-format-lint --clang-format-executable /clang-format/clang-format11 -r --exclude ./Enzyme/SCEV --style llvm --inplace true ./Enzyme`

The result of formatting step is strange as there are reformatted parts in `Enzyme.cpp` that I haven't touched at all. Any idea what is going on?

wsmoses:
Something odd is still hapenning int he PR since per CI above, the earlier version tests are failing.

stepasite:
> Something odd is still hapenning int he PR since per CI above, the earlier version tests are failing.

Yes, there is something wrong. Hopefully, I will have enough time to go through the changes through the weekend.

stepasite:
@wsmoses I am getting _unknown function pass 'simplifycfg'_ and _unknown function pass 'inline'_ with the new pm when passing those phases into `function()` option. Could you please advise?

wsmoses:
Inlining is a module level pass so put it on its own outside the function.

wsmoses:
Apparently simplifycfg was spelled "simplifycfg" in the new PM on LLVM 11 +, and "simplify-cfg" before.

Perhaps for ease, we add that pass name as a custom replacement rule in here https://github.com/EnzymeAD/Enzyme/blob/a72f89e6afd9eebc2df5cad07f1e17ec4bc51231/enzyme/test/lit.site.cfg.py.in#L53 ?

stepasite:
_/usr/lib/llvm-15/bin/opt: unknown function pass 'early-cse-memssa'_
_/usr/lib/llvm-15/bin/opt: unknown function pass 'inline'_

I am wondering what is going on, both passes should exist, any idea @wsmoses ?

wsmoses:
@jdoerfert @tgymnich can you assist @stepasite ?

wsmoses:
Finished this up and merged in https://github.com/EnzymeAD/Enzyme/pull/1207, thanks for your help!!

stepasite:
> Finished this up and merged in #1207, thanks for your help!!

Great to see llvm 16 is supported. Anyway, looking at all the changes you had to do, the cake was too big for my first bite. Thanks.

PR #1125, opened 2023-04-21
actually run some postopt

wsmoses:


martinjm97 (approved):

martinjm97 (approved):

ZuseZ4:
can't we have this by default  be on?
I think in most cases people won't notice a hidden flag

wsmoses:
The issue is determining what optimization level was previously set (eg O0, O1, etc). We do not want to run at O0, we want the same O1 behavior at O1, etc.

wsmoses:
To be clear @zuseZ4 we should have it auto determined by default, but didn't finding it quickly for this PR. Help is welcome.

ZuseZ4:
My assumption was just that we can determine that for all languages and cases somewhat reliable and overwrite it, except of LLDEnzyme. And for LLD I'd prefer to default to opt=2, since it's already the hardest to set up, so I expect people to forget or not know about this flag. I can look into LLD tomorrow though, if you want.

wsmoses:
Also to be clear optimizations after Enzyme in the pipeline will still be run, so this is really primarily for nested differentiation.

PR #1134, opened 2023-04-24
Added reverse mode adjoint for linalg.generic

umatin:
This has been tested on the MLIR-Reverse fork without activity analysis. With activity analysis I am not able to compile differentiate any code containing memrefs. This would probably have to be fixed before this can be merged.


enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+    Region *newOpRegion = newOp.getBlock()->getParent();
+    int numInputsNewOp = cast<linalg::GenericOp>(newOp).getInputs().size();
+    Region *adjointRegion = &adjoint.getBodyRegion();
+    int numInputsAdjoint = adjoint.getInputs().size();
+    Location loc = op->getLoc();
+    int numCaches = 0;
+    SmallVector<Value> pushCaches;
+
+    // std::function<std::pair<Value, Value>(Type)> hook = nullptr;
+    std::function<std::pair<Value, Value>(Type)> hook =
```
tgymnich, line 138:
```suggestion
    auto hook =
```
I would rather use `auto` here so that the compiler could avoid creating a `std::function` object.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+      body->addArgument(opOperand.getType(), opOperand.getLoc());
+    }
+    OpBuilder builderAdd(yieldOp);
+    for (auto it : llvm::enumerate(yieldOp.getOperands())) {
+      Value arg = body->getArgument(outputs.size() + numCaches + it.index());
+      auto diffeType = cast<AutoDiffTypeInterface>(arg.getType());
+      Value grad = diffeType.createAddOp(builderAdd, it.value().getLoc(), arg,
+                                         it.value());
+      yieldOp.setOperand(it.index(), grad);
+    }
```
tgymnich, line 176:
```suggestion
    for (auto [index, op] : llvm::enumerate(yieldOp.getOperands())) {
      Value arg = body->getArgument(outputs.size() + numCaches + index);
      auto diffeType = cast<AutoDiffTypeInterface>(arg.getType());
      Value grad = diffeType.createAddOp(builderAdd, op.getLoc(), arg, op);
      yieldOp.setOperand(index, grad);
    }
```

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+      retrievedValue = invertMemref(retrievedValue, builder2, op->getLoc());
+      adjoint.getInputsMutable().append(ValueRange({retrievedValue}));
+      indexing_maps_adjoint.insert(
+          indexing_maps_adjoint.begin() + numInputsAdjoint + i,
+          AffineMap::getMultiDimIdentityMap(iterationDomains.size(),
+                                            builder2.getContext()));
+    }
+    SmallVector<Attribute> indexing_maps_attr;
+    SmallVector<Attribute> indexing_maps_attr_adjoint;
+    for (auto map : indexing_maps) {
```
tgymnich, line 207:
```suggestion
    for (auto &map : indexing_maps) {
```
maybe make this a reference

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+      indexing_maps_attr.push_back(AffineMapAttr::get(map));
+    }
+    for (auto map : indexing_maps_adjoint) {
```
tgymnich, line 210:
```suggestion
    for (auto &map : indexing_maps_adjoint) {
```

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -122,14 +124,15 @@ struct ForOpInterfaceReverse
         gutils->popCache(caches[2], builder), nArgs); // TODO
     repFor.getRegion().begin()->erase();
 
-    buildReturnFunction buildFuncReturnOp = [](OpBuilder &builder, Location loc,
-                                               SmallVector<Value> retargs) {
-      builder.create<scf::YieldOp>(loc, retargs);
-      return;
-    };
+    std::function<buildReturnFunction> buildFuncReturnOp =
```
tgymnich, line 127:
```suggestion
    auto buildFuncReturnOp =
```

enzyme/Enzyme/MLIR/Interfaces/EnzymeLogicReverse.cpp
```diff
@@ -381,13 +382,13 @@ FunctionOpInterface MEnzymeLogic::CreateReverseDiff(
   Region &oldRegion = gutils->oldFunc.getFunctionBody();
   Region &newRegion = gutils->newFunc.getFunctionBody();
 
-  buildReturnFunction buildFuncReturnOp = [](OpBuilder &builder, Location loc,
-                                             SmallVector<Value> retargs) {
+  brf buildFuncReturnOp = [](OpBuilder &builder, Location loc,
```
tgymnich, line 385:
```suggestion
  auto buildFuncReturnOp = [](OpBuilder &builder, Location loc,
```

enzyme/Enzyme/MLIR/Interfaces/GradientUtilsReverse.cpp
```diff
@@ -72,10 +72,34 @@ Type mlir::enzyme::MGradientUtilsReverse::getCacheType(Type t) {
   return cacheType;
 }
 
+void MGradientUtilsReverse::registerCacheCreatorHook(
```
tgymnich, line 75: It may make sense to template the closure parameter in order to avoid creating a `std::function` object. `template <typename Fun>`

enzyme/Enzyme/MLIR/Interfaces/GradientUtilsReverse.cpp
```diff
@@ -72,10 +72,34 @@ Type mlir::enzyme::MGradientUtilsReverse::getCacheType(Type t) {
...
+    std::function<std::pair<Value, Value>(Type)> hook) {
+  if (hook != nullptr) {
+    cacheCreatorHook.push_back(hook);
+  }
+}
+
+void MGradientUtilsReverse::deregisterCacheCreatorHook(
+    std::function<std::pair<Value, Value>(Type)> hook) {
```
tgymnich, line 83: It may make sense to template the closure parameter in order to avoid creating a `std::function` object. `template <typename Fun>`

enzyme/Enzyme/MLIR/Interfaces/GradientUtilsReverse.h
```diff
@@ -79,6 +81,13 @@ class MGradientUtilsReverse {
   Type getIndexType();
   Value insertInit(Type t);
 
+  SmallVector<std::function<std::pair<Value, Value>(Type)>> cacheCreatorHook;
+  void
+  registerCacheCreatorHook(std::function<std::pair<Value, Value>(Type)> hook);
+  void
+  deregisterCacheCreatorHook(std::function<std::pair<Value, Value>(Type)> hook);
+  std::pair<Value, Value> getNewCache(Type t);
```
tgymnich, line 89: It may make sense to template the closure parameter in order to avoid creating a `std::function` object. `template <typename Fun>`

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+      shapes.push_back(-1);
+    }
+
+    SmallVector<Value> inputs, outputs;
+    SmallVector<AffineMap> indexingMaps;
+    SmallVector<StringRef> iteratorTypes{linalgOp.getNumLoops(),
+                                         getParallelIteratorTypeName()};
+
+    for (OpOperand *output : linalgOp.getOutputOperands()) {
+      if (gutils->hasInvertPointer(output->get())) {
```
ftynse, line 100: Nit: use `continue` here to simplify code https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+        Value view = invertMemref(inp, builder, op->getLoc());
+        inputs.push_back(view);
+      }
+    }
+
+    linalg::GenericOp adjoint = builder.create<linalg::GenericOp>(
+        op->getLoc(), outputs, inputs, indexingMaps, iteratorTypes);
+
+    int numInputs = inputs.size();
+    std::function<buildReturnFunction> buildFuncReturnOp =
```
ftynse, line 121: `std::function` constructor performs dynamic allocation, which is bad for performance. Use `llvm::function_ref` or keep lambdas as `auto` unless (1) they are recursive or (2) the function-typed variable will outlive the function itself.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+         numInputsNewOp, numInputsAdjoint, &pushCaches = pushCaches](Type t) {
+          OpBuilder builder(newOpRegion);
+          Value pushCache = builder.create<enzyme::InitOp>(loc, t);
+          pushCaches.push_back(pushCache);
+          newOpRegion->addArgument(t, loc);
+
+          Value popCache = adjointRegion->insertArgument(
+              numInputsAdjoint + numCaches, t, loc);
+          numCaches++;
+          return std::pair<Value, Value>(pushCache, popCache);
```
ftynse, line 149: Nit: `std::make_pair(pushCache, popCache);` is more readable.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+        };
+
+    gutils->Logic.differentiate(
+        gutils, *linalgOp.getBlock()->getParent(), adjoint.getBodyRegion(),
+        /*parentRegion=*/false, buildFuncReturnOp, hook);
+
+    // TODO add pushCaches to the yield in newOp
+    auto newOpYield = cast<linalg::YieldOp>(
+        cast<linalg::GenericOp>(newOp).getBodyRegion().front().getTerminator());
+    for (auto pc : pushCaches) {
```
ftynse, line 159: MLIR uses `auto` when the type is obvious from context (e.g., the RHS is a cast or a build) or difficult/impossible to spell (iterators, lambdas). Spell out the type in all the other cases as it will likely improve readability.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    Block *body = &(adjoint.getBodyRegion().front());
+    auto yieldOp = cast<linalg::YieldOp>(body->getTerminator());
+
+    for (auto opOperand : yieldOp.getOperands()) {
+      body->addArgument(opOperand.getType(), opOperand.getLoc());
+    }
+    OpBuilder builderAdd(yieldOp);
+    for (auto it : llvm::enumerate(yieldOp.getOperands())) {
```
ftynse, line 170: Use `auto &&` for `llvm::enumerate`, this *will* break on version update.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+      Value arg = body->getArgument(outputs.size() + numCaches + it.index());
+      auto diffeType = cast<AutoDiffTypeInterface>(arg.getType());
+      Value grad = diffeType.createAddOp(builderAdd, it.value().getLoc(), arg,
+                                         it.value());
+      yieldOp.setOperand(it.index(), grad);
+    }
+
+    auto indexing_maps = newOp.getIndexingMapsArray();
+    auto indexing_maps_adjoint = adjoint.getIndexingMapsArray();
```
ftynse, line 179: Nit: be consistent with naming, all the code around uses `camelBack`.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+    for (int i = 0; i < numCaches; i++) {
+      Value cacheArg = body->getArgument(outputs.size() + i);
+
+      Type ct = cacheArg.getType();
+      Type type = MemRefType::get(ArrayRef(shapes), ct);
+      Value alloc = cacheBuilder.create<memref::AllocOp>(
+          op->getLoc(), type, ValueRange(iterationDomains));
+      Value cache = gutils->initAndPushCache(alloc, cacheBuilder);
+      alloc.getDefiningOp()->setAttr("operand_segment_sizes",
+                                     cacheBuilder.getDenseI32ArrayAttr({1, 0}));
```
ftynse, line 189: I'm sure there is a better way of setting attributes than using string-based setters. Furthermore, if the code above hadn't prematurely casted `memref::AllocaOp` into a `Value`, this line wouldn't need to supriously call `getDefiningOp` and would be able to do something like `alloc.getFooOperandsMutable().append()`.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+      cast<linalg::GenericOp>(newOp).getOutputsMutable().append(
+          ValueRange({alloc}));
+      indexing_maps.push_back(AffineMap::getMultiDimIdentityMap(
+          iterationDomains.size(), cacheBuilder.getContext()));
+
+      OpBuilder builder2(adjoint);
```
ftynse, line 196: It is usually a bad practice to have separate builders. Prefer one builder and use `OpBuilder::InsertionGuard` RAII object to control insertion points.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+    for (auto map : indexing_maps) {
+      indexing_maps_attr.push_back(AffineMapAttr::get(map));
+    }
+    for (auto map : indexing_maps_adjoint) {
+      indexing_maps_attr_adjoint.push_back(AffineMapAttr::get(map));
+    }
+    newOp->setAttr("indexing_maps",
+                   cacheBuilder.getArrayAttr(indexing_maps_attr));
+    adjoint->setAttr("indexing_maps",
+                     builder.getArrayAttr(indexing_maps_attr_adjoint));
```
ftynse, line 216: Same as above, I'm sure there is a better way that doesn't rely on hardcoding attribute names as strings. Also, this **will** break on a version bump because these attributes become properties.

enzyme/Enzyme/MLIR/Passes/SimplifyMemrefCache.cpp
```diff
@@ -0,0 +1,181 @@
...
+    : public enzyme::SimplifyMemrefCachePassBase<SimplifyMemrefCachePass> {
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+
+    getOperation()->walk([&](Operation *op) {
+      if (auto initOp = dyn_cast<enzyme::InitOp>(op)) {
+
+        if (auto c1 = dyn_cast<enzyme::CacheType>(initOp.getType())) {
+          if (auto memref = dyn_cast<MemRefType>(c1.getType())) {
+            if (auto c2 =
```
ftynse, line 33: This is a particularly egregious example of unnecessarily right-shifted code. Please follow https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,237 @@
...
+    SmallVector<StringRef> iteratorTypes{linalgOp.getNumLoops(),
+                                         getParallelIteratorTypeName()};
+
+    for (OpOperand *output : linalgOp.getOutputOperands()) {
+      if (gutils->hasInvertPointer(output->get())) {
+        indexingMaps.push_back(linalgOp.getMatchingIndexingMap(output));
+        Value out = gutils->invertPointerM(output->get(), builder);
+        Value view = invertMemref(out, builder, op->getLoc());
+        outputs.push_back(view);
+      }
```
tgymnich, line 105:
```suggestion
      if (!gutils->hasInvertPointer(output->get())) {
        continue;
      }
      indexingMaps.push_back(linalgOp.getMatchingIndexingMap(output));
      Value out = gutils->invertPointerM(output->get(), builder);
      Value view = invertMemref(out, builder, op->getLoc());
      outputs.push_back(view);
```

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    // get iteration domain
+    AffineMap aMap = newOp.getShapesToLoopsMap();
+    SmallVector<Value> dims;
+    for (OpOperand *input : newOp.getInputOperands()) {
+      auto shape = cast<MemRefType>(input->get().getType()).getShape();
+      for (int i = 0; i < shape.size(); i++) {
+        auto dimI =
+            cacheBuilder.create<arith::ConstantIndexOp>(op->getLoc(), i);
+        // input->get().getParentBlock()->dump();
```
tgymnich, line 79: remove commented out code?
```suggestion
```

enzyme/Enzyme/MLIR/Interfaces/GradientUtilsReverse.cpp
```diff
@@ -267,21 +291,23 @@ bool mlir::enzyme::MGradientUtilsReverse::hasInvertPointer(mlir::Value v) {
...
+    if (diffe_type == DIFFE_TYPE::OUT_DIFF) {
+      if (auto iface = dyn_cast<AutoDiffTypeInterface>(val.getType())) {
+        Value zero = iface.createNullValue(initializationBuilder, val.getLoc());
+        mapInvertPointer(val, zero, initializationBuilder);
+      } else {
+        llvm_unreachable(
+            "Type does not have an associated AutoDiffTypeInterface");
+      }
     }
   }
```
tgymnich, line 312:
```suggestion
  for (const auto &[val, diffe_type] : llvm::zip(
           this->oldFunc.getFunctionBody().getArguments(), argDiffeTypes)) {
    if (diffe_type != DIFFE_TYPE::OUT_DIFF) {
      continue;
    }
    auto iface = dyn_cast<AutoDiffTypeInterface>(val.getType());
    if (!iface) {
      llvm_unreachable("Type does not have an associated AutoDiffTypeInterface");
    }
    Value zero = iface.createNullValue(initializationBuilder, val.getLoc());
    mapInvertPointer(val, zero, initializationBuilder);
  }
```

enzyme/Enzyme/MLIR/Passes/SimplifyMemrefCache.cpp
```diff
@@ -0,0 +1,182 @@
...
+        linalgOp.getRegion().eraseArgument(outputIndex);
+        linalgOp.getRegion().insertArgument(outputIndex, c2.getType(),
+                                            output.value().getLoc());
+
+        Value cache = linalgOp.getRegion().front().getTerminator()->getOperand(
+            (unsigned)output.index());
+        for (auto user : cache.getUsers()) {
+          if (auto popOp = dyn_cast<enzyme::PopOp>(user)) {
+            llvm_unreachable("PopOp should not be used in forward pass");
+          }
```
tgymnich, line 66:
```suggestion
          if (isa<enzyme::PopOp>(user)) {
            llvm_unreachable("PopOp should not be used in forward pass");
          }
```

enzyme/Enzyme/MLIR/Passes/SimplifyMemrefCache.cpp
```diff
@@ -0,0 +1,182 @@
...
+    OpBuilder allocBuilder(allocOp);
+    auto newAllocOp = allocBuilder.create<memref::AllocOp>(
+        allocOp.getLoc(), dyn_cast<MemRefType>(newType),
+        allocOp.getDynamicSizes(), allocOp.getSymbolOperands(),
+        allocOp.getAlignmentAttr());
+
+    for (auto user : allocOp->getUsers()) {
+      auto linalgOp = dyn_cast<linalg::GenericOp>(user);
+      if (auto pushOp = dyn_cast<enzyme::PushOp>(user)) {
+        continue;
```
tgymnich, line 42:
```suggestion
      if (isa<enzyme::PushOp>(user)) {
        continue;
```

enzyme/Enzyme/MLIR/Interfaces/GradientUtilsReverse.cpp
```diff
@@ -72,10 +72,34 @@ Type mlir::enzyme::MGradientUtilsReverse::getCacheType(Type t) {
...
+
+// We assume that caches will only be written to at one location. The returned
+// cache is (might be) "pop only"
 Value MGradientUtilsReverse::initAndPushCache(Value v, OpBuilder &builder) {
-  Value cache = insertInit(getCacheType(v.getType()));
-  builder.create<enzyme::PushOp>(v.getLoc(), cache, v);
-  return cache;
+  std::pair<Value, Value> cache = getNewCache(getCacheType(v.getType()));
+  auto pushOp = builder.create<enzyme::PushOp>(v.getLoc(), cache.first, v);
+  return cache.second;
```
tgymnich, line 102:
```suggestion
  auto [pushCache, popCache] = getNewCache(getCacheType(v.getType()));
  auto pushOp = builder.create<enzyme::PushOp>(v.getLoc(), pushCache, v);
  return popCache;
```

tgymnich (approved):

ftynse (commented):
There is one place where I suspect the adjoint generator produces invalid IR. Otherwise looks good. Some tests would be beneficial.

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,256 @@
...
+}
+
+template <typename T_>
+struct GenericOpInterfaceReverse
+    : public ReverseAutoDiffOpInterface::ExternalModel<
+          GenericOpInterfaceReverse<T_>, T_> {
+  void createReverseModeAdjoint(Operation *op, OpBuilder &builder,
+                                MGradientUtilsReverse *gutils,
+                                SmallVector<Value> caches) const {
+    // TODO: currently in progress, this doesn't work
```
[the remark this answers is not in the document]
umatin: Yeah

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,256 @@
...
+    for (int i = 0; i < numCaches; i++) {
+      Value cacheArg = body->getArgument(outputs.size() + i);
+
+      Type ct = cacheArg.getType();
+      Type type = MemRefType::get(shapes, ct);
+      auto alloc = cacheBuilder.create<memref::AllocOp>(
+          op->getLoc(), type, ValueRange(iterationDomains));
+      Value cache = gutils->initAndPushCache(alloc, cacheBuilder);
+      alloc->setAttr(alloc.getOperandSegmentSizesAttrName(),
+                     cacheBuilder.getDenseI32ArrayAttr({1, 0}));
```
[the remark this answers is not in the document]
umatin: I agree there was a bug here. Which highlevel API are you referring to? AllocOP does not seem to be exposing anything for me.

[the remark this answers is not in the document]
ftynse: Absolutely all operations produced from ODS define getters and mutators for all named arguments. For operands, the getter is `getArgName` (with automatic capitalization from `arg_name`), and the mutator is `getArgNameMutable`. The latter returns a mutable object that can be used to change the operand list. For attributes, the getter is `getAttrName` and the setter is `setAttrName` for `attr_name`. These (de)construct the attribute on-the-fly to get you the underlying value, e.g. an integer for integer attributes. There are also `getAttrNameAttr` and `setAttrNameAttr` that don't do the deconstruction.

In this case, it should be possible to do `alloc.getDynamicSizesMutable().erase(1, alloc.getDynamicSizesMutable().size() - 1)`. 

On the general basis, the code operating on specific ops should (almost) never use `Operation*` API. That is, given an object of concrete `FooOp` type, the code must not do `fooOp->`, only `fooOp.`. This will result in subtle problems when somebody inevitably changes the ODS definition. E.g., they swap the order of arguments in it to follow the alphabetical order, and suddenly `->getOperand(0)` is no longer what it used to be (but not null!) and there is no way to know.

And the code must under no circumstances modify the "internal" attributes such as operand segment sizes.

That being said, it's unclear to me why the code here can't just directly construct the correct op.

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -132,4 +134,15 @@ def ShadowedGradient : Enzyme_Type<"ShadowedGradient"> {
   let assemblyFormat = "`<` $basetype `>`";
 }
 
+def AddToOp : Enzyme_Op<"addTo">, //, [Pure, Terminator, ReturnLike]
```
umatin, line 137: @wsmoses Is it possible to have this inherit from linalg.yield to pass the verifier of linalg.generic? Also any idea for a better name?

wsmoses: Not that I'm aware of unfortunately.

wsmoses: @ftynse may have more info tho

umatin: What should I use if thats not possible? Just make a custom linalg generic op?

enzyme/Enzyme/MLIR/Implementations/LinalgAutoDiffOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,294 @@
...
+      if (!gutils->hasInvertPointer(input->get())) {
+        continue;
+      }
+      indexingMaps.push_back(linalgOp.getMatchingIndexingMap(input));
+      Value inp = gutils->invertPointerM(input->get(), builder);
+      Value view = invertMemref(inp, builder, op->getLoc());
+      inputs.push_back(view);
+    }
+
+    // linalg::GenericOp adjoint = builder.create<linalg::GenericOp>(
```
wsmoses, line 158: cleanup here?

PR #1140, opened 2023-04-28
More opaque pointer improvements

wsmoses:


jeremylt:
I can confirm that this fixed the issue I had building Enzyme + libCEED + PETSc + Ratel

PR #1154, opened 2023-05-03
Trace unification

tgymnich:


enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -511,6 +511,15 @@ bool ActivityAnalyzer::isFunctionArgumentConstant(CallInst *CI, Value *val) {
       CI->getArgOperand(0) == val)
     return true;
 
+  if ((F->hasFnAttribute("enzyme_trace") || CI->hasFnAttr("enzyme_trace")) &&
```
wsmoses, line 514: There are two function-related activity ones. Add to the second as well? Or was it only missing this

tgymnich: I am not sure what you mean. I see 3 prob prog related functions here, all of them handle both fn attrs and call attrs.

PR #1192, opened 2023-05-12
fix floatRet TT

ZuseZ4:


enzyme/tools/enzyme-tblgen/blasTAUpdater.h
```diff
@@ -75,7 +76,7 @@ void emit_BLASTA(TGPattern &pattern, raw_ostream &os) {
   }
   if (name == "dot" || name == "asum" || name == "nrm2") {
```
wsmoses, line 77: Do these need to be hardcoded or could the get derived from tablegen (by saying what the returntype of the fn is in addition to args)?

ZuseZ4: I need to put it at some place. Either here, or by updating the blas td. 
Since the vast majority of BLAS function doesn't return anything and the blas declarations are already getting larger (think of mutable args, order of rule execution, ...) I thought it's cleaner to just add it here.

wsmoses: yeah that's fine

wsmoses (approved):

PR #1200, opened 2023-05-15
Update AdjointGenerator.h with maximum and minimum intrinsics

motabbara:


wsmoses:
Can you add a forward and reverse mode ll test? (look inside repo/enzyme/test/ReverseMode/ and repo/enzyme/test/ForwardMode).

motabbara:
Added a couple of tests @wsmoses

motabbara:
How's that look @wsmoses ?

motabbara:
Along similar lines, is maxnum.ll correct in forward mode?

wsmoses:
@motabbara are you able to see the results of CI or run the tests locally?

I'm fine to merge as soon as the pass (and as a consequence test for the correct result :P )

enzyme/test/Enzyme/ForwardMode/maximum.ll
```diff
@@ -0,0 +1,29 @@
+; RUN: %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -S | FileCheck %s
```
wsmoses, line 1: The old pass manager line needs to be version gated to be < 16.

Separately if this intrinsic is only LLVM15+, you also need to version gate all tests to be 15+.

enzyme/test/Enzyme/ForwardMode/maximum.ll
```diff
@@ -0,0 +1,29 @@
...
+declare float @llvm.maximum(float, float)
+
+; Function Attrs: nounwind
+declare float @__enzyme_fwddiff(float (float, float)*, ...)
+
+; CHECK: define internal {{(dso_local )?}}float @fwddiffetester(float %x, float %"x'", float %y, float %"y'")
+; CHECK-NEXT: entry:
+; CHECK-NEXT:   %0 = fcmp fast olt float %x, %y
+; CHECK-NEXT:   %1 = select {{(fast )?}}i1 %0, float %"x'", float %"y'"
+; CHECK-NEXT:   ret float %1
```
wsmoses, line 27: This looks to be computing the derivative for min?

Since it returns (x < y) ? x' : y'

motabbara: I found that confusing too. This test is identical to that for maxnum that is passes when I run locally. Is it that the forward mode implementation for that needs a fix?

motabbara: Changed the implementation but have only updated a few of the tests.

tgymnich (approved):
Looks good. Thank you!

motabbara:
@tgymnich don't think this will get auto-merged while CI pipeline is still in flux. Can you please manually merge?

PR #1204, opened 2023-05-17
handle negative strides in memcpy

ZuseZ4:
fixes https://github.com/EnzymeAD/Enzyme.jl/pull/842


wsmoses (approved):
Add test?

ZuseZ4:
We will get the julia dot tests and I will update the llvm-ir ddot tests, that should be enough?

PR #1216, opened 2023-05-21
Add sincos for float

wsmoses:


wsmoses:
@heltonmc check that this is to your liking, we define d/dx of sinpi(x) as pi * sinpi(x+0.5)

heltonmc:
> pi * sinpi(x+0.5)

My initial concern (without any testing) is that for large `x` you'd lose some of the phase information but kinda looks like `sinpi` starts to become unreliable for large values. @oscardssmith would be better to say here though. Any thoughts on this? 

```julia
julia> x = rand()*1e18
2.105206367465433e17

julia> sinpi(x)
0.0

julia> sinpi(x)
0.0

julia> sincospi(x)
(0.0, 1.0)

julia> cospi(x)
1.0

julia> cos(pi*x)
0.18385213876343298

julia> sin(pi*x)
-0.9829539109602806
```

oscardssmith:
this should probably use sincospi for the derivative.

wsmoses:
An implementation of sincospi is not guaranteed to be available at this level. This applies for any sinpi-like function, e.g. one from a fortran API, etc. 

As for large values, this should be equivalent to sincospi. For one thing Julia ensures it is in range:
https://github.com/JuliaLang/julia/blob/a6ad9ea099fb9557e53186e5974a0d68395602a2/base/special/trig.jl#L804

Secondly, if you look at sinpi, cospi, sincospi, they all compute the equivalent values here:

```
function sinpi(_x::T) where T<:Union{IEEEFloat, Rational}
    x = abs(_x)
    if !isfinite(x)
        isnan(x) && return x
        throw(DomainError(x, "`x` cannot be infinite."))
    end
    # For large x, answers are all 1 or zero.
    if T <: AbstractFloat
        x >= maxintfloat(T) && return copysign(zero(T), _x)
    end

    # reduce to interval [0, 0.5]
    n = round(2*x)
    rx = float(muladd(T(-.5), n, x))
    n = Int64(n) & 3
    if n==0
        res = sinpi_kernel(rx)
    elseif n==1
        res = cospi_kernel(rx)
    elseif n==2
        res = zero(T)-sinpi_kernel(rx)
    else
        res = zero(T)-cospi_kernel(rx)
    end
    return ifelse(signbit(_x), -res, res)
end
"""
    cospi(x)

Compute ``\\cos(\\pi x)`` more accurately than `cos(pi*x)`, especially for large `x`.
"""
function cospi(x::T) where T<:Union{IEEEFloat, Rational}
    x = abs(x)
    if !isfinite(x)
        isnan(x) && return x
        throw(DomainError(x, "`x` cannot be infinite."))
    end
    # For large x, answers are all 1 or zero.
    if T <: AbstractFloat
        x >= maxintfloat(T) && return one(T)
    end

    # reduce to interval [0, 0.5]
    n = round(2*x)
    rx = float(muladd(T(-.5), n, x))
    n = Int64(n) & 3
    if n==0
        return cospi_kernel(rx)
    elseif n==1
        return zero(T)-sinpi_kernel(rx)
    elseif n==2
        return zero(T)-cospi_kernel(rx)
    else
        return sinpi_kernel(rx)
    end
end
"""
    sincospi(x)

Simultaneously compute [`sinpi(x)`](@ref) and [`cospi(x)`](@ref) (the sine and cosine of `π*x`,
where `x` is in radians), returning a tuple `(sine, cosine)`.

!!! compat "Julia 1.6"
    This function requires Julia 1.6 or later.

See also: [`cispi`](@ref), [`sincosd`](@ref), [`sinpi`](@ref).
"""
function sincospi(_x::T) where T<:Union{IEEEFloat, Rational}
    x = abs(_x)
    if !isfinite(x)
        isnan(x) && return x, x
        throw(DomainError(x, "`x` cannot be infinite."))
    end
    # For large x, answers are all 1 or zero.
    if T <: AbstractFloat
        x >= maxintfloat(T) && return (copysign(zero(T), _x), one(T))
    end

    # reduce to interval [0, 0.5]
    n = round(2*x)
    rx = float(muladd(T(-.5), n, x))
    n = Int64(n) & 3
    si, co = sinpi_kernel(rx),cospi_kernel(rx)
    if n==0
        si, co = si, co
    elseif n==1
        si, co  = co, zero(T)-si
    elseif n==2
        si, co  = zero(T)-si, zero(T)-co
    else
        si, co  = zero(T)-co, si
    end
    si = ifelse(signbit(_x), -si, si)
    return si, co
end
```

PR #1221, opened 2023-05-23
Add enzyme/tools to clang-format check

### LINKED ISSUE #1211: Add .clang-format and corresponding CI pass to check code format
Like [llvm-project](https://github.com/llvm/llvm-project), I think we can add a .clang-format file in the root dir (and the file content is pretty simple since we just use the LLVM-style code format `BasedOnStyle: LLVM`), and then add a github action to check every PR commit to ensure the code style is aligned before it is merged.

PragmaTwice:
Some misc work are done in this PR:
- move `.clang-format` to the top directory, as a common way
- add `pull_request` event for the clang-format check CI
- add enzyme/tools dir for clang-format check
- format enzyme/tools dir via clang-format
- remove some useless content (e.g. CMakeLists.txt in the exclude list, which is nonsense)

Close #1211. cc @wsmoses @ZuseZ4

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,11 +1,13 @@
 name: Benchmarking
 
 on: 
-  workflow_run:
-    workflows: [Format]
-    types: 
-      - completed
-
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
+    
 concurrency: 
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,10 +1,12 @@
 name: C/C++ CI
 
 on: 
-  workflow_run:
-    workflows: [Format]
-    types: 
-      - completed
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
 
 jobs:
   build:
--- a/.github/workflows/enzyme-julia.yml
+++ b/.github/workflows/enzyme-julia.yml
@@ -1,10 +1,12 @@
 name: Julia Enzyme CI
 
 on:
-  workflow_run:
-    workflows: [Format]
-    types: 
-      - completed
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
 
 jobs:
   build:
--- a/.github/workflows/enzyme-mlir.yml
+++ b/.github/workflows/enzyme-mlir.yml
@@ -1,10 +1,12 @@
 name: MLIR
 
 on:
-  workflow_run:
-    workflows: [Format]
-    types: 
-      - completed
+  push:
+    branches:
+    - main
+  pull_request:
+    branches:
+    - main
 
 jobs:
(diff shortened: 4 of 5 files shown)


ZuseZ4:
@tgymnich @wsmoses iirc we had a reason to exclude PRs?
Also, how about running the rest of CI only once clang-fmt passes, if it's not too much work? Formating needs to get fixed before merging anyway and might save us CI time or the work to manually cancel old runs which already failed due to clang-format.

PragmaTwice:
> Also, how about running the rest of CI only once clang-fmt passes, if it's not too much work? Formating needs to get fixed before merging anyway and might save us CI time or the work to manually cancel old runs which already failed due to clang-format.

Refer to https://stackoverflow.com/questions/58457140/dependencies-between-workflows-on-github-actions , it seems we can add `workflow_run` clause to other workflows.

ZuseZ4:
@wsmoses any objections to it? 
It's nothing big, but it is easy to fix formating when it breaks CI and it is annoying to go trough all of our CI categories to cancel runs in which the formating has failed. Having clang-format as first check would prevent that.

wsmoses:
I'm fine with the don't run the rest of CI, but let's enable it for everything except the LLVM CI (that way it can still check that it builds on all versions), which takes the least time.

ZuseZ4:
Sounds good
@PragmaTwice, would you be fine with implementing these changes (removing runs on PR and adding the two level CI)?

PragmaTwice:
> yeah if its already run on every commit push, also running for PR (which would require a commit) is redundant.

Got your point. But I think it is not friendly to outside contributors who fork this repo and open pull requests from their own fork, in which case these CI status does not appear in the PR page.

ZuseZ4:
That sounds good, feel free to implement it this way :)

ZuseZ4 (approved):

PragmaTwice:
Conflict solved now.

PragmaTwice:
It seems this method cannot work well, i.e. these workflows by `workflow_run` event cannot be triggered as expected.

Maybe need some investigation or workaround, or just keep the original behavior and try it in another PR.

PragmaTwice:
https://stackoverflow.com/questions/72551643/workflow-run-not-triggered-as-expected-after-parent-workflow-completes

Maybe the root cause is from the quote. Some trying...

ZuseZ4:
It does look like Integration CI is now blocked on all Enzyme CI tests passing. Especially the apple runners take a while, so by now it would increase our CI times by a good amount. Can you update it please to make sure that we are only waiting for the clang-format check to pass before starting Integration CI?

PragmaTwice:
> Edit: looks like we noticed it simultaneously.

:rofl:

> https://stackoverflow.com/questions/72551643/workflow-run-not-triggered-as-expected-after-parent-workflow-completes
Maybe the root cause is from the quote. Some trying...

Some try, still not triggered. I will revert all these `workflow_run` changes firstly.

PragmaTwice:
Hi @ZuseZ4 @wsmoses , now it is reverted from `workflow_run` clauses to build dependencies between workflows.

I think we can firstly merge current changes to check tblgen code format and prevent further conflicts.

PR #1239, opened 2023-05-28
Don't set all frees

wsmoses:


vchuravy (approved):

vchuravy:
The langref states that this is a "may tail-call" but I agree that the weirdness may be related

PR #1256, opened 2023-06-04
Use zeroext on duplicated

wsmoses:


vchuravy (approved):
Should only matter in platforms like PPC where smallest register is 32bit

PR #1272, opened 2023-06-10
Fix string usage

tgymnich:
`std::map<std::string, T> `--> `StringMap<T>`
`std::set<std::string>` --> `StringSet<>`

Additionally replace `std::string` in function signatures with `StringRef` and avoid materialising `std::strings` where possible

### DIFF
--- a/enzyme/Enzyme/EnzymeLogic.cpp
+++ b/enzyme/Enzyme/EnzymeLogic.cpp
@@ -2040,7 +2040,7 @@ const AugmentedReturn &EnzymeLogic::CreateAugmentedPrimal(
       SmallVector<Value *, 3> fwdargs;
       int act_idx = 0;
       while (arg != NewF->arg_end()) {
-        arg->setName("arg" + std::to_string(act_idx));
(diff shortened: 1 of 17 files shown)


enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -2894,7 +2896,6 @@ const AugmentedReturn &EnzymeLogic::CreateAugmentedPrimal(
   for (auto user : fnusers) {
     if (removeStruct) {
       IRBuilder<> B(user);
-      auto n = user->getName().str();
       user->setName("");
       SmallVector<Value *, 4> args(user->arg_begin(), user->arg_end());
       auto rep = B.CreateCall(NewF, args);
```
tgymnich, line 2901: @wsmoses this call here should get the name of `user` right?

wsmoses (approved):

enzyme/Enzyme/EnzymeLogic.cpp
wsmoses: Yeah that would be nice

We can use takeName to do

PR #1273, opened 2023-06-10
Replace std::function with llvm::function_ref

tgymnich:


wsmoses:
The non owning part makes be a bit worried at least for things that capture via the C API for conversion.

Also is there guidance anywhere that llvm vs stl on this is preferred, given the recent switch to std optional?

tgymnich:
I made sure to only use `llvm::function_ref` on non-captured closures

wsmoses:
Nice!

I guess I'm still a bit skeptical about merging since the function ref makes a potential larger bug surface if accidentally used in a capturing way, and I'm also worried about it getting deprecated.

What prompted the idea to change out of curiosity?

wsmoses (approved):

PR #1277, opened 2023-06-13
err on under-allocations and not on over-allocations

tthsqe12:


wsmoses (commented):
Mind adding a test that triggers to prevent regression?

tthsqe12:
Sure. Unless you object, I will put it in the integration folder.

tthsqe12:
I do apologize as it is taking longer than expected to get everything setup. I am almost there.

tthsqe12:
@wsmoses should be done now.

wsmoses:
Mind also adding a test in the case it should fail as well?

Something like: https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/test/Integration/ReverseMode/err_too_few_args.c

wsmoses (approved):

enzyme/test/Integration/ReverseMode/allocatedtape_err.c
```diff
@@ -0,0 +1,21 @@
...
+// RUN: if [ %llvmver -ge 12 ]; then %clang -std=c11 -g -O2 %s -S -emit-llvm -o -  %newLoadClangEnzyme -Xclang -verify; fi
+// RUN: if [ %llvmver -ge 12 ]; then %clang -std=c11 -g -O3 %s -S -emit-llvm -o -  %newLoadClangEnzyme -Xclang -verify; fi
+
+extern int enzyme_allocated, enzyme_tape;
+void __enzyme_reverse(void*, ...);
+
+void f(double* x) {
+    x[0] += x[1]*x[1];
+    x[1] = -x[1];
+}
```
wsmoses, line 16: Looks like for LLVM 10+ the caching infra does a better job and instead emits this error. Mind doing a simpler test case (perhaps x[0] = x[1]*x[1]), and opening an issue to inspect the LLVM 9 cache size.

```
 error: 'error' diagnostics expected but not seen: 
  File /home/runner/work/Enzyme/Enzyme/enzyme/test/Integration/ReverseMode/allocatedtape_err.c Line 19: Enzyme: need 24 bytes have 1 bytes
error: 'error' diagnostics seen but not expected: 
  File /home/runner/work/Enzyme/Enzyme/enzyme/test/Integration/ReverseMode/allocatedtape_err.c Line 19: Enzyme: need 8 bytes have 1 bytes
  ```

tthsqe12: Hmm - Are you saying that with llvm version  9, you get a 24 byte tape, and with llvm version 10, you get an 8 byte tape? I ran this with llvm 15.0.7 and got a 24 byte tape. Nevertheless, I will try to find a simpler example where the tape size is incontrovertibly 8.

wsmoses: In part, I also imagine it depends on optimization level, I think.

PR #1280, opened 2023-06-14
Fix bazel build in ci

wsmoses:


.github/workflows/enzyme-bazel.yml
```diff
@@ -28,14 +28,7 @@ jobs:
...
           sudo apt-get update
-          sudo apt-get install -y ninja-build git autoconf cmake gcc g++ libtool gfortran python3 python3-dev 
-    - uses: actions/checkout@v3
-    - uses: actions/checkout@v3
-      with:
-        repository: 'llvm/llvm-project'
-        path: 'llvm'
-        submodules: true
-
+          sudo apt-get install -y ninja-build git autoconf cmake gcc g++ libtool gfortran python3 python3-dev
```
tgymnich, line 31: do we need autoconf, libtool and gfortran?

PR #1281, opened 2023-06-15
use tg declarations to handle mat and vec dimensions

ZuseZ4:


ZuseZ4:
@wsmoses are we sure that this is correct?
https://github.com/EnzymeAD/Enzyme/blob/c5a6954570b6929ac10cdd7f6c8b3c505ecab82c/enzyme/Enzyme/Utils.cpp#L2103
I just added handling for transposed matrices and my tests indicate that it's not.
I did modify the gemm_f_c_transpose_lacpy.ll testcase by setting transa to 'T' instead of 'N' (and A is modified, thus cached).

wsmoses:
That code is wrong there should never be a builder nor any load or store
generated

wsmoses:
The condition at the end should be return false at the end of the function
not return true

wsmoses:
There should also be a continue after the for loop adding all the users
into todo

ZuseZ4:
@wsmoses thanks. Behaves better now.

ZuseZ4:
This now also has enough to remove this one: `(ld $A, $transa, $lda, $m, $n),`,
but since we already solved that I'll focus on nested rules before cleaning working code up.

wsmoses (approved):

tgymnich:
this broke both macOS build and macOS tests

ZuseZ4:
weren't they broken already for a while @tgymnich ?

tgymnich:
Yeah i fixed them last WE. The Xcode build is the only one still broken.

PR #1286, opened 2023-06-17
Tablegen bessel fn

wsmoses:


enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -113,8 +123,9 @@ class GlobalExpr<string val> {
 class ConstantFP<string val> {
   string value = val;
 }
-class ConstantInt<int val> {
+class ConstantInt<int val, int size=0> {
   int value = val;
+  int bitwidth=size;
```
tgymnich, line 128: use similar names

tgymnich (approved):

enzyme/Enzyme/InstructionDerivatives.td
wsmoses: will resolve in follow up

PR #1289, opened 2023-06-19
tblgen cleanup

tgymnich:


enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
```diff
@@ -1809,14 +1831,14 @@ size_t fwd_call_args(TGPattern &pattern, size_t actArg,
       result.append((Twine("arg_") + name).str());
     } else if (ty == argType::cblas_layout) {
       // layout is only allowed as first type (which we skipped)
-      llvm::errs() << "name: " << name << " typename: " << ty
```
wsmoses, line 1812: Perhaps this should get turned into the tablegen error?

tgymnich: what is a tablegen error?

ZuseZ4: we can't use tblgen assertions due to supporting oder llvm versions, or did you find an error around that?

ZuseZ4: @tgymnich this: https://llvm.org/docs/TableGen/ProgRef.html#assert-check-that-a-condition-is-true
We could split up the current blas into blas-lv1-returns-float, blas-lv1-others and blas-lv-23, where number two and three have one single arg that they modify, while the first one only returns a float and is readonly.
The third case would then optionally have an extra layout arg (implicitely). I think it might not safe much to model this on the tablegen level.

tgymnich: ok lets not do this in a code cleanup PR then

ZuseZ4 (approved):
thanks

PR #1291, opened 2023-06-20
Activate clang-format for tblgen

tgymnich:


.github/workflows/format.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: DoozyX/clang-format-lint-action@v0.16.2
       with:
+        extensions: 'h,cpp,c,td'
```
wsmoses, line 14: I'm not sure we want to do this for the .td files. Thoughts?

tgymnich: Thats the whole point for the PR.

tgymnich: I guess thats about taste. And having one fixed representation of the code.

wsmoses: Yeah I can see pros and cons, and just haven't thought about it deeply enough that I'm curious for your thoughts.

cc @ftynse @umatin @pengmai re MLIR since that will be substantial amts of tablegen.

tgymnich: I must admit though that some of the formatted code blocks look quite shabby.

PR #1292, opened 2023-06-20
Extend mpi support

l90lpa:


enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -136,7 +136,7 @@ const StringSet<> InactiveGlobals = {
     "_ZTVN10__cxxabiv121__vmi_class_type_infoE"
 };
 
-const llvm::StringMap<size_t> MPIInactiveCommAllocators = {
+const llvm::StringMap<size_t> MPIInactiveAllocators = {
```
l90lpa, line 139: I've changed the name because I've added some non-comm allocator routines.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -147,7 +147,8 @@ class AdjointGenerator
   }
 
   llvm::Value *MPI_TYPE_SIZE(llvm::Value *DT, llvm::IRBuilder<> &B,
-                             llvm::Type *intType) {
+                             llvm::Type *intType,
+                             llvm::BasicBlock *allocBlock = nullptr) {
```
l90lpa, line 151: When handling MPI_Allreduce using min/max op, I replace the call in the primal code with an MPI_Allreduce using the minloc/maxloc (+ some loops for copying data, and mallocs for temporary arrays). I'm not sure if it is the best strategy, but construct this replacement code inside a function, similar to `getOInsertCheckedFree`. Because I'm constructing the replacement code in a function I added the extra parameter `allocaBlock` to MPI_TYPE_SIZE etc so that I could use them in this case.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -4790,9 +4799,103 @@ class AdjointGenerator
     assert(called);
     assert(gutils->getWidth() == 1);
 
-    IRBuilder<> BuilderZ(gutils->getNewFromOriginal(&call));
+    auto newCall = gutils->getNewFromOriginal(&call);
+    IRBuilder<> BuilderZ(newCall);
     BuilderZ.setFastMathFlags(getFast());
 
+    enum class MPI_Op_Kind { SUM, MIN, MAX, MINLOC, MAXLOC, UNKNOWN };
```
l90lpa, line 4806: I've added a few helper functions and maps at the start of handleMPI that I extracted from MPI_Allreduce handling, and extended.

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -6252,83 +6335,650 @@ class AdjointGenerator
...
+          return Ty;
+        };
 
-#if LLVM_VERSION_MAJOR >= 11
-          auto callval = call.getCalledOperand();
+        // If the primal op is min/max then replace it with the minloc/maxloc
+        // op in the primal code
+        CallInst *newNewCall;
+        Value *recvlocbuf;
+        if (opKind == MPI_Op_Kind::MIN || opKind == MPI_Op_Kind::MAX) {
```
l90lpa, line 6372: This is the block where I construct the replacement code for MPI_Allreduce when the op is min/max, and then replace the primal call.  The replacement code is constructed in a function (I took a similar approach as getOrInsertCheckedFree. Is this the best approach? Or is it better to generate the code in-place?

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -6252,83 +6335,650 @@ class AdjointGenerator
...
+                           recvlocbuf,
+                           count,
+                           compLocOp,
+                           comm};
+
+          newNewCall =
+              BuilderZ.CreateCall(mpi_allreduce_comploc_fp->getFunctionType(),
+                                  mpi_allreduce_comploc_fp, args);
+
+          gutils->replaceAWithB(newCall, newNewCall);
```
l90lpa, line 6636: Is this the best way to replace the primal code?

PR #1309, opened 2023-07-01
Support nested BLAS rules

ZuseZ4:


ZuseZ4:
We now avoid the double free of matrices, 
have less cases where we emit strange blas declarations (in case that we pass constant floats),
and can allocate tmp vectors or matrices for nested rules.
I'll update the two scalar gemv/gemm tests tmrw

wsmoses (approved):

wsmoses:
@ZuseZ4 this broke macos CI if you can fix those tests?

ZuseZ4:
@wsmoses Sure. I tried to find yesterday where those mac runners got stuck, but I couldn't find them in the github action window, so I had to merge without waiting for them.

PR #1323, opened 2023-07-07
Fix build for LLVM 17

pengmai:
This PR fixes a failing build that I tested with LLVM commit `73a0ae021ec6112911b98025055b8a2e881b1376`. It matches [a corresponding `LLVM_VERSION_MAJOR` check in the top of that same function](https://github.com/EnzymeAD/Enzyme/blob/28e04cb96f7d9ba3f17ca77abde3cedf4e045b5f/enzyme/Enzyme/Enzyme.cpp#L635C1-L647C7).


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -1120,11 +1120,19 @@ class EnzymeBase {
       return {};
     }
 
+#if LLVM_VERSION_MAJOR > 16
```
wsmoses, line 1123: Alternatively, we can just do return Options({....});

pengmai: Oh true, I like that better. Do you happen to know if the automatic conversion happens for both `std::optional` and `llvm::Optional`? Or do we still need the check?

wsmoses: it should happen for both, we use this convention elsewhere

wsmoses (approved):

PR #1324, opened 2023-07-08
add blas runtime activity

ZuseZ4:


wsmoses (approved):
Add a test?

ZuseZ4:
test does not show the expected changes yet, looking into why.

enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
```diff
@@ -1618,12 +1618,31 @@ void emit_helper(const TGPattern &pattern, raw_ostream &os) {
...
+  for (size_t i = 0; i < actArgs.size(); i++) {
+    auto name = nameVec[actArgs[i]];
+
+    // floats are passed by calue, except of the Fortran Abi (byRef)
+    auto ty = argTypeMap.lookup(actArgs[i]);
+      os << "  if (";
+      if (ty == ArgType::fp)
+        os << "byRef && ";
+      os << "active_" << name << " && arg_" << name
+         << " == gutils->invertPointerM(orig_" << name << ", BuilderZ))\n"
```
wsmoses, line 1641: This is a compile time check it needs to be a runtime check aka builder.createICMpEq(primal, shadow)

wsmoses (commented):
Add a test?

enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
ZuseZ4: That's what I was just figuring out, updating it.

ZuseZ4:
@wsmoses In the case of !RuntimeActivity, nothing changes.
In the case of RuntimeActivity, I do have currently 2 BB for each variable that is marked as active. 
The first BB is always entered and based on a previous ptr-comparison it will then either enter the corresponding
implementation BB, or it will move to the first BB of the next potentially active variable, where it again will branch
based on the ptr comparison result.

We do have all the ptr comparison results available beforehand, so I guess we could try to do more math to just have one BB for each active variable and just jump into the runtime-active bb, if you think that this is worth it?

tgymnich (approved):

ZuseZ4:
auto-merging, because with the latest diffUse update it now passes all tests.
If we need larger changes for blas diffUse, let's make them sepeartely

enzyme/test/Enzyme/ForwardMode/blas/cblas_ddot.ll
```diff
@@ -1,5 +1,5 @@
-; RUN: if [ %llvmver -lt 16 ]; then %opt < %s %loadEnzyme -enzyme -S | FileCheck %s; fi
-; RUN: %opt < %s %newLoadEnzyme -passes="enzyme" -enzyme-preopt=false -S | FileCheck %s
+; RUN: if [ %llvmver -lt 16 ]; then %opt < %s %loadEnzyme -enzyme -enzyme-runtime-activity=1 -S | FileCheck %s; fi
+; RUN: %opt < %s %newLoadEnzyme -passes="enzyme" -enzyme-preopt=false -enzyme-runtime-activity=1 -S | FileCheck %s
```
wsmoses, line 2: @ZuseZ4 we should error if forward mode runtime activity is requetsed (rather than silently doing the wrong thing, like here).

PR #1332, opened 2023-07-15
simplify and fix blasDiffUse

ZuseZ4:
Apparently you can introduce subtle bugs if you re-do work unnecessarily.
@wsmoses this new behavior should be always correct, right?

```
   27   if (val == arg_lda && need_lda && !cache_lda)
   26      return true;
```

At least it will fix the lda issue for runtime-activity. Adding some more tests there rn.

### DIFF
--- a/enzyme/tools/enzyme-tblgen/blasDiffUseUpdater.h
+++ b/enzyme/tools/enzyme-tblgen/blasDiffUseUpdater.h
@@ -49,20 +49,10 @@ void emit_BLASDiffUse(TGPattern &pattern, llvm::raw_ostream &os) {
   }
 
   emit_scalar_caching(pattern, os);
-  // we currently cache all vecs before we cache all matrices
-  // once fixed we can merge this calls
   for (size_t i = 0; i < nameVec.size(); i++) {
     auto ty = typeMap.lookup(i);
-    if (ty != ArgType::vincData)
+    if (ty != ArgType::vincData && ty != ArgType::mldData)
       continue;
-    assert(typeMap.lookup(i + 1) == ArgType::vincInc);
-    emit_mat_vec_caching(pattern, i, os);
-  }
-  for (size_t i = 0; i < nameVec.size(); i++) {
-    auto ty = typeMap.lookup(i);
-    if (ty != ArgType::mldData)
-      continue;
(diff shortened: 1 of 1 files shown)


wsmoses:
I think the cache part shouldn't be here since diffuse analysis is used by the algorithm to determine caching.

ZuseZ4:
cacheMode  just checks if it's forwardModeonly.
cache_<X> is also manually estimated, based on uses only:
```
    2   bool cache_alpha = cacheMode && byRef && overwritten_alpha && need_alpha;
    3   bool need_lda = active_A || active_B || active_alpha
    4   bool cache_lda = cacheMode && byRef && overwritten_lda && need_lda;
```
So it's only the name, or am I using any information that I should not?

tgymnich (approved):

enzyme/tools/enzyme-tblgen/blasDiffUseUpdater.h
```diff
@@ -49,37 +49,20 @@ void emit_BLASDiffUse(TGPattern &pattern, llvm::raw_ostream &os) {
...
-    for (auto a : users) {
-      auto name = nameVec[a];
-      // The following shows that I probably should change the tblgen
-      // logic and the Blas.td declaration
-      if (a == i) // a == i? argpos ?
-        continue;
-      os << "    if (active_" << name << ") return true;\n";
-    }
-    os << "  }\n";
+    os << "  if (val == arg_" << name << " && need_" << name << " && !cache_"
```
wsmoses, line 63: Like mentioned on slack, I don't think it's right to do cache checks here since this analysis is used by cache analysis.

Can you revert the cache change?

wsmoses: Ah sorry I meant on here, not slack.

But yeah I believe there's an effective circular dependency, which here will be manifested by inconsistent results (not a compile time error), which is potentially very bad.

ZuseZ4: just to clarify @wsmoses. If it existend then it existed just on the same level before. I did use exactly the same logic before this change, I just re-did some of the calculation. If you want this to be fundamentally change reverting will not change it, we would rather need to see what to use else

PR #1335, opened 2023-07-17
assert no RuntimeActivity is used for fwd-mode BLAS AD

ZuseZ4:


enzyme/tools/enzyme-tblgen/blas-tblgen.cpp
```diff
@@ -670,6 +670,8 @@ void emit_fwd_rewrite_rules(const TGPattern &pattern, raw_ostream &os) {
      << "    auto callval = call.getCalledValue();           \n"
      << "#endif                                            \n\n";
 
+  os << "    assert(!EnzymeRuntimeActivityCheck && \"RuntimeActivity for fwd-mode is not implemented yet!\");\n";
```
wsmoses, line 673: Can this be changed to CustomErrorHandler and emitfailure like below?

https://github.com/EnzymeAD/Enzyme/blob/3c71081c998a0aef7454e4e3d31b1dfd58549bd6/enzyme/Enzyme/AdjointGenerator.h#L346

This way the compiler won't die, but instead say that it failed to compile

enzyme/test/Enzyme/ForwardMode/blas/cblas_ddot.ll
```diff
@@ -44,6 +44,7 @@ entry:
   ret double 1.0
 }
 
+; XFAIL: *
```
wsmoses, line 47: You should make one test with runtime activity, and one test without (rather than this single test).

wsmoses (approved):

PR #1338, opened 2023-07-18
support more names from MS

tthsqe12:
~~How do you feel about in Enzyme.cpp replacing all of the
`metaString->startswith("enzyme_")`
by
`metaString->contains("enzyme_")`
?
These can be mangled too.~~

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -4468,10 +4468,13 @@ Constant *GradientUtils::GetOrCreateShadowConstant(
         NewOps.push_back(i == 0 ? C : arg->getOperand(i));
       return arg->getWithOperands(NewOps);
     }
+  } else if (auto arg = dyn_cast<GlobalAlias>(oval)) {
+    return GetOrCreateShadowConstant(Logic, TLI, TA, arg->getAliasee(), mode, width, AtomicAdd);
   } else if (auto arg = dyn_cast<GlobalVariable>(oval)) {
     if (arg->getName() == "_ZTVN10__cxxabiv120__si_class_type_infoE" ||
         arg->getName() == "_ZTVN10__cxxabiv117__class_type_infoE" ||
-        arg->getName() == "_ZTVN10__cxxabiv121__vmi_class_type_infoE")
+        arg->getName() == "_ZTVN10__cxxabiv121__vmi_class_type_infoE" ||
+        arg->getName().startswith("??_R")) // any of the MS RTTI manglings
       return arg;
 
     if (hasMetadata(arg, "enzyme_shadow")) {
@@ -9182,15 +9185,21 @@ llvm::CallInst *freeKnownAllocation(llvm::IRBuilder<> &builder,
     freefunc = LibFunc_ZdaPv;
     break;
 
-  case LibFunc_msvc_new_int:               // new(unsigned int);
-  case LibFunc_msvc_new_int_nothrow:       // new(unsigned int, nothrow);
   case LibFunc_msvc_new_longlong:          // new(unsigned long long);
   case LibFunc_msvc_new_longlong_nothrow:  // new(unsigned long long, nothrow);
-  case LibFunc_msvc_new_array_int:         // new[](unsigned int);
-  case LibFunc_msvc_new_array_int_nothrow: // new[](unsigned int, nothrow);
+    freefunc = LibFunc_msvc_delete_ptr64_longlong;
+    break;
+
   case LibFunc_msvc_new_array_longlong:    // new[](unsigned long long);
   case LibFunc_msvc_new_array_longlong_nothrow: // new[](unsigned long long,
                                                 // nothrow);
+    freefunc = LibFunc_msvc_delete_array_ptr64_longlong;
+    break;
+
+  case LibFunc_msvc_new_int:               // new(unsigned int);
+  case LibFunc_msvc_new_int_nothrow:       // new(unsigned int, nothrow);
+  case LibFunc_msvc_new_array_int:         // new[](unsigned int);
+  case LibFunc_msvc_new_array_int_nothrow: // new[](unsigned int, nothrow);
     llvm_unreachable("msvc deletion not handled");
 
   default:
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -4469,7 +4469,8 @@ Constant *GradientUtils::GetOrCreateShadowConstant(
       return arg->getWithOperands(NewOps);
     }
   } else if (auto arg = dyn_cast<GlobalAlias>(oval)) {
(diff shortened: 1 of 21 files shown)


enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -9182,15 +9185,21 @@ llvm::CallInst *freeKnownAllocation(llvm::IRBuilder<> &builder,
...
-  case LibFunc_msvc_new_array_int:         // new[](unsigned int);
-  case LibFunc_msvc_new_array_int_nothrow: // new[](unsigned int, nothrow);
+    freefunc = LibFunc_msvc_delete_ptr64_longlong;
+    break;
+
   case LibFunc_msvc_new_array_longlong:    // new[](unsigned long long);
   case LibFunc_msvc_new_array_longlong_nothrow: // new[](unsigned long long,
                                                 // nothrow);
+    freefunc = LibFunc_msvc_delete_array_ptr64_longlong;
+    break;
```
wsmoses, line 9197: Add some tests?

tthsqe12: Where would these tests go?

wsmoses: Inside the corresponding subfolder here https://github.com/EnzymeAD/Enzyme/tree/main/enzyme/test/Enzyme 

e.g. a reverse mode test in ReverseMode/ etc

tthsqe12: Do you have a CI job targeting the MS x64 ABI?

wsmoses: You shouldn’t need a CI job, just the LLVM emitted by a code with the MS x86 ABI

tthsqe12: I see. You are asking for the thing I feared you might ask for. So that I might make the uber MS test case, can you address the first question up there, namely, can we get `metaString->contains("enzyme_")` in Enzyme.cpp? Unfortunately, the `enzyme_dup, enzyme_const, ...` are mangled by `LLVMEnzyme.dll` here in the `metaString`. They are not mangled by `ClangEnzyme.dll`.

wsmoses: Would an extern "C" int enzyme_int; work?

I'm also fine supporting the mangled names, but we may want to not just do includes but instead unmangled name equals

@ZuseZ4 set that up previously for activity analysis so many be able to help.

tthsqe12: extern "C" seems to work fine, so it is just the c++ test code that needs to change. I'll do that too.

tthsqe12: I just added a file on which you can run opt and hit most of these changes.

tthsqe12:
@wsmoses  If you can figure out how to run the .ll file attached in [06af6af](https://github.com/EnzymeAD/Enzyme/pull/1338/commits/06af6af989ca8afb34439fb4d03bfa05e48864ad), then I suppose this is good to from my end.

tthsqe12:
@wsmoses This pr allows enzyme to work better with MS's ABI. Please let me know anything I can do further to get this merged. I am not sure how the testing system expects to process the attached ll file.

tthsqe12:
@wsmoses If this last commit is not the way to run the integration, please let me know. This PR should be close to being over the line.

tthsqe12:
@wsmoses I see that the CI for LLVM 15 has failed with the error message
`error: no check strings found with prefix 'CHECK:'`
Please advise on what to put after these CHECKs. I do not understand the existing examples.

wsmoses:
Yeah if you look at other test files, there should be some CHECK and
CHECK-NEXT lines at the bottom that will test if the output of the AD
matches some desired results.

That’s missing here hence the error

wsmoses:
I’d personally run the command on the file to find the output, check that
it it is as expected, then use that to write the run line

tthsqe12:
The output file has about 400 lines. From the other examples I saw, it looks like not all lines need to be included. Is this correct, that is, can I just select a few choice lines and include those in the check?

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -4468,10 +4468,14 @@ Constant *GradientUtils::GetOrCreateShadowConstant(
...
     }
+  } else if (auto arg = dyn_cast<GlobalAlias>(oval)) {
+    return GetOrCreateShadowConstant(Logic, TLI, TA, arg->getAliasee(), mode,
+                                     width, AtomicAdd);
   } else if (auto arg = dyn_cast<GlobalVariable>(oval)) {
     if (arg->getName() == "_ZTVN10__cxxabiv120__si_class_type_infoE" ||
         arg->getName() == "_ZTVN10__cxxabiv117__class_type_infoE" ||
-        arg->getName() == "_ZTVN10__cxxabiv121__vmi_class_type_infoE")
+        arg->getName() == "_ZTVN10__cxxabiv121__vmi_class_type_infoE" ||
+        arg->getName().startswith("??_R")) // any of the MS RTTI manglings
```
wsmoses, line 4478: Should all MS mangled globals have this treatment though? What about user defined globals!

enzyme/test/Integration/ReverseMode/MSinput.ll
```diff
@@ -0,0 +1,178 @@
...
+!11 = distinct !{!11, !12, !13}
+!12 = !{!"llvm.loop.mustprogress"}
+!13 = !{!"llvm.loop.unroll.disable"}
+!14 = distinct !{!14, !12, !13}
+!15 = !{!16, !16, i64 0}
+!16 = !{!"vtable pointer", !8, i64 0}
+!17 = !{!18, !10, i64 8}
+!18 = !{!"?AUObject@@", !10, i64 8}
+
+; CHECK: %rtti.CompleteObjectLocator = type { i32, i32, i32, i32, i32, i32 }
```
wsmoses, line 178: I would change the check lines to be the generated derivative functions

enzyme/Enzyme/GradientUtils.cpp
tthsqe12: I didn't like this either but there were too many, and I didn't know how many more there could be. I think this is not a problem because the closest an MS user can get from C/C++ is `?_R...` which has one question mark and comes from `_R`. Besides, such Identifiers are reserved and shouldn't be used anyways.

wsmoses (approved):

enzyme/test/Integration/ReverseMode/MSinput.ll
```diff
@@ -0,0 +1,180 @@
+; RUN: if [ %llvmver -eq 15 ]; then %opt < %s %newLoadEnzyme -enzyme -S | FileCheck %s; fi
```
wsmoses, line 1:
```suggestion
; RUN: if [ %llvmver -ge 15 ]; then %opt < %s %newLoadEnzyme -enzyme -S | FileCheck %s; fi
```

ZuseZ4:
@wsmoses this broke LLVM-16 CI

PR #1350, opened 2023-07-27
Add missing check for aggregate type

### LINKED ISSUE #1348: compiler crash on simple `__enzyme_fwddiff()` call
complete example: https://fwd.gymni.ch/lBJVX3

Specifying the wrong return type in `__enzyme_fwddiff` leads to a compiler crash rather than an error message.
It's not clear to me if this is a problem with enzyme or clang.

Tested with recent version of enzyme with clang 16

edit: code duplicated below:

```cpp
extern int enzyme_dup;
extern int enzyme_dupnoneed;
extern int enzyme_out;
extern int enzyme_const;

template < typename return_type, typename ... T >
extern return_type __enzyme_fwddiff(void*, T ... );

template < typename return_type, typename ... T >
extern return_type __enzyme_autodiff(void*, T ... );

double tolerance = 1.0e-15;

double f(double x) { return x * x; }

int main() {
  double x = 1.7;
  double dx = 1.2;
  double J = 2 * x;

  // wrong return type specified, crashes
  __enzyme_fwddiff<float>((void*)f, enzyme_dup, x, dx);

  // correct return type specified, compiles without issue
  __enzyme_fwddiff<double>((void*)f, enzyme_dup, x, dx);
}
```

tgymnich:
fixes #1348

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -496,7 +496,8 @@ static bool ReplaceOriginalCall(IRBuilder<> &Builder, Value *ret,
     return true;
   }
 
-  if (mode != DerivativeMode::ReverseModePrimal) {
+  if (mode != DerivativeMode::ReverseModePrimal &&
+      diffret->getType()->isAggregateType()) {
     auto diffreti = Builder.CreateExtractValue(diffret, {0});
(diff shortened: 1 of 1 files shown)


wsmoses (approved):
lgtm, add a test to prevent regression?

PR #1359, opened 2023-07-30
Begin better C++ language level support

wsmoses:
@jedbrown @LeilaGhaffari @samuelpmishLLNL if you look at the customlog1p.c test below, you'll see that now the custom behaves like can now take a proper attribute instead of the global (`__attribute__((enzyme_function_like("log1p")))`) Note that essentially that attribute is syntactic sugar for the global, which is later parsed and removed in LLVM.

Nevertheless, it makes things slightly nicer.

Moreover, now using ClangEnzyme, there will automatically be `ENZYME_VERSION_MAJOR` and related macros defined if Enzyme is available, and not if it isn't. That means that one can do the following as the enzyme availability check
```
#ifdef ENZYME_VERSION_MAJOR
 .. code that requires Enzyme
#endif
```

Finally, I'm not sure (and CI will inform), but hopefully the args required for Enzyme are now simpler, just requiring `/clang-16   -fplugin=/path/to/Enzyme/ClangEnzyme-16.so ` rather than the version specific other variations.


wsmoses:
@tgymnich @ZuseZ4 in principle, after we add an enzyme_inactive variant, we may be able to auto include a header marking various c++ std io as inactive, perhaps reducing the size of the activityanalysis list. However, we may need to add an arg to the clang plugin which specifies not running the actual AD transformation during clang, but only preserve-nvvm and the relevant arg parsing.

Part of the initial work here was doing with the hope of eventually having C++ types parsed and preserved on functions, bypassing the need for interprocedural type analysis, like in Julia

tgymnich:
I would love to get rid of the global variables at some point, since the code for parsing and handling all this is super ugly.

samuelpmishLLNL:
@wsmoses thanks for looking into improving the C++ interface!

----------

> I would love to get rid of the global variables at some point, since the code for parsing and handling all this is super ugly.

One thing I've been thinking is that encoding the annotations (`enzyme_dup, enzyme_dupnoneed, enzyme_const`) could be accomplished in a more C++-friendly way through some specific types. e.g. instead of

```cpp
__enzyme_fwddiff<return_type>((void*)f, enzyme_const, x, enzyme_dup, y, dy);
```

one could write

```cpp
enzyme::forward(foo, enzyme::constant(x), enzyme::dup(y, dy));
```

Rather than manually specifying the return type, it can be determined internally by doing

```cpp
using return_type = decltype(foo(x, y));
```

Also, the `enzyme_dupnoneed` could be accomplished with

```cpp
enzyme::reverse(foo, enzyme::constant(x), enzyme::dup(enzyme::unused{}, dy));
```

------------

Some where the implementations of the new types above are essentially just

```cpp
// enzyme.hpp
namespace enzyme {

  template < typename T >
  struct constant { T value; };

  // or maybe a more descriptive name?
  template < typename S, typename T >
  struct dup { S value; T shadow; };

  struct unused{};

}
```

wsmoses:
@samuelpmishLLNL so I was thinking about having attributes express the custom derivatives (this is actually what we did long ago, before we were a plugin).

Something perhaps like

__attribute__((enzyme_custom_reverse(augmentfwd, reverse)))
double myfunc(...);

As a bonus, it also works for C code, and also if you're not compiling with enzyme will just produce a warning unknown/unused attribute

wsmoses:
That said the template/overload variant would also provide a lot of functionality to only overload one set of activities (e.g. if this is constant, that is active), see the Julia custom rules for example: https://enzyme.mit.edu/julia/stable/generated/custom_rule/

Currently the global variable version auto detects/assumes all variables are active, which is prone to errors/difficulty.

PR #1362, opened 2023-07-31
Add option to continue to execute unreachable code

wsmoses:


enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -1081,9 +1087,11 @@ void calculateUnusedValuesInFunction(
     for (auto &BB : func)
       for (auto &I : BB) {
         bool ivn = DifferentialUseAnalysis::is_value_needed_in_reverse<
-            ValueType::Primal>(gutils, &I, mode, PrimalSeen, oldUnreachable);
+            ValueType::Primal>(gutils, &I, mode, PrimalSeen,
+                               gutils->notForAnalysis);
```
tgymnich, line 1091: Could we get the `mode` and `oldUnreachable` from `gutils` instead of passing each to the constructor?

PR #1366, opened 2023-08-03
Blas opt

ZuseZ4:


ZuseZ4:
Not production ready and also not enabled by default, but let me merge this first, then I can fix the new PM and add more assertions while working on the next optimization.

enzyme/Enzyme/OptBlas.cpp
```diff
@@ -0,0 +1,134 @@
+//===- PreserveNVVM.cpp - Mark NVVM attributes for preservation.  -------===//
```
wsmoses, line 1: rename comment?

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -3490,6 +3491,10 @@ llvmGetPassPluginInfo() {
             PB.registerPipelineParsingCallback(
                 [](llvm::StringRef Name, llvm::ModulePassManager &MPM,
                    llvm::ArrayRef<llvm::PassBuilder::PipelineElement>) {
+                  // if (Name == "blas-opt") {
```
wsmoses, line 3494: add new pm version?

enzyme/test/Integration/BlasOpt/first.c
```diff
@@ -0,0 +1,42 @@
+enum CBLAS_ORDER {CblasRowMajor=101, CblasColMajor=102};
```
wsmoses, line 1: add a test line?

ZuseZ4: How? Results are same, it would need to go into the benchmark section later.
Only added it for now to keep track of source examples.

wsmoses: look at the other integration tests. There's a line at the top that wlll cause this to be compiled and run (including doing any verification the results are the same assertions)

wsmoses: So if you have a fixed input, and thus know the correct output, you can assert it always gives that, even with the optimization

ZuseZ4: Ah sure, I can do that.
I was more wondering how to check that the optimization will actually be applied.
But I guess for that we have llvm tests.

wsmoses:
@ZuseZ4 this still looks useful and seemed like just needed some love to get over the finish line to merge?

PR #1367, opened 2023-08-06
add enzyme header and fix capitalization issue with installation directory

samuelpmishLLNL:
when writing some C++ examples for the enzyme webpage, it seemed like it would be easier to just put the enzyme boilerplate

```cpp
int enzyme_dup;
int enzyme_dupnoneed;
int enzyme_out;
int enzyme_const;

template < typename return_type, typename ... T >
return_type __enzyme_fwddiff(void*, T ... );

template < typename return_type, typename ... T >
return_type __enzyme_autodiff(void*, T ... );
```

into its own header rather than copying and pasting it into every example. User projects could do this too, but it seems more natural to put this into the Enzyme repo once and for all. I wasn't sure what to call this header since there's already an "Enzyme.h" in the project, so I just call it "declarations.hpp" and put it in a directory such that the actual #include statement reads

```cpp
#include "enzyme/declarations.hpp"
```

There's some small cmake changes to install this new header with the other Enzyme targets. There's also a fix where Enzyme was previously installing its cmake config files to `/path/install/lib/CMake/Enzyme` rather than the conventional `/path/install/lib/cmake/Enzyme` which was confusing CMake's `find_package` calls on linux machines.

These CMake changes have been tested on ubuntu 22.04 with llvm versions 11-16 with `add_subdirectory(path/to/enzyme)` and `find_package(enzyme ...)`.


enzyme/Enzyme/include/enzyme/declarations.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+extern int enzyme_dup;
+extern int enzyme_dupnoneed;
+extern int enzyme_out;
+extern int enzyme_const;
+
+template <typename return_type, typename... T>
+return_type __enzyme_fwddiff(void *, T...);
```
tgymnich, line 9: We could also template the function type.
```suggestion
template <typename return_type, typename F, typename... T>
return_type __enzyme_fwddiff(F, T...);
```

enzyme/Enzyme/include/enzyme/declarations.hpp
```diff
@@ -0,0 +1,12 @@
...
+
+template <typename return_type, typename... T>
+return_type __enzyme_autodiff(void *, T...);
```
tgymnich, line 12:
```suggestion
template <typename return_type, typename F, typename... T>
return_type __enzyme_autodiff(F, T...);
```

tgymnich (changes requested):

enzyme/Enzyme/include/enzyme/declarations.hpp
samuelpmishLLNL: Out of curiosity, is it feasible to make `__enzyme_fwddiff` (and similar) support stateful lambdas directly? At present, users can differentiate stateful lambdas through a wrapper: https://fwd.gymni.ch/cKDhZ5 but it places more effort on the user to adapt their code to meet Enzyme's requirements.

tgymnich: ok lets make it a pointer then.
```suggestion
template <typename return_type, typename F, typename... T>
return_type __enzyme_fwddiff(F*, T...);
```

wsmoses: I think so, but it would end up looking more similar to the Julia API for Enzyme. Specifically, if there is state, one may need to have a const vs duplicated vs active (aka enzyme_out) version of the lambda.

wsmoses:
If we're going to make a C++ only header (aka no c), I'm kind of inclined to make an API sort of like:

auto vars = enzyme::autodiff(func, enzyme::Const(x), enzyme::Duplicated(y, dy), enzyme::Active(z));

We can make the LLVM/Clang auto generate the requisite C++ return types and inner definition code.

```cpp
namespace enzyme {
   template<typename T>
   struct ActivityType {
   };
   template<typename T>
   struct ConstType : ActivityType<T> {
      T &&val;
   }
   ...
   
   struct NoDiff {};
   
   // return the inner type of ActivityType<T> if ActivityType == Active, else return NoDiff
   template<typename T>
   inner_type_if_active_else_nodiff<T>
   
   template<typename Args...>
   struct ReverseModeTuple {
        inner_type_if_active_else_nodiff<Arg0> arg0;
        inner_type_if_active_else_nodiff<Arg1> arg1;
        ...
   };
   
   template<typename Fn, typename... ActivityArgs> // make a constraint that fn/activityargs are subset of ActivityType? 
   ReverseModeTuple<Fn, ActivityArgs...> autodiff(Fn &&, ActivityArgs... &) {
      assert(0 && "this definition will be autofilled by the compiler");
   }
}
```

wsmoses:
Now @samuelpmishLLNL I'm not a sufficient C++ expert to know how to design the above C++ side offhand, but if you can prototype something, I can start doing the parts to hook up the compiler side for autofilling in/generating/extracting the variables.

enzyme/Enzyme/include/enzyme/declarations.hpp
samuelpmishLLNL: in practice, I think stateful lambdas would need to be implicitly `enzyme_const` w.r.t. the lambda member variables, since (in C++) there's no natural way to copy a lambda and access its member variables.

wsmoses:
Incidentally this style of API would also work around the C++ lambda local scope limitation, and we could also add actual type checking, including fixing the pass by reference mismatch.

wsmoses:
also @samuelpmishLLNL can you extract out the cmake fix to a different PR, since I think we may discuss API design longer than we should wait for a cmake fix xD

enzyme/Enzyme/include/enzyme/declarations.hpp
wsmoses: Yeah the julia default is to have them const, unless explicitly passed a duplicated one. Unfortunately, it may be necessary to have a dup version if used as a temporary, or one otherwise wants to diff wrt.

tgymnich (approved):

enzyme/CMakeLists.txt
```diff
@@ -230,7 +229,7 @@ set(INSTALL_INCLUDE_DIR include CACHE PATH "Installation directory for header fi
 if(WIN32 AND NOT CYGWIN)
   set(DEF_INSTALL_CMAKE_DIR CMake)
```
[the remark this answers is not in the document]
samuelpmishLLNL: I don't know the naming conventions for installation directories on windows. 

Another thing I was wondering is why these `DEF_INSTALL_CMAKE_DIR` and `INSTALL_CMAKE_DIR` variables are set up in the first place-- I thought the default built-in `CMAKE_INSTALL_PREFIX` handled this sort of thing automatically.

PR #1391, opened 2023-08-29
fix a control flow bug

tthsqe12:
@wsmoses I am trying to fix some bugs in Enzyme. There are at least two bugs in `__enzyme_reverse` in play here:
1. `getDynamicLoopLimit` creates an undefined value if the loop induction variable does not belong to the immediate predecessor. Here, if `f` is exited via `B`, then `A`'s loop induction variable is the one that becomes undefined in the block of code called `return.loopexit`. I see nothing wrong with "fixing" this the way I did here.
2. The previous bug occurs in the forward portion of `__enzyme_reverse`. There is also a bug in the reverse portion, whereby `A`'s induction variable (iv ac) is simply reading junk from an uninitialized alloca. The code responsible for making this decision is `ComputeIndexOfChunk`. However, I could not infer what your intentions were with this part of the code, so I have not tried to fix it.

The executable crashes with -O0, but higher optimizations may mask the junk values.

### DIFF
--- a/enzyme/Enzyme/CacheUtility.cpp
+++ b/enzyme/Enzyme/CacheUtility.cpp
@@ -459,11 +459,7 @@ llvm::AllocaInst *CacheUtility::getDynamicLoopLimit(llvm::Loop *L,
     auto Limit = B.CreatePHI(found.var->getType(), 1);
 
     for (BasicBlock *Pred : predecessors(ExitBlock)) {
-      if (LI.getLoopFor(Pred) == L) {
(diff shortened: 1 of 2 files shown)


tthsqe12:
I can do that, but the given C code and clang 15 should be sufficient to reproduce everything exactly.

tthsqe12:
It turn's out that Enzyme has changed a bit since I last looked at this bug, so my analysis is not completely accurate anymore. In particular, the 'fix' doesn't fix the first problem anymore. I am moving this to an issue.

PR #1396, opened 2023-08-31
fix invsqrt example

tthsqe12:
Previous code always returned 0 for 64 bit longs.


tthsqe12:
Still doesn't resolve the issue of why `aug_rsqrt` is allowed to be empty here.

tgymnich (approved):

Issue #1394, opened 2023-08-30
Assertion `f != originalToNewFn.end()' failed with Eigen


gaurav-arya:
MWE reduced to only Eigen:
```c
#define _USE_MATH_DEFINES
#include <cmath>
#include <iostream>
#include <Eigen/Core>

extern double __enzyme_autodiff(void *, double);
typedef float T;
typedef Eigen::Matrix<T, 3, 1> State;

class LinearizedSystemModel
{
public:
    Eigen::Matrix<typename State::Scalar,State::RowsAtCompileTime, State::RowsAtCompileTime> F;
    
    virtual void updateJacobians( const State& x)
    {
        // No update by default
        (void)x;
    }
};

double simulate(double input) {
  LinearizedSystemModel sys;
  sys.F.setIdentity();
  sys.F  = ( sys.F * sys.F * sys.F.transpose() ) + ( sys.F * sys.F * sys.F.transpose() );

  return 0.0;
}

int main(int argc, char **argv) {
    double df_dx = __enzyme_autodiff((void *)simulate, 1.0);
    return 0;
}
```
Error:
[out.txt](https://github.com/EnzymeAD/Enzyme/files/12479715/out.txt)

cc @ZuseZ4

ZuseZ4:
https://fwd.gymni.ch/sdcQmZ
http error code 0. cc @tgymnich playground seems unable to access eigen?

wsmoses:
```
@_ZTI21LinearizedSystemModel = internal constant { ptr, ptr } { ptr getelementptr inbounds (ptr, ptr @_ZTVN10__cxxabiv117__class_type_infoE, i64 2), ptr @_ZTS21LinearizedSystemModel }, comdat, align 8
ld.lld: /home/gridsan/aryag/Enzyme/enzyme/Enzyme/GradientUtils.cpp:550: llvm::Value* GradientUtils::getNewFromOriginal(const llvm::Value*) const: Assertion `f != originalToNewFn.end()' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
[...]

PR #1412, opened 2023-09-03
Add const global test

gaurav-arya:
Add test from #1394


ZuseZ4:
Iirc that bug is only reproducible with opaque-ptr=1, so we might need to adjust the RUN line here to exclude too old llvm versions which don't support opaque ptrs sufficiently, let's see what CI says.

enzyme/test/Enzyme/ReverseMode/constglobal.ll
```diff
@@ -0,0 +1,138 @@
+; RUN: if [ %llvmver -lt 16 ] && [ %llvmver -ge 15 ]; then %opt < %s %loadEnzyme -enzyme -opaque-pointers=1 -S | FileCheck %s; fi
```
tgymnich, line 1:
```suggestion
; RUN: if [ %llvmver -eq 15 ]; then %opt < %s %loadEnzyme -enzyme -opaque-pointers=1 -S | FileCheck %s; fi
```

tgymnich (changes requested):

wsmoses (approved):

PR #1414, opened 2023-09-04
Update benchmark.yml

tgymnich:


tgymnich:
@wsmoses we are encountering out of memory exceptions.
```
Sep 05 13:20:15 github22-ci-3 systemd[1]: actions.runner.EnzymeAD.github22-ci-3.service: Failed with result 'oom-kill'.
```

PR #1425, opened 2023-09-15
controlflow fixes

### LINKED ISSUE #1392: bug in loop counters generates code that crashes
compile the C code in the next post with clang15 and something roughly equivalent to `"%clang -std=c11 -O0 -fno-unroll-loops -fno-tree-vectorize %s -S -emit-llvm -o - %newLoadClangEnzyme"`. The most important thing is `-O0` so that things are clear. The code is really quite minimal in terms of both manifesting the bug in the llvm and at run time.

Some important variables to keep in mind when reading the llvm in my next post:
`%10` is `A`'s tape
`%iv` is `A`'s loop counter

problem no. 1

The problem occurs when the original code exits via the `B` block. In this case the block `return.loopexit` is reached and the following is obviously a problem.
```ll
return.loopexit:
  %19 = getelementptr inbounds i64, ptr %.lcssa27, i64 undef ; BAD
  %20 = load i64, ptr %19, align 8, !invariant.group !6 ; %20 is supposed to be B's loop counter as loaded off of A's tape
```

problem no. 2
`"iv'ac.0"` is used in `invertA` as `A`'s reverse loop counter. This variable comes from the following places:
```ll
invertA:
  %"iv'ac.0" = phi i64 [ %"iv'ac.1", %invertif.else ], [ %iv.lcssa, %invertA.loopexit9 ]

invertA.loopexit9:
  %iv.lcssa = phi i64 [ %iv, %__enzyme_exponentialallocation.exit ]                     ; all good

invertif.else:
  %"iv'ac.1" = phi i64 [ %40, %incinvertA ], [ %"iv'ac.2", %invertB.preheader ]         ; all good. note %40 = %"iv'ac.0" - 1

invertB:
  %"iv'ac.2" = phi i64 [ %"iv'ac.3", %invertif.else35 ], [ undef, %return.loopexit ]    ; second is BAD

invertif.else35:
  %"iv'ac.3" = phi i64 [ %40, %invertA.loopexit ], [ %"iv'ac.2", %incinvertB ]          ; all good
```
Again, it is clear that enzyme has gotten confused about `invertA`'s loop counter when coming from `invertB`.

tthsqe12:
test code from  #1392


tthsqe12:
@wsmoses Could you say something about the viability of these changes before I start changing all of your doc tests?

wsmoses:
Yeah, I'll take a look in the next few days

enzyme/Enzyme/CacheUtility.cpp
```diff
@@ -459,14 +459,18 @@ llvm::AllocaInst *CacheUtility::getDynamicLoopLimit(llvm::Loop *L,
     auto Limit = B.CreatePHI(found.var->getType(), 1);
 
     for (BasicBlock *Pred : predecessors(ExitBlock)) {
-      if (LI.getLoopFor(Pred) == L) {
+      if (LI.getLoopFor(Pred) == L || L->contains(Pred)) {
```
wsmoses, line 462: Okay I agree with this change, as previously this had an assumption that an ExitBlcok would only be the exitblock for a single loop.

Will try to ponder a different test of this regarding the other changes.

wsmoses:
I inspected this more deeply, and at least one problem that this partially fixes is the fact that Enzyme (incorrectly) assumed that an exit block only had one loop predecessor. I fixed that assumption here: https://github.com/EnzymeAD/Enzyme/pull/1430 . That is effectively your change, however, the `store limit` part unfortunately had to be a lot more complex (see the PR).

I'm not sure that is the entirety of the issue you saw, but try that out and take a look?

tthsqe12:
Yes, there are two issues that the AB loops illuminate. Your PR does (1). What remains is (2) the antivars may be loaded uninitialized in the reverse pass. You can see these in the llvm in EnzymeLogic after enzyme has done its thing but before the post optimization passes are run. The promote pass from llvm is what obscures the bad loads. There is a function computeIndexOfChunk that loads from uninitialized antivars. This happens in ...'merge blocks that only initialize one of the antivars. This doesn't work because the whole tower of antivars needs initialization.

Try looking at the llvm (before post optimization) for the C code I posted.

tthsqe12:
@wsmoses I see that you did have a try at (2) in #1430 for v0.0.84. This gets closer, but it is not there yet. Here is the problem very specifically:
Suppose we have loop C inside a loop B inside a loop A, with induction vars ivC, ivB, ivA and limit caches llcC, llcB, llcA.
The primal code now has a correct block that looks like:
```return.loopexitC:
llcA = ivA
llcB[ivA] = ivB
llcC[ivA][ivB] = ivC
br return
```
Now, ivC'ac, ivB'ac, ivA'ac are uninitialized allocas to circumvent the SSA requirement. As for the reverse code:
*What we see in v0.0.82*:
```
invertreturn.loopexitC:
t1 = llcC[ivA'ac][ivB'ac]
ivC'ac = t1
```
This is doubly wrong because ivA'ac and ivB'ac are uninitialized when read, and ivA'ac and ivB'ac are not set. As a consequence of the first, llcC[ivA'ac][ivB'ac] reads from an undefined address.
*What we see in v0.0.84*:
```
invertreturn.loopexitC:
t3 = llcA
t2 = llcB[ivA'ac]
t1 = llcC[ivA'ac][ivB'ac]
ivA'ac = t3
ivB'ac = t2
ivC'ac = t1
```
This is getting closer but is still bad because ivA'ac and ivB'ac are uninitialized when used.
*What we need to see*:
```
invertreturn.loopexitC:
t3 = llcA
ivA'ac = t3
t2 = llcB[ivA'ac]
ivB'ac = t2
t1 = llcC[ivA'ac][ivB'ac]
ivC'ac = t1
```
Since I considered this an invasive change, in this PR I just put the initializations of ivA'ac, ivB'ac, and ivC'ac in the primal part.
Please decide where they best go.

wsmoses:
Excellent catch, I believe https://github.com/EnzymeAD/Enzyme/pull/1443 should remedy what you found

tthsqe12:
Looks good. Would you like the test code here in the integration folder?

wsmoses:
Yeah, the more the merrier. Rebase and let's merge?

tthsqe12:
This has only the test code now. By the way, the following tests broke sometime between v0.0.70 and v.0.0.80
```
ReverseMode/fbuff.cpp -O0 -std=c++11 -fno-exceptions -ffast-math
ReverseMode/map.cpp -O0 -std=c++11 -fno-exceptions -ffast-math
```

wsmoses (approved):

PR #1440, opened 2023-09-19
[BLAS] fix row transpose arg

wsmoses:


wsmoses:
@ZuseZ4 @gbaraldi this did not resolve the segfault of the test. Was this the correct fix ?

ZuseZ4:
I need to look at the Fortran code again, I think it might miss a bit.

ZuseZ4:
Found a second bug for gemv

ZuseZ4:
should be a strict improvement, so let's merge.
gemm (alpha, beta) might need more adjustments, but we can do that in a follow-up pr.

ZuseZ4 (approved):

PR #1453, opened 2023-09-23
cublas support

ZuseZ4:


enzyme/test/Integration/blasinfra.h
```diff
@@ -513,6 +588,30 @@ void cblas_dgemm(char layout, char transA, char transB, int M, int N, int K, dou
                                 M, N, K, lda, ldb, ldc});
 }
 
+using cublasOperation_t::CUBLAS_OP_UNUSED;
+
+__attribute__((noinline))
+cublasStatus_t cublas_ddot(cublasHandle_t *handle, int N, double* X, int incx, double* Y, int incy, double* result) {
+    CuBlasCall call = {inDerivative, CallType::DOT,
```
wsmoses, line 595: I would probalby not make a separate cucalls, but add an extra arg to calls saying device or API

ZuseZ4: I did consider / tried that for a while, but it turned out to be a bit annoying for a few reasons.
Let me try this one first, if I run into stronger issues or later I can still merge it?

enzyme/Enzyme/Utils.cpp
```diff
@@ -2511,9 +2548,18 @@ llvm::Value *is_normal(IRBuilder<> &B, llvm::Value *trans, bool byRef) {
 // However, if we ask openBlas c ABI,
 // it is one of the following 32 bit integers values:
 // enum CBLAS_TRANSPOSE {CblasNoTrans=111, CblasTrans=112, CblasConjTrans=113};
-llvm::Value *transpose(IRBuilder<> &B, llvm::Value *V) {
+llvm::Value *transpose(IRBuilder<> &B, llvm::Value *V, bool cublas) {
   llvm::Type *T = V->getType();
   Value *out;
+  if (cublas) {
```
ZuseZ4, line 2554: @wsmoses 
This relies on the integer representation of enums in c++, I took the order from some docs, is there a nicer way?

wsmoses: We can write our own enum or constant variables Eg Cublas_N to make it nicer, but unfortunately otherwise not. We just need to keep the ABI the same and we don't want to require cuBLAS to compile enzyme.

enzyme/Enzyme/Utils.cpp
```diff
@@ -655,8 +655,7 @@ void callMemcpyStridedBlas(llvm::IRBuilder<> &B, llvm::Module &M, BlasInfo blas,
 void callMemcpyStridedLapack(llvm::IRBuilder<> &B, llvm::Module &M,
                              BlasInfo blas, llvm::ArrayRef<llvm::Value *> args,
                              llvm::ArrayRef<llvm::OperandBundleDef> bundles) {
-  std::string copy_name =
-      (blas.prefix + blas.floatType + "lacpy" + blas.suffix).str();
+  std::string copy_name = blas.prefix + blas.floatType + "lacpy" + blas.suffix;
```
tgymnich, line 658: unnecessary heap allocations for construction temporary std::strings!
https://godbolt.org/z/Ecdcc5beK

wsmoses:
Note macos Julia CI is failing which we'll need to fix. Merging now to avoid a bunch of git nonsense per @ZuseZ4

PR #1472, opened 2023-10-10
ensure that double stores to tape do not incure additional mallocs

tthsqe12:
Previous version of enzyme vacillated between (1) storing double directly on the tape, and (2) storing a malloc(8) pointer on the tape and then storing the double in this new pointer. Since (1) is greatly to be preferred from a performance and memory consumption perspective, this test is meant to ensure that enzyme doesn't regress again in this aspect.


tthsqe12:
@wsmoses  It looks like enzyme now stores doubles directly on the tape (yay!), and this is what I am checking here. However, I have seen it - in the case of caching some integer quantities - doing the unnecessary extra indirection:
```
tape = malloc(...)
int* p = malloc(sizeof(int))
*(tape + some_offset) = p
// now use *p as the cache
```
There is no need to do this as the int can live directly on the tape. Could you possibly look into eliminating more of these extra fixed-size mallocs? I can give examples if needs be.

PR #1473, opened 2023-10-10
add pedagogically sound examples for custom derivatives

### LINKED ISSUE #1395: Assertion `!subretused' failed when compiling simple example with custom derivatives.
I don't see anything wrong with doing
```cpp
extern double F(double);
extern double Fp(double);

__attribute__((noinline)) double f(double x) {return F(x);}
void f_aug(double x) { }
double f_rev(double x, double grad_out) {return grad_out*Fp(x);}
void* __enzyme_register_gradient_f[3] = { (void*)f, (void*)f_aug, (void*)f_rev };

void g(double x[5]) {
    x[2] = f(x[0] + x[1]);
}

template<typename ...T> void __enzyme_autodiff(void*, T...);

void g_auto(double x[5], double x_b[5]) {
    __enzyme_autodiff((void*)g, x, x_b);
}
```
but https://github.com/EnzymeAD/Enzyme/blob/b2f11907b4153371c58cece26e91aa71f31e8b63/enzyme/Enzyme/AdjointGenerator.h#L7257 fails here. My command line is nothing unusual
```
0.	Program arguments: /home/pc/cpp/llvm-project-llvmorg-15.0.7/build/bin/clang++ -O3 -fno-exceptions tcustom.cpp -S -emit-llvm -o output.ll -fpass-plugin=/home/pc/cpp/git/Enzyme/enzyme/build/Enzyme/ClangEnzyme-15.so -Xclang -load -Xclang /home/pc/cpp/git/Enzyme/enzyme/build/Enzyme/ClangEnzyme-15.so
```

tthsqe12:
resolves https://github.com/EnzymeAD/Enzyme/issues/1395
As the integration folder seems to be the best documentation of the C/C++ interface, it is important to have high-quality examples here.

### DIFF
--- a/enzyme/test/Integration/ReverseMode/invsqrt.c
+++ b/enzyme/test/Integration/ReverseMode/invsqrt.c
@@ -35,7 +35,7 @@ float Q_rsqrt( float number )
 
   x2 = number * 0.5F;
   y  = number;
-  i  = * ( int32_t * ) &y;                       // evil floating point bit level hacking
+  i  = * ( int32_t * ) &y;                    // evil floating point bit level hacking
   i  = 0x5f3759df - ( i >> 1 );               // what the [...]?
   y  = * ( float * ) &i;
   y  = y * ( threehalfs - ( x2 * y * y ) );   // 1st iteration
@@ -51,7 +51,7 @@ double invmag(double* __restrict__ A, int n) {
   return Q_rsqrt(sumsq);
 }
 
-
+//// (1) An interface that just happens to work because Q_rsqrt is simple enough.
 
 // Returns { optional tape, original return (if pointer), shadow return (if pointer) }
 void aug_rsqrt(float x) {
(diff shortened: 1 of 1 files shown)


enzyme/test/Integration/ReverseMode/invsqrt.c
```diff
@@ -51,7 +51,7 @@ double invmag(double* __restrict__ A, int n) {
   return Q_rsqrt(sumsq);
 }
 
-
+//// (1) An interface that just happens to work because Q_rsqrt is simple enough.
```
wsmoses, line 54: So the issue here isn't because of Q_sqrt, but rather it is because of the use of Q_sqrt within the function being differentated.

In particular, someones one may need the original primal return to compute the derivative of a function. e.g. if we did expr(Q_sqrt(x)). However, since we simply return Q_rsqrt (and the default semantics of autodiff without a flag is to just return the derivative, not the primal), we didn't need it.

In the second case, becuase it is stored into a duplicated, the same value must be stored in the derivative (and thus the primal return is required). If the variable A was marked dup_noneed, then the return-less version would've been fine.

The julia custom rule API goes into more details: https://enzyme.mit.edu/index.fcgi/julia/stable/generated/custom_rule/

That said this is very undocumented, and counter-intuitive.

@samuelpmish and I are working on trying to make a nicer c/c++ interface, including for custom derivatives so this is a great example to have.

enzyme/test/Integration/ReverseMode/invsqrt.c
```diff
@@ -66,33 +66,79 @@ float rev_rsqrt(float x, float grad_out) {
...
+    double* tape = malloc(sizeof(double));
+    *tape = *x;
+    *y = Q_rsqrt(*x);
+    return tape;
+}
+void rev_rsqrt3(double* y, double* y_b, double* x, double* x_b, void* tape) {
+    double x0 = *(double*)tape; // x points to junk; original input *x must be obtained from the tape
+    *x_b -= (*y_b) * Q_rsqrt(x0) / (2 * x0);
+    *y_b = 0;   // Since y is used purely as an output, changes in y do not affect the calculation
+    free(tape);
```
jedbrown, line 96: This is perhaps off-topic, but what if the user wishes to apply reverse mode multiple times? Should freeing the tape not be decoupled from the numerics of the reverse pass?

wsmoses: If the code is generated by Enzyme, the no-free flag will keep that cache around. We need to figure out a nice interface for specifying that (and other options) for the custom derivatives.

Alternatively, perhaps we just have a special allocator that manages this behind the scenes.

wsmoses: Also at least in this case, one doesn't need to return a void*, one can just use a double, rather than double* as the tape.

tthsqe12: I agree that in this case the tape can be of type double, but a void star will get you a lot further in more general cases. I will respond to the initial question. It is my understanding that when one writes an augmented function (f_fwd) and a reverse function (f_rev), one is writing a pair of coupled functions; every call to the augmented forward is expected to eventually followed up by precisely one corresponding call to the reverse function - and with the corresponding tapes. Given this interface, there this no other place but the reverse function to cleanup whatever mess was made by the augmented forward function. Furthermore, since the same code is run for multiple calls to f (suppose you want to differentiate a function g which calls f twice), the tape is the only way in general to distinguish the two separate pieces of data that f_rev will need.
I am not sure what this no-free flag is. However, I do know the enzyme_allocated option can handle fairly general cases - but this option is not (and cannot be?) integrated with the custom derivatives.

enzyme/test/Integration/ReverseMode/invsqrt.c
tthsqe12: Ah, yes. I see. It is nice to know what legalCombinedForwardReverse does now.
For some reason, replacing the return with `return __builtin_exp(Q_sqrt(sumsq))` wasn't enough to trigger the assert, but additionally deleting the definition of `Q_sqrt` and replacing it with `extern float Q_sqrt(float)` was.

PR #1482, opened 2023-10-13
Add flag to disable AD.

ludgerpaehler:
cc @wsmoses

### DIFF
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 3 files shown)


wsmoses:
@ludgerpaehler this needs to import the commandline headers in LLVM <= 13, if you can add that fix let's merge it!

wsmoses:
@ZuseZ4 @gaurav-arya for context, this allows us to still use the ClangEnzyme nicer syntactic sugar/etc, but still use LLDEnzyme for the actual differentiation

PR #1485, opened 2023-10-15
Tblgen diffuse

wsmoses:
@ludgerpaehler see if using this improves the alpaca stuff.

Regardless, this should also introduce some better logging for -enzyme-print-diffuse and -enzyme-unnecessary to try to find the root cause of the issue you saw.


ludgerpaehler:
Thank you! Will try it out right now!

PR #1501, opened 2023-10-24
Fix transposed lda on cache

wsmoses:
@gaurav-arya this was resulting in memory errors if a matrix was cached and non-square


ZuseZ4 (approved):

ZuseZ4:
@wsmoses /h/344/drehwald/prog/Enzyme/enzyme/test/Enzyme/ReverseMode/blas/gemv_c_loop3_matcopy.ll 
switched from select false in all cases to select true in all cases.
Looks suspicious to me.

PR #1521, opened 2023-11-03
Ring of springs integration test using modulo

martinjm97:
This PR is a slight change to the simplified springs example that changes the topology of the springs so that is a ring of 1D springs rather than a line of springs. Executing this code produced the following result on my machine:
```row=0 col=7 val=nan
Assertion failed: (abs(val) > 0.00001), function inner_store, file new_sparse.cpp, line 68.
[1]    24375 abort      ./new_sparse.out
```


enzyme/test/Integration/Sparse/ringspring.cpp
```diff
@@ -59,14 +66,14 @@ void ident_store(double , int64_t idx, size_t i) {
 __attribute__((always_inline))
 double ident_load(int64_t idx, size_t i, size_t N) {
     idx /= sizeof(double);
-    return (double)(idx == i);// ? 1.0 : 0.0;
+    return (double)(idx % N == i % N);// ? 1.0 : 0.0;
```
wsmoses, line 69: this should not be modulo'd

martinjm97:
The requested. changes are in the updated branch. The dense code now works, but we run into issues for the auto-sparsified code.
> Assertion failed: (i_nocapture < OperandTraits<PHINode>::operands(this) && "getOperand() out of range!"), function getOperand, file Instructions.h, line 2956.

PR #1549, opened 2023-11-15
Jmp/dataflow activity analysis

wsmoses:


tgymnich (commented):
You should try to reduce the nesting depth of your code a bit. E.g. add some early returns / continues where possible.
Here is what I would suggest: https://github.com/EnzymeAD/Enzyme/pull/1552

enzyme/Enzyme/MLIR/Analysis/AliasAnalysis.h
```diff
@@ -0,0 +1,176 @@
+#ifndef ENZYME_MLIR_ANALYSIS_DATAFLOW_ALIASANALYSIS_H
```
ftynse, line 1: Can we consistently add license headers to files, same as LLVM and MLIR do?

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+      }
+    }
+  }
+
+private:
+  // A pointer to the entry block and argument activities of the top-level
+  // function being differentiated. This is used to set the entry state
+  // because we need access to the results of points-to analysis.
+  Block *entryBlock;
+  ArrayRef<enzyme::Activity> argumentActivity;
```
ftynse, line 575: (+) It's dangerous to store a non-owning reference here. I'd rather store a copy, the loss of space is marginal compared to the case where the analysis object outlive whoever owns this.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+                                    const BackwardMemoryActivity &after,
+                                    BackwardMemoryActivity *before) override {
+    meet(before, after);
+  }
+
+  void setToExitState(BackwardMemoryActivity *lattice) override {}
+
+private:
+  FunctionOpInterface parentOp;
+  ArrayRef<enzyme::Activity> argumentActivity;
```
ftynse, line 700: Ditto.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+};
+
+void traverseCallGraph(FunctionOpInterface root,
+                       SymbolTableCollection *symbolTable,
+                       function_ref<void(FunctionOpInterface)> processFunc) {
+  std::deque<FunctionOpInterface> frontier{root};
```
ftynse, line 706: Nit: using `SmallVector` is likely much more memory-efficient than `deque` (or `queue`). Though the traversal is not a pure BFS, it's still valid.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+  auto isConstantValue = [&](Value value) {
+    // TODO: integers/vectors that might be pointers
+    if (isa<LLVM::LLVMPointerType, MemRefType>(value.getType())) {
+      assert(returnOps.size() == 1);
+      auto *fma = solver.lookupState<ForwardMemoryActivity>(*returnOps.begin());
+      auto *bma = solver.lookupState<BackwardMemoryActivity>(
+          &callee.getFunctionBody().front().front());
+
+      auto *pointsToSets =
+          solver.lookupState<enzyme::PointsToSets>(*returnOps.begin());
```
ftynse, line 747: (*) why is this looking at points-to at the exit point of the function? Can't there be a situation where the value is known to _not yet_ point to some class and we are being excessively conservative here? Is there a case where that wouldn't be correct?

(Also, we'd need to handle the multi-return case correctly by joining the states at all return points).

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+        }
+      }
+      // Otherwise, it's constant
+      return true;
+    }
+
+    return !isActiveData(value);
+  };
+
+  function_ref<bool(Operation *)> isConstantInstruction = [&](Operation *op) {
```
ftynse, line 783: (+) `function_ref` is a non-owning reference to a function. You most likely want an `std::function` here so the value is not a reference to a temporary.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+}
+
+void enzyme::runDataFlowActivityAnalysis(
+    FunctionOpInterface callee, ArrayRef<enzyme::Activity> argumentActivity,
+    bool print, bool verbose, bool annotate) {
+  SymbolTableCollection symbolTable;
+  DataFlowSolver solver;
+
+  solver.load<enzyme::PointsToPointerAnalysis>();
+  solver.load<enzyme::AliasAnalysis>(callee.getContext());
```
ftynse, line 928: (*) do we actually need these while computing the activity analyses, or can we finish them entirely beforehand? If so, we could abstract away the queries and compare the time it takes to run it in two stages vs. one stage.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.h
```diff
@@ -0,0 +1,47 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// If using this code in an academic setting, please cite the following:
+// @incollection{enzymeNeurips,
+// title = {Instead of Rewriting Foreign Code for Machine Learning,
+//          Automatically Synthesize Fast Gradients},
+// author = {Moses, William S. and Churavy, Valentin},
+// booktitle = {Advances in Neural Information Processing Systems 33},
+// year = {2020},
+// note = {To appear in},
```
ftynse, line 16: I'm quite sure it has already appeared in 2020. :)

enzyme/Enzyme/MLIR/Passes/PrintActivityAnalysis.cpp
```diff
@@ -0,0 +1,202 @@
...
+        argActivities[idx] = enzyme::Activity::enzyme_const;
+      else if (isa<FloatType, ComplexType>(argType))
+        argActivities[idx] = enzyme::Activity::enzyme_out;
+      else if (isa<LLVM::LLVMPointerType, MemRefType>(argType))
+        argActivities[idx] = enzyme::Activity::enzyme_dup;
+      else
+        argActivities[idx] = enzyme::Activity::enzyme_const;
+    }
+
+    for (const auto &[idx, resType] :
```
ftynse, line 55: Nit: prefer `auto &&` for `enumerate`.

enzyme/Enzyme/MLIR/Passes/PrintAliasAnalysis.cpp
```diff
@@ -0,0 +1,129 @@
...
+            taggedPointers.push_back({op->getAttr("tag"), result});
+          }
+        }
+      });
+    });
+
+    for (const auto &[tag, value] : taggedPointers) {
+      const auto *state = solver.lookupState<enzyme::AliasClassLattice>(value);
+      if (state) {
+        os << "tag " << tag << " " << *state << "\n";
```
ftynse, line 64: Given that `state` is a dense map (hashtable), this output does not have a stable order and cannot be reliably tested...

enzyme/Enzyme/MLIR/Passes/PrintAliasAnalysis.cpp
```diff
@@ -0,0 +1,129 @@
...
+        for (OpResult result : op->getResults()) {
+          auto *state = solver.lookupState<enzyme::AliasClassLattice>(result);
+          if (state) {
+            if (state->isUnknown()) {
+              op->setAttr("ac",
+                          StringAttr::get(result.getContext(), "<unknown>"));
+            } else {
+              for (auto aliasClass : state->getAliasClasses()) {
+                op->setAttr("ac", aliasClass);
+              }
```
ftynse, line 113: This overwrites the attribute for every new result, should use an array attribute instead with as many entries as the op has results.

ftynse (commented):
Please _answer_ the questions marked with (*) and address the requests marked with (+). The rest of the comments are more of a TODOs, many of which are already implemented in follow-ups.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
pengmai: Activity analysis never feeds into alias analysis, so we can finish them entirely beforehand.

> we could abstract away the queries and compare the time it takes to run it in two stages vs. one stage.

Completely agree!

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+
+/// From LLVM Enzyme's activity analysis, there are four activity states.
+// constant instruction vs constant value, a value/instruction (one and the same
+// in LLVM) can be a constant instruction but active value, active instruction
+// but constant value, or active/constant both.
+
+// The result of activity states are potentially different for multiple
+// enzyme.autodiff calls.
+
+enum class ActivityKind { Constant, ActiveVal, Unknown };
```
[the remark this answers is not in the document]
pengmai: I'm not sure that there's a meaningful difference between an optimistic initialization to constant and explicitly processing a value as constant. I'm certainly not opposed to an `Undefined` state, I just don't immediately see the benefit.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -0,0 +1,976 @@
...
+
+  void visitOperation(Operation *op,
+                      ArrayRef<const ForwardValueActivity *> operands,
+                      ArrayRef<ForwardValueActivity *> results) override {
+    if (op->hasTrait<OpTrait::ConstantLike>())
+      return;
+
+    // Bail out if this op affects memory.
+    if (!isPure(op))
+      return;
```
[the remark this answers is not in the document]
pengmai: The reason is that the analysis assume ops that affect memory will be processed by the dense analysis, not the sparse one. The sparse one is really only meant for ops that propagate activity via the creation of active data from active data operands, such as `arith.addf`, `math.tanh`, etc.

Notably, load ops should be reasoned about by the dense analysis. I'm in favour of finding better ways to represent these; interface methods will likely help to make this more explicit.

enzyme/Enzyme/MLIR/Analysis/AliasAnalysis.cpp
```diff
@@ -0,0 +1,800 @@
...
+
+// The LLVM dialect is only used for attribute names.
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+
+using namespace mlir;
+using namespace mlir::dataflow;
+
+static bool isPointerLike(Type type) {
+  return isa<MemRefType, LLVM::LLVMPointerType>(type);
+}
```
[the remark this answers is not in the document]
wsmoses: Haven't looked fully at users but would this semantically be better as "is mutable"

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
pengmai: I think it's definitely a potential way to improve the precision of the results, but looking at intermediate steps requires some thought *at what point* the activity of a value is important. I'll think through this further.

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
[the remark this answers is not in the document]
ftynse: Thinking about the `setToEntryState` mess... Are mlir analysis supposed to be optimistic?

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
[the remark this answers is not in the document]
ftynse: Yeah, in the meantime, a comment saying "ops with memory effects will be processed by the dense analysis"  would make it much clearer than "bail".

Theoretically, a sparse analysis is just an abstraction over a dense analysis where the lattice maps values to "sparse lattice" instances. We could remove that abstraction and have a dense analysis with an explicit map in the lattice (though not very efficient for storage).

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
ftynse: Ultimately, activity analysis is used to identify which operations are active. I suppose it is sufficient to reason about activity of values, which depends on points-to, _around_ that operation (before or after depending on the flow of activity) when deciding whether the operation is active.

ftynse: Concretely, I'm thinking about something like but with extra level of indirection

```mlir
%0 = ... // points to active value
store %0, %p  // %p transitively points to active value here
%a = load %p // %a points to active value (trivially)
// do something with %p
%1 = ... // points to constant values only
store %1, %p  // %p does *not* transitively point to active value here
%2 = load %p  // %2 points to constant values only trivially
// do something else with %p
```

if we are looking at the end of the function, and since we currently treat all stores as "may store" and join the points-to lattices, the return state will be that `%p` may be pointing to either constant or active pointer.

ftynse (approved):

enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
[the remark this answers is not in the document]
pengmai: Concretely, I know liveness analysis, which has an implementation upstream, [is initialized optimistically assuming all values are dead.](https://github.com/llvm/llvm-project/blob/4233176f9bf8d0309c03b127aa113ae826718dfa/mlir/include/mlir/Analysis/DataFlow/LivenessAnalysis.h#L66C3-L69C23)

Thinking about it more, I may have slightly misinterpreted the role of `setToEntryState`. It seems that its role is to *set to the pessimistic state* ($\top$), and the framework assumes that public function arguments should have this pessimistic state. This assumption isn't necessarily true for activity analysis because the activity of top-level function arguments should be explicitly provided by the user, and I believe most AD tools will take these user annotations as the truth even if dataflow can't be reasoned about at all.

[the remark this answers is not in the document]
ftynse: Basically, I have been trying (and so far mostly failing) to classify if the various analysis structures in terms of the following terminology. Optimistic analyses initialize lattice instances as the pessimistic fixpoint ($\top$) and transfer function may meet ($\cap$) lattice instances and progress towards optimistic fixpoint ($\bot$). Pessimistic analyses initialize lattice instances as the optimistic fixpoint ($\bot$) and progress towards $\top$ using join ($\cup$).

In the framework, forward analyses use join and backward analyses use meet, but I don't believe forward classes are meant for pessimistic and backward classes are meant for optimistic analyses. This may be a misinterpretation of moving "forward" (towards $\top$) or "backward" (towards $\bot$) along the lattice in the original implementation. I *think* forward and backward should only relate to the direction in which state is propagated, and both should be compatible with optimistic and pessimistic analyses. In fact, the `setToEntry/ExitState` should be more of a `setToFixPointState`. (Technically, having either `join` or `meet` should be sufficient since we cannot reuse the same lattice in two analyses, the lattice class is used as identifier of the analysis and `join`/`meet` are dual).

In practice, I see `setToEntry/ExitState` being used _in the framework_ for unanalyzable control flow situations: when we don't know all callsites of a function (e.g. the function is public and may be called from another translation unit), or when we don't know block or region predecessors/successors for forward/backward analysis respectively. These are the kind of irrecoverable situations where we would want to set the lattice instances to some edge state. I'm not sure which one... One interpretation is that in _both_ optimistic and pessimistic we could set the lattice to the $\top$ state, because it is always a conservatively correct overapproximation of the value. Optimistic analysis will be able to refine the approximation by approaching $\bottom$, but pessimistic analysis will be not be. Any thoughts on this? We are clearly missing one or two bits.

For initialization specifically, we may want to provide overrides for annotated functions and potentially mechanism to further customize other cases of control flow that are not analyzable through interfaces but may be specified by the user for specific analyses. That may be enough for activity/aliasing, but I'd really like to get to the bottom of the optimistic/pessimistic fixpoint story (pun intended).

PR #1552, opened 2023-11-16
reduce nesting depth

tgymnich:
attempted to reduce nesting depth a bit.


pengmai:
Thanks for doing this @tgymnich! I agree that the code organization should be cleaned up, I just want to also make sure that it remains easily extensible to add new interfaces. I'm planning on doing a comb-through to see which places can be cleaned up this way without hurting that extensibility.

pengmai:
As an update, there's a lot of in-flight work happening in branches floating around and @ftynse has requested that we save code clean up until after #1549 is merged to simplify the necessary rebasing. I'm planning on revisiting cleanup after we get those branches merged.

PR #1556, opened 2023-11-17
[mlir] Update Enzyme MLIR pieces to LLVM 18

pengmai:
- Clean up warnings, clang-format
- Remove unused/outdated lower to llvm pass

This PR is an effort to chunk up #1549 into more manageable pieces.


enzyme/Enzyme/MLIR/enzymemlir-opt.cpp
```diff
@@ -92,7 +94,7 @@ int main(int argc, char **argv) {
         PtrElementModel<LLVM::LLVMPointerType>>(*ctx);
     LLVM::LLVMArrayType::attachInterface<PtrElementModel<LLVM::LLVMArrayType>>(
         *ctx);
-    enzyme::CacheType::attachInterface<MemRefInsider>(*ctx);
+    // enzyme::CacheType::attachInterface<MemRefInsider>(*ctx);
```
ftynse, line 97: If we no longer need cache type inside memref, let's just delete this line.

enzyme/Enzyme/MLIR/Dialect/Ops.h
```diff
@@ -16,7 +16,7 @@
 #include "mlir/Interfaces/ControlFlowInterfaces.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 
-#include "mlir/IR/DialectImplementation.h"
+#include "mlir/Bytecode/BytecodeOpInterface.h"
```
ftynse, line 19: Are we using bytecode interface somewhere?

ftynse (approved):

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
wsmoses, line 1: This should probably be shadowType = gutils->getShadowType(…)

tgymnich (approved):

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -89,10 +88,10 @@ struct ForOpInterface
       if (!gutils->isConstantValue(std::get<0>(r)))
         nYields.push_back(gutils->invertPointerM(std::get<1>(r), builder));
     }
-    repFor.getBody()->push_back(
-        oldYield->create(oldYield->getLoc(), oldYield->getName(), TypeRange(),
-                         nYields, oldYield->getAttrs(),
-                         oldYield->getSuccessors(), oldYield->getNumRegions()));
+    repFor.getBody()->push_back(oldYield->create(
```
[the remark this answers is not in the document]
pengmai: I think `clone` followed by `setOperands` is clearer in its intentions. I'm happy to change it to that for these parts.

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
pengmai: The MLIR gradient utils doesn't appear to have a `getShadowType` method, I'll need to either add one or do `shadowType = cast<AutoDiffTypeInterface>(...).getShadowType()`

enzyme/Enzyme/MLIR/Dialect/Ops.h
pengmai: My understanding (correct me if I'm wrong) is that this header is necessary to get the code to compile. It's in the [standalone dialect example](https://github.com/llvm/llvm-project/blob/main/mlir/examples/standalone/include/Standalone/StandaloneDialect.h) and it looks like the tablegen header files include references to `::mlir::DialectBytecodeWriter`, so compilation fails without it.

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
wsmoses: Ah fair, yeah we can do that later then.

wsmoses (approved):
Feel free to merge whenever you're ready!

enzyme/Enzyme/MLIR/Dialect/Ops.h
ftynse: Have you tried removing it and seeing what happens? (If it is indeed implied by the bare structure, we should make it much more obvious in the documentation).

pengmai: Yup, removing it causes compilation to fail. Are you referring to the MLIR documentation or having a comment here?

ftynse: MLIR documentation, but I don't necessarily expect _you_ to add that or make it a precondition for this to land. I'll investigate once this lands.

PR #1567, opened 2023-12-04
Use early_inc_range

tgymnich:
Get rid of quite a few allocations by using `make_early_inc_range`

### DIFF
--- a/enzyme/Enzyme/TypeAnalysis/TypeTree.h
+++ b/enzyme/Enzyme/TypeAnalysis/TypeTree.h
@@ -191,8 +191,7 @@ public:
     // Check if there is an existing match, e.g. [-1, -1, -1] and inserting
     // [-1, 8, -1]
     {
-      std::set<std::vector<int>> toremove;
-      for (const auto &pair : mapping) {
+      for (const auto &pair : llvm::make_early_inc_range(mapping)) {
         if (pair.first.size() == SeqSize) {
           // Whether the the inserted val (e.g. [-1, 0] or [0, 0]) is at least
           // as general as the existing map val (e.g. [0, 0]).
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/TypeAnalysis/TypeTree.h
```diff
@@ -248,7 +247,8 @@ class TypeTree : public std::enable_shared_from_this<TypeTree> {
             if (CT == BaseType::Anything || CT == pair.second) {
               // previous equivalent values or values overwritten by
               // an anything are removed
-              toremove.insert(pair.first);
+              changed = true;
```
wsmoses, line 250: Is this legal here (std map iirc), for some std containers I remember erasure invalidates all iterators.

wsmoses:
This is fantastic and should be done throughout the codebase. However, I don't remember if the datastructre here (sstd::map) would support this.

enzyme/Enzyme/TypeAnalysis/TypeTree.h
tgymnich: > References and iterators to the erased elements are invalidated. Other references and iterators are not affected.
https://en.cppreference.com/w/cpp/container/map/erase

wsmoses:
References and iterators to the erased elements are invalidated. Other references and iterators are not affected.

from `https://en.cppreference.com/w/cpp/container/map/erase`

So looks like this should be fine

wsmoses (approved):

tgymnich:
> This is fantastic and should be done throughout the codebase. However, I don't remember if the datastructre here (sstd::map) would support this.
@wsmoses I've got some more of this lined up. Just testing the waters with this PR.

PR #1570, opened 2023-12-07
[mlir] don't incorrectly bail from alias analysis

ftynse:
* Alias transfer behavior isn't always completely described by memory
  effects, be conservative about that.
* "Global" effects with no target value should not be ignored.
* Interpret `readnone` as "not reading or writing" rather than just "not
  reading", in line with LLVM.
* Simplify code in external callee attribute-based transfer function.
* Treat all stores as "may" because to meet the expectations of the
  activity post-analysis that only looks at function exit points and
  therefore needs conservative results.


wsmoses (approved):

wsmoses:
@ftynse CI is now running and the CI failures are actually real from the PR:
```
********************
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.. 
********************
Failed Tests (10):
  Enzyme :: MLIR/ActivityAnalysis/activereturn.mlir
  Enzyme :: MLIR/ActivityAnalysis/allocator.mlir
  Enzyme :: MLIR/ActivityAnalysis/allocret.mlir
  Enzyme :: MLIR/ActivityAnalysis/callgraph.mlir
  Enzyme :: MLIR/ActivityAnalysis/memalgn.mlir
  Enzyme :: MLIR/ActivityAnalysis/multiptr.mlir
  Enzyme :: MLIR/ActivityAnalysis/string.mlir
  Enzyme :: MLIR/ActivityAnalysis/subld.mlir
  Enzyme :: MLIR/AliasAnalysis/escaping_memory.mlir
  Enzyme :: MLIR/AliasAnalysis/llvm_casts.mlir
  ```

wsmoses:
Hm looks like those errors already exist in main, regardless

PR #1571, opened 2023-12-07
[mlir] correctly handle "unknown" state in activity analysis

ftynse:
This requires injecting ModRef information about library functions.

Depends on #1570.


wsmoses:
This reduces the number of failed tests on main from 10 to 5 so that's progress.

wsmoses (approved):

PR #1572, opened 2023-12-07
[mlir] proper "undefined" state in pointer analysis and correct handling of fresh state

ftynse:
Historically, `AliasClassSet` has been using empty set as the bottom
state of the join-semilattice. This made it impossible to differentiate
between "not yet analyzed" and "known not to alias with anything"
states. While not necessarily problematic for the naive alias analysis,
this is crucial for points-to-pointer analysis that was forced to become
unnecessarily conservative by treating "not yet analyzed" points-to
state as "points to unknown" (top sate of the join-semilattice). Worse,
some transfer functions weren't trivially monotonous because of that.

Introduce a new "undefined" set to `AliasClassSet` and lattice. Update
PointsToLattice to explicitly store the cases when a known alias class
points to unknown alias class and interpret unknown (not yet listed)
alias class as pointing to undefined alias class. Make all transfer
functions monotonous by only processing known points-to entries. Unknown
points-to classes can be treated by the client as potentially pointing
to unknown, but so far this wasn't needed as, at the fixpoint state, all
available classes should have been analyzed. Assertions are put in place
in activity analysis to catch undefined classes.

Additionally, make it more difficult to move the join-semilattice to a
previous (<= current) value by exposing mostly the APIs that can only
advance the lattice. This is not yet complete due to partially incorrect
handling of distinct identifiers that get recreated every time in
transfer functions.


enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
```diff
@@ -835,27 +834,39 @@ void printActivityAnalysisResults(const DataFlowSolver &solver,
       // Traverse the points-to sets in a simple BFS
       std::deque<DistinctAttr> frontier;
       DenseSet<DistinctAttr> visited;
-      auto scheduleVisit = [&](auto range) {
-        for (DistinctAttr neighbor : range) {
+      // auto scheduleVisit = [&](auto range) {
```
wsmoses, line 837: dead code?

Issue #1600, opened 2024-01-03
Support for x86_fp80 (long double)


maxaehle:
I'm wondering if Enzyme can handle the 80-bit x87 extended precision floating-point format known as `x86_fp80` in LLVM IR, which is used by Clang (and GCC) to represent the C/C++ type `long double`.

From https://fwd.gymni.ch/tpmWJm, I presume that it is not supported right now.

Support for `x86_fp80` could be desirable at some point in the future, because even if users avoid `long double` in their C++ source code, this type can appear in C++ standard library headers included from their source code. For example, libstdc++ uses `long double` to implement [`std::generate_canonical`](https://en.cppreference.com/w/cpp/numeric/random/generate_canonical), if I understand [this code](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libstdc%2B%2B-v3/include/bits/random.tcc;h=7f4bf5ea1836e6540aa92f38ca6562bef41f6e1d;hb=HEAD#l3354) correctly. I suspect that this is also the reason why https://fwd.gymni.ch/HHKrZj fails.

wsmoses:
Yeah this would be good to add (and test). Would you be interested in
looking into this?

The location that Type Analysis assertion is thrown would be a good start.

maxaehle:
Yes, I'll look into this.

PR #1604, opened 2024-01-04
Improve support for x86_fp80 (long double) 

maxaehle:
This is an attempt to increase Enzyme's coverage of the LLVM type `x86_fp80`, in response to #1600. I'm not at all familiar with Enzyme from a developer perspective, and very happy about any kind of input.

## Existing coverage
Some `x86_fp80` coverage is already there, e.g. because the differentiation of [elementary operations](https://github.com/EnzymeAD/Enzyme/blob/a6e8dc066ee9834f19e9d163604ca8d29ec366c5/enzyme/Enzyme/InstructionDerivatives.td#L866-L893) and [floating-point conversions](https://github.com/EnzymeAD/Enzyme/blob/a6e8dc066ee9834f19e9d163604ca8d29ec366c5/enzyme/Enzyme/AdjointGenerator.h#L1427-L1429C17) is defined independent of the type (if my understanding is correct).

Hence, the new division and sine testcases suggested in this PR pass from the beginning.

## Missing coverage

A few locations in the code where `x86_fp80` support is missing can be identified by searching for `isDoubleTy` and `isFloatTy` in the code:
- Handling of **bit-tricks** (where real arithmetic is performed using e.g. bitwise logical operations), e.g. in [AdjointGenerator.h](https://github.com/EnzymeAD/Enzyme/blob/a6e8dc066ee9834f19e9d163604ca8d29ec366c5/enzyme/Enzyme/AdjointGenerator.h#L2717) and [TypeAnalysis.cpp](https://github.com/EnzymeAD/Enzyme/blob/a6e8dc066ee9834f19e9d163604ca8d29ec366c5/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L2965). 
- A particular case distinction that appears four times in **TypeTree.h**. At [one of them](https://github.com/EnzymeAD/Enzyme/blob/a6e8dc066ee9834f19e9d163604ca8d29ec366c5/enzyme/Enzyme/TypeAnalysis/TypeTree.h#L811) the 80-bit case is already considered, at the other three it is not.
- There are **some more places**, related to e.g. conversion from floating-point type to integer type of the same size, obtaining the type as a string, etc. 

I'm not sure whether additional work is necessary at other places. Please post if you are aware of some. Eventually, tests should also tell us.


maxaehle:
Commit [b01d87f](https://github.com/EnzymeAD/Enzyme/pull/1604/commits/b01d87f8833de6b9d46c4cebc0c7a4ee30ccab54) adds `x86_fp80` support in TypeTree.h and at the "some more places" mentioned above. With these changes, `opt` can successfully form differentiated code for the new `fp80` testcases (before [b01d87f](https://github.com/EnzymeAD/Enzyme/pull/1604/commits/b01d87f8833de6b9d46c4cebc0c7a4ee30ccab54), `opt` would crash as reported in #1600)). [895ed2e](https://github.com/EnzymeAD/Enzyme/pull/1604/commits/895ed2e86a303abe68ee7d5e7383d59746ee6481) updates the expected output of the `opt` run. 

Next, I'll think about extending the bit-trick handling.

enzyme/Enzyme/TypeAnalysis/TypeTree.h
```diff
@@ -650,6 +652,8 @@ class TypeTree : public std::enable_shared_from_this<TypeTree> {
                 chunk = 8;
               } else if (flt->isHalfTy()) {
                 chunk = 2;
+              } else if (flt->isX86_FP80Ty()) {
```
wsmoses, line 655: We might as well here (within isFloat) simplify this to `chunk = dl.getTypeSizeInBits(flt) / 8;`

enzyme/Enzyme/TypeAnalysis/TypeTree.h
```diff
@@ -731,6 +735,8 @@ class TypeTree : public std::enable_shared_from_this<TypeTree> {
         chunk = 8;
       } else if (flt->isHalfTy()) {
         chunk = 2;
+      } else if (flt->isX86_FP80Ty()) {
```
wsmoses, line 738: Same comment here about `chunk = dl.getTypeSizeInBits(flt) / 8;`

enzyme/Enzyme/TypeAnalysis/TypeTree.h
```diff
@@ -554,6 +554,8 @@ class TypeTree : public std::enable_shared_from_this<TypeTree> {
                 chunk = 8;
               } else if (flt->isHalfTy()) {
                 chunk = 2;
+              } else if (flt->isX86_FP80Ty()) {
```
wsmoses, line 557: Also here

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -5181,6 +5181,8 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
                 chunk = 8;
               } else if (flt->isHalfTy()) {
                 chunk = 2;
+              } else if (flt->isX86_FP80Ty()) {
```
wsmoses, line 5184: and here

maxaehle:
Does anyone know what the purpose of [this piece of code in AdjointGenerator.h](https://github.com/EnzymeAD/Enzyme/blob/a6e8dc066ee9834f19e9d163604ca8d29ec366c5/enzyme/Enzyme/AdjointGenerator.h#L2613) is:
```c++
switch (BO.getOpcode()) {
    case Instruction::And: {
      // If & against 0b10000000000 and a float the result is 0
      
      // [...]
      auto FT = TR.query(&BO).IsAllFloat(size);
      auto eFT = FT;
      if (FT)
        for (int i = 0; i < 2; ++i) {
          auto CI = dyn_cast<ConstantInt>(BO.getOperand(i));
          if (CI && dl.getTypeSizeInBits(eFT) ==
                        dl.getTypeSizeInBits(CI->getType())) {
            if (eFT->isDoubleTy() && CI->getValue() == -134217728) {
              setDiffe(&BO, Constant::getNullValue(diffTy), Builder2);
              // Derivative is zero (equivalent to rounding as just chopping off
              // bits of mantissa), no update
              return;
            }
          }
        }
```
Is some weird bit-trick handled here? How does the bit-trick work? 134217728 is 2**27, but I don't see right now how real arithmetic could be hidden here...

wsmoses:
From the comment above it looks like it is saying float & signbit -> 0 derivative.

Also note that it was checking against a negative number (which is twos complement has different bits set)

enzyme/Enzyme/TypeAnalysis/TypeTree.h
maxaehle: Thanks, changed in [24c8098](https://github.com/EnzymeAD/Enzyme/pull/1604/commits/24c8098f99d83af45a0d681a44d9ec9411d40be3)

enzyme/Enzyme/TypeAnalysis/TypeTree.h
maxaehle: also changed in [24c8098](https://github.com/EnzymeAD/Enzyme/pull/1604/commits/24c8098f99d83af45a0d681a44d9ec9411d40be3)

enzyme/Enzyme/GradientUtils.cpp
maxaehle: also changed in [24c8098](https://github.com/EnzymeAD/Enzyme/pull/1604/commits/24c8098f99d83af45a0d681a44d9ec9411d40be3)

enzyme/Enzyme/TypeAnalysis/TypeTree.h
maxaehle: I couldn't simplify the code here like at the other places because there is no `DataLayout` instance (as far as I can see).

wsmoses: What would it look like to add an argument for a datalayout here?

wsmoses:
Could we also update the shift indices function that already manually handled fp80 to use the datalayout size check as well?

enzyme/Enzyme/TypeAnalysis/TypeTree.h
maxaehle: In every place where `IsAllFloat` is called, there is a DataLayout instance lying around. In 6a707777e364fd060d6f21da78f0ed371322c7f5, I add a DataLayout argument to the signature of IsAllFloat, in order to simplify its implementation in the same way as in the other places. (Could you please take another look and confirm that this change is correct?)

maxaehle:
> Could we also update the shift indices function that already manually handled fp80 to use the datalayout size check as well?

Yes, I also changed it in [24c8098](https://github.com/EnzymeAD/Enzyme/pull/1604/commits/24c8098f99d83af45a0d681a44d9ec9411d40be3).

maxaehle:
I can now compile and run the other example from #1600, https://fwd.gymni.ch/HHKrZj, on my local system. The `Integration CI LLVM 11 Release ubuntu-20.04` run is marked as "failed" because three testcases unexpectedly pass now. They all use the C++ header `<random>` in a similar way (see e.g. [here](https://github.com/EnzymeAD/Enzyme/blob/5422797090b89c1e22e836eb74a852de544febc1/enzyme/test/Integration/Sparse/ringspring3Drestlengthone.cpp#L141)) as this example, so probably that's a good message.

wsmoses (approved):

wsmoses:
If you're up for it, it may also be useful to test fp128 in a follow up PR.

maxaehle:
Thanks! I might look at bit-tricks and/or fp128 in a separate PR in the next days.

PR #1605, opened 2024-01-05
Support more C math.h functions

maxaehle:
As a follow-up to #1604, try to support more C math functions from here:

https://en.cppreference.com/w/c/numeric/math

Functions that were missing and work now:   
sinhl, coshl, tanhl, frexpl, modf, modff, modfl, fdim, fdimf, fdiml  
asinh asinhf asinhl, acosh acoshf acoshl; atanh atanhf atanhl; erff erfl; erfcf erfcl; nearbyint nearbyintf nearbyintl;


maxaehle:
So far I didn't add `tanhl` to the places where `tanh` and `tanhf` appear in GradientUtils.cpp, [here](https://github.com/EnzymeAD/Enzyme/blob/f92888173d4d4e328f7aa283cc8bf952d1279a26/enzyme/Enzyme/GradientUtils.cpp#L4012) and [here](https://github.com/EnzymeAD/Enzyme/blob/f92888173d4d4e328f7aa283cc8bf952d1279a26/enzyme/Enzyme/GradientUtils.cpp#L4162). I presume that having them there is irrelevant because [`isMemFreeLibMFunction`](https://github.com/EnzymeAD/Enzyme/blob/f92888173d4d4e328f7aa283cc8bf952d1279a26/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.h#L61) will anyhow return `true` if `n` is `"tanhl"`. 

Also, I think no changes are necessary in [PreserveNVVM.cpp](https://github.com/EnzymeAD/Enzyme/blob/f92888173d4d4e328f7aa283cc8bf952d1279a26/enzyme/Enzyme/PreserveNVVM.cpp#L313) as there are no NVVM bitcode functions for 80-bit floating-point arithmetic.

maxaehle:
`frexpl` was giving NaN derivatives before. The derivative computation for `frexp(double x)`  makes a copy of `x`, zeroes the sign and mantissa bits (which gives the largest power of two not exceeding `x` in terms of absolute values), and use the inverse of the double of that as the partial derivative of `frexp`. `x86_fp80` differs from `double` and `float` in that the format has no implicit leading mantissa digit 1. Therefore, we should not zero the most-significant mantissa bit when differentiating `frexpl`.

I guess that the entire bit-trick does not work for sub-normal numbers (because it uses 1./(2.*0.0)=NaN  as the partial derivative). The present set of changes does not fix this.

wsmoses:
@maxaehle any chance you can revisit this?

maxaehle:
Thanks for the reminder! I've added a few more functions, and will try to create another test case tomorrow to see whether the math library is now entirely covered.

maxaehle:
I have no idea why [2d1cdef..85391ce](https://github.com/maxaehle/Enzyme/compare/2d1cdef..85391ce) breaks the Julia CI tests. I'm trying to investigate, but unfortunately I cannot make a development build of Enzyme.jl at the moment (EnzymeAD/Enzyme.jl#1291).

maxaehle:
Ok, then this PR would now be ready for review. A few items that could be done in a later PR:
- Add derivatives of tgamma and lgamma variants
- Add a test that checks correct handling of all math.h functions (and explicitly enumerates those that don't work yet)
- 128-bit floating-point arithmetic (quadmath.h)

enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -350,6 +366,33 @@ def : CallPattern<(Op $x),
...
                   >;
+def : CallPattern<(Op $x),
+                  ["sinhl"],
+                  [(FMul (DiffeRet), (Call<(SameTypesFunc<"coshl">), [ReadNone,NoUnwind]> $x))],
+                  (ForwardFromSummedReverse),
+                  [ReadNone, NoUnwind]
+                  >;
+
+def : CallPattern<(Op $x),
+                  ["asinh", "asinhf", "asinhl", "__nv_asinh", "__nv_asinhf"],
```
wsmoses, line 377: can you add tests for asinh/acosh/atanh?

enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -839,6 +892,16 @@ def : IntrPattern<(Op $x, $y),
                   (Select (FCmpOLT $x, $y), (SelectIfActive $y, (Shadow $y), (Zero $y)), (SelectIfActive $x, (Shadow $x), (Zero $x)))
                   >;
 
+def : CallPattern<(Op $x, $y),
+                  ["fdim", "fdimf", "fdiml"],
+                  [
+                    (Select (FCmpOLT $x, $y), (ConstantFP<"0"> $x), (DiffeRet)),
```
wsmoses, line 898: This feels wrong?

if fdim(x, y) = abs|x-y| then shouldnt this be either differet or -differet?

maxaehle: [`fdim`](https://en.cppreference.com/w/cpp/numeric/math/fdim) is max(x-y,0), except for corner cases related to NaN.

enzyme/Enzyme/InstructionDerivatives.td
maxaehle: Right, [4001e41](https://github.com/EnzymeAD/Enzyme/pull/1605/commits/4001e41d9780681159cc473d638a97a11c659bde) adds tests for asinh, asinhf, asinhl, acosh, atanh in the forward and reverse mode.

There are also no tests so far for the error function and nearbyint. Is this ok or should I add tests for those?

wsmoses (approved):

PR #1614, opened 2024-01-17
make fpic work on windows

ZuseZ4:
Let's see if this passes CI. Credits to @OwenTrokeBillard


ZuseZ4:
looks like this works on both linux and Mac. Nice!

wsmoses (approved):

PR #1615, opened 2024-01-18
Add functions to truncate and expand fp values

ivanradanov:


enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -5206,10 +5226,52 @@ class TruncateGenerator : public llvm::InstVisitor<TruncateGenerator> {
...
+
+  if (fromwidth == towidth) {
+    context.req->eraseFromParent();
+    return true;
+  }
+
+  if (fromwidth < towidth) {
+    std::string s;
+    llvm::raw_string_ostream ss(s);
+    ss << "Cannot truncate into a large width\n";
```
wsmoses, line 5242: Can we change this into an EmitFailure (see above places)

wsmoses: oh jk

Issue #1621, opened 2024-01-22
Support for llvm.smin and llvm.smax


rmoyard:
I have some code that generates llvm including `llvm.smin` and `llvm.smax` and get the following error when running the Enzyme passes.
```
cannot handle (reverse) unknown intrinsic
llvm.smin
  %36 = tail call i64 @llvm.smin.i64(i64 %35, i64 3) #6
```
Are those operations supported by Enzyme? If not what is the best way to add them?

ZuseZ4:
Oh hi, someone else from Toronto!
Enzyme doesn't really support computing gradients of integers, but I assume you try to get them for some float based variable and then it's possible that our Analysis isn't strong enough here to understand what's really going on. You can try enzyme-loose-ta, see here: https://enzyme.mit.edu/getting_started/UsingEnzyme/#loose-type-analysis 
You might also need to add support in this style: https://github.com/EnzymeAD/Enzyme/blob/0d8c77cad4f0888d4dea4e9b32143c7332f65d5e/enzyme/Enzyme/AdjointGenerator.h#L2379

rmoyard:
Hi @ZuseZ4 nice to see another Torontonian here :+1: Thanks a lot for your answer!

This is correct I have some float based variable, I have this function (python) where we slice and sum. We (Catalyst) take the gradient and this where we use Enzyme.
```
import catalyst
import jax

def f(x):
    return jax.numpy.sum(x[::2])

x = jax.numpy.array([0.1, 0.2, 0.3, 0.4])
print(catalyst.qjit(catalyst.grad(f)(x))
```
it is lowered to MLIR and then to LLVM (we add `_enzyme_autodiff`), we apply O2 and then we apply the Enzyme passes. We end up with the LLVM (before Enzyme) for the original function. More details here https://gist.github.com/rmoyard/8402c4991eed5bced0c9a23cfaf8ecf3
```
define internal void @f.cloned(ptr nocapture readnone %0, ptr nocapture readonly %1, i64 %2, i64 %3, i64 %4, ptr nocapture readnone %5, ptr nocapture writeonly %6, i64 %7) {
.critedge:
  ...
  %46 = tail call i32 @llvm.smax.i32(i32 %45, i32 0)
  %47 = zext nneg i32 %46 to i64
  %48 = tail call i64 @llvm.smin.i64(i64 %47, i64 3)
  %49 = getelementptr double, ptr %1, i64 %48
  %50 = load double, ptr %49, align 8
[...]

PR #1634, opened 2024-01-25
Add smax, smin, umin, umax to type analysis

rmoyard:
This PR adds, smin, smax, umin, umax llvm intrinsic to the type analysis. This allows the activity to consider them as inactive and instead of active as before. This unlocks programs using those 4 functions to be differentiate.

Closes https://github.com/EnzymeAD/Enzyme/issues/1621


enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -3861,6 +3861,18 @@ void TypeAnalyzer::visitIntrinsicInst(llvm::IntrinsicInst &I) {
         &I);
     return;
 
+  case Intrinsic::smax:
+  case Intrinsic::smin:
+  case Intrinsic::umax:
+  case Intrinsic::umin:
+    // No direction check as always valid
```
wsmoses, line 3868: I think this may be a bit too aggressive. 

E.g. I think here we would want:
  min(int, int) -> int
  min(ptr, ptr) -> ptr
 
 int return -> two args are int's
 ptr return -> two args are ptr's

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -3861,6 +3861,18 @@ void TypeAnalyzer::visitIntrinsicInst(llvm::IntrinsicInst &I) {
         &I);
     return;
 
+  case Intrinsic::smax:
```
wsmoses, line 3864: You'll also need to version gate this, as smax/etc didn't exist in older LLVM's

rmoyard: Do you know what version are we talking about? 16 maybe?

enzyme/test/TypeAnalysis/smax.ll
```diff
@@ -0,0 +1,35 @@
+; RUN: if [ %llvmver -lt 16 ]; then %opt < %s %loadEnzyme -opaque-pointers -print-type-analysis -type-analysis-func=smax -o /dev/null | FileCheck %s; fi
```
rmoyard, line 1: @wsmoses How do you usually not run test for older versions?

wsmoses: So this test shouldn't need a ptr regardles.

E.g. a minimal test would be:


```
define i64 f(i64 a0, i64 b0) {
   c0 = call i53 llvm.smax(a, b)
   a = call i64 @run()
   b = call i64 @run()
   c = call i53 llvm.smax(a, b)
   ret c;
}
```

Arguments and returns are given reasonable types (aka initialied to int here). This the c0 call will test that we forward propagate from args to return, and the a/b/c test will test that we backwards propagate from return to args.

In any case the way we version gate is the `if [ %llvmver -lt 16 ]; then` you can see in the run file. That is a if version < 16 use old pass manager tests, and the line below (presently unconditionally) runs new pass manager versions

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
wsmoses: Not sure, you can check the CI logs, Enzyme-level tests do 11-16, integration runs through 17, and bazel/mlir does nightly-ish

rmoyard: https://github.com/llvm/llvm-project/commit/deb4bb2b3abd216aceaa109a5304f025a234540f

enzyme/test/TypeAnalysis/smax.ll
```diff
@@ -0,0 +1,24 @@
+; RUN: if [ %llvmver -lt 16 ]; then %opt < %s %loadEnzyme -print-type-analysis -type-analysis-func=smax -o /dev/null | FileCheck %s; fi
+; RUN: %opt < %s %newLoadEnzyme -passes="print-type-analysis" -type-analysis-func=smax -S -o /dev/null | FileCheck %s
+
+define i32 @smax(i32 %a, i32 %b) {
+entry:
+  %0 = call i32 @llvm.smax.i32(i32 %a, i32 %b)
+  %1 = add i32 %a, 1
```
wsmoses, line 7: You need this to be an unknown value (eg from a random call), since otherwise it will already be known as an int from the argument.

Otherwise this won't actually test the propagation

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
rmoyard: Thanks for the feedback, I have updated it. Let me know what you think. I am not sure if this part 

```
int return -> two args are int's
ptr return -> two args are ptr's
```

is covered by the changes.

wsmoses: Not quite, you'd need to do something like

if (directions & up) {
  Auto retty = getAnalysis(I)[{-1}];
  if (retty == BaseType::Integer or Pointer) {
    Update analysis of op 0/1 with  that type tree
  }
}

if (directions & down) {
  Auto op0 = getAnalysis of operand 0, [{-1}]
  And op1
  If these are both int or both pointer
    update return analysis with corresponding Type tree
}

rmoyard: Thanks for the help, I think understand a bit better

rmoyard: I have updated the code as suggested, but it does not help my issue anymore. I have checked with my .ll file and the types are the following: (you see that they are never two pointers or two integers)

```
UP Unknown
DOWN Unknown Anything
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWNI nteger Anything
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Anything
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Integer Anything
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
UP Unknown
DOWN Unknown Integer
```
Does it make sense to you if I add for DOWN that if any are integer then we propagate the integer?

wsmoses: Does it suffice if we have integer  & (int or anything) -> int

And to test this we would copy the same two min/max except change one input argument to be 0.

rmoyard: Unfortunately (int or anything) does not work for my case. Is it not fair to assume that if we have `InterType` on the first operand then necessarily we have another `InterType` on the second operand?

wsmoses: I'd like to see the case where it fails with the int / anything first. Can you restore the int or anything functionality, and add the zero test. Then we can either add a second PR or second test for the too-conservative handling of that.

I worry that Int with unknown may be wrong/overly aggressive so want to be careful here.

wsmoses: As an example, one side effect of int / unknown, means that the c = min(a : int, b) will always cause b to be an int. That may be valid, but I'm not sure. E.g. c will be set to an int since a is an int, then we can propagate up to b via c.

This may be problematic, for example if you have `min(16 : int, ptr-type)` which then will try to propagate int to the ptr and ptr to the int.

rmoyard: Ok let me go back to anything and add the test!

enzyme/test/TypeAnalysis/smaxsmin.ll
```diff
@@ -0,0 +1,29 @@
...
+  %6 = load i32, ptr %5, align 64
+  %7 = tail call i32 @llvm.smax.i32(i32 %6, i32 0)
+  %8 = zext nneg i32 %7 to i64
+  %9 = tail call i64 @llvm.smin.i64(i64 %8, i64 3)
+  %10 = getelementptr double, ptr %0, i64 %9
+  %11 = load double, ptr %10, align 8
+  ret void
+}
+
+; CHECK-NOT: %7 = tail call i32 @llvm.smax.i32(i32 %6, i32 0): {}
```
rmoyard, line 28: @wsmoses Here is the test not passing, let me know if you see anything that could help.

wsmoses: Yeah from your snippet here I actually agree with Type analysis that there isn't enough info to ascertain the types.

There are several potential ways to resolve this, but it would probably be easiest to see the MLIR which is lowered to figure out which.

For example presently the arg ptr %0 has no information about what it is [eg whether it is a pointer, int etc].

Since there is a gep %0, %9, that may be one good way to inject information. If it was a gep inbounds %0, %9 we can deduce that %9 must be an integer [and then with this PR it propagates upward perfectly]. However without the inbounds, it is possible that %0 is actually null and the pointer is %9 [which can come from an inttoptr]. This latter case actually is emitted by LLVM in practice, in particular for virtual C++ functions.

Another route may be attaching type information to _mlir_memref_to_llvm_alloca to say it is a memref of int, so all loads / stores are ints. Alternatively one can put TBAA on the load/store generated by the memref.

etc.

I'd say for now to move this test out of the PR into an issue, so we can get this PR merged, and then follow up for this next.

wsmoses: FYI, the results with GEP being marked inbounds:

```
wmoses@beast:~/git/Enzyme/enzyme/build15 (add_max_min_type_analysis) $  /home/wmoses/llvms/llvm15/buildD/./bin/opt < /home/wmoses/git/Enzyme/enzyme/test/TypeAnalysis/smaxsmin.ll  -load-pass-plugin=/home/wmoses/git/Enzyme/enzyme/build15/Enzyme/LLVMEnzyme-15.so -load=/home/wmoses/git/Enzyme/enzyme/build15/Enzyme/LLVMEnzyme-15.so --enzyme-attributor=0 -passes="print-type-analysis" -type-analysis-func=smaxsmin -S -o /dev/null
smaxsmin - {} |{[-1]:Pointer}:{} 
ptr %0: {[-1]:Pointer}
entry
  %1 = tail call ptr @_mlir_memref_to_llvm_alloc(i64 72): {}
  %2 = ptrtoint ptr %1 to i64: {}
  %3 = add i64 %2, 63: {}
  %4 = and i64 %3, -64: {[-1]:Pointer, [-1,0]:Integer, [-1,1]:Integer, [-1,2]:Integer, [-1,3]:Integer}
  %5 = inttoptr i64 %4 to ptr: {[-1]:Pointer, [-1,0]:Integer, [-1,1]:Integer, [-1,2]:Integer, [-1,3]:Integer}
  %6 = load i32, ptr %5, align 64: {[-1]:Integer}
  %7 = tail call i32 @llvm.smax.i32(i32 %6, i32 0): {[-1]:Integer}
  %8 = zext i32 %7 to i64: {[-1]:Integer}
  %9 = tail call i64 @llvm.smin.i64(i64 %8, i64 3): {[-1]:Integer}
  %10 = getelementptr inbounds double, ptr %0, i64 %9: {[-1]:Pointer}
  %11 = load double, ptr %10, align 8: {}
  ret void: {}
  ```

File: enzyme/test/TypeAnalysis/smax0.ll [hunk 6]
```diff
@@ -0,0 +1,23 @@
...
+  ret i32 %2
+}
+
+declare i32 @llvm.smax.i32(i32, i32)
+
+declare i32 @getint()
+
+; CHECK: smax0 - {[-1]:Integer} |{[-1]:Integer}:{} {[-1]:Integer}:{}
+; CHECK-NEXT: i32 %a: {[-1]:Integer}
+; CHECK_NEXT: i32 %b: {[-1]:Integer}
```
wsmoses, line 18: nit this should be a dash not underscore (you can run type analysis tests locally with `make -j check-typeanalysis`)

enzyme/test/TypeAnalysis/smax.ll
```diff
@@ -0,0 +1,26 @@
...
+}
+
+declare i32 @llvm.smax.i32(i32, i32)
+
+declare i32 @getint()
+
+
+; CHECK: smax - {[-1]:Integer} |{[-1]:Integer}:{} {[-1]:Integer}:{}
+; CHECK-NEXT: i32 %a: {[-1]:Integer}
+; CHECK_NEXT: i32 %b: {[-1]:Integer}
```
wsmoses, line 20: same dash comment

enzyme/test/TypeAnalysis/smaxsmin.ll
rmoyard: Thanks a lot @wsmoses for all the explanation!

wsmoses (approved):

PR #1636, opened 2024-01-26
[Truncate] Handle libm calls

ivanradanov:


wsmoses (approved):
Needs reformat but lgtm otherwise (but perhaps comment back in end of tesy?)

PR #1637, opened 2024-01-26
Allow disabling memmove warning

wsmoses:


ZuseZ4 (approved):
Also a way to solve this. ^^
I don't remember why this was so hard to fix, can you please write it down somewhere?

wsmoses:
I mean it is a warning that the implementation for aliasing pointers in memmove is not correct with this derivative (in a very specific case).

ZuseZ4:
Ah, it was aliasing. Fair, then I don't mind. #noalias

PR #1641, opened 2024-01-27
add compile time stresstest

ZuseZ4:
Based on some real-world example.
Only differentiating f_1 takes 5 sec, f1 and f2 takes 35 sec, f1-f5 take 120 sec.
My longer real-world testcase takes 6h in C++ and (as lower bound) 5 min in Rust (with seeded TA).
Jax Jit was worse than C++ iirc.

Run through:
`time /u/drehwald/prog/llvm16/build/bin/clang++ -O2 stress_5_500.cpp -fplugin=../build/Enzyme/ClangEnzyme-16.so`


wsmoses:
@ZuseZ4 does this need to be vectors or can it be reproduced with raw pointers?

ZuseZ4:
@wsmoses Interesting enough, ptrs are faster than vec (5*500 takes 80s with ptr, and 120s with &vec).
I therefore added both versions.

ZuseZ4:
I didn't add a generated .rs testcase to not blow up the codesize too much. 
Last time I tested you need ~60x more functions to get to the same compiletime.
However, I hope that TA improvements here not only benefit C++, but also Rust.

ZuseZ4:
Actually adding two (safe vs. uncheckd) rust versions.
It also allows to benchmark clang vs rustc without AD, rustc is surprisingly slow given that almost all time is spent in the backend/codegen (2.5s vs 4.8s).

wsmoses (changes requested):
So looking more closely at this PR I'm wondering if this would be better as an issue than a PR.

For example this isn't run in CI that would make sense to merge (nor should it be atm if it breaks CI with compile time).

Moreover keeping it as an issue is a reminder to fix it whereas merging doesn't have as issue to be tracked.

Thoughts?

ZuseZ4:
I don't mind too much either way. Having it merged is a bit nicer so I always have it at hand and can easier point people to it (e.g. I had a longer discussion on this in a rustc-dev channel about the compile times of this without Enzyme). But an issue works too

wsmoses:
@ZuseZ4 since I don't see a strong push for a PR rather than issue, I'm going to go ahead and close this and you should open an issue.

ZuseZ4:
@wsmoses Unchanged superlinear comptime increases. (Commenting out 994-1006)

wsmoses:
Can you put this in a currently open issue? If not this is lost

PR #1649, opened 2024-01-29
Reimplement MLIR tangent for scf.for generically

ftynse:
Introduce an interface for ops with region control flow and use that interface inside the tangent implementation for scf.for. This interface can also work for other kinds of control flow ops as long as they don't transfer control flow between regions (e.g. like scf.while does).

Further cleanup is necessary to clearly delegate implementation of AutoDiffOpInterface to this new interface for any compatible op via traits.


enzyme/test/MLIR/ForwardMode/while.mlir
```diff
@@ -0,0 +1,44 @@
...
+
+module {
+  func.func @while(%x : f64) -> f64 {
+    %cst = arith.constant 10.000000e+00 : f64
+    %c0 = arith.constant 0 : index
+    %c1 = arith.constant 1 : index
+    %c10 = arith.constant 10 : index
+
+    %r:2 = scf.while (%arg1 = %c0, %arg2 = %cst) : (index, f64) -> (index, f64) {
+      %1 = arith.cmpi slt, %arg1, %c10 : index
```
wsmoses, line 11: just to confirm, can we add a differentiable op here like arg2 = arg2 * 10

PR #1681, opened 2024-02-08
Remove unnecessary uses of templating

tgymnich:


wsmoses:
Ah this would be so nice. Fix format otherwise LGTM if all passes

tgymnich:
@wsmoses Should be good to go

wsmoses (approved):

Issue #1743, opened 2024-02-19
Performance of type-analysis


vchuravy:
@swilliamson7 code from https://github.com/EnzymeAD/Enzyme.jl/issues/1284#issuecomment-1939215870 highlights very long compilation time in type-analysis. Happy to share the vtunes file if needed.

The hottest function are all related to type-tree:

![image](https://github.com/EnzymeAD/Enzyme/assets/145258/9450b781-66db-454b-92ed-a45278d1eb44)

in particular VisitGEPOperator seems to be dominating

![image](https://github.com/EnzymeAD/Enzyme/assets/145258/f9f05dea-61ec-4e9a-b6e7-5ac036d4480b)

wsmoses:
So at least some of the shiftindices calls within GEP should be able to be made extremely quick, which should cut down on this quite a bit.

I see almost all of the visitCallBase is hitting the julia rule, which means this isn't running interprocedurally (which is good) so this is possible in theory to extract to a single llvm function which has the same behavior

vchuravy:
The version in https://github.com/michel2323/ShallowWaters.jl/commit/63419446119619c15cb3f83e835d37e11c08436a where @michel2323 fixed Checkpointing.jl takes ~4m and shows a second bottleneck in the call to julia_type_rule. I wonder if we could cache those calls?

![image](https://github.com/EnzymeAD/Enzyme/assets/145258/52933b0d-8703-493f-a88a-883fa120393c)

wsmoses:
Even better than caching the call we could also just use type metadata and never call back into julia.

Enzyme now supports the ability to parse type data on args/insts/etc from metadata/attributes, that we could do a pass first within julia to add to all julia args that are known, then just still tell TA in Julia to be disabled [and it should still use the llvm native metadata].

wsmoses:
here's an example where we do so currently for an alloca: https://github.com/EnzymeAD/Enzyme.jl/blob/1829706d90c4fa2110edb8ac341a8aed68ed3f6f/src/compiler.jl#L4130

we already push the julia type as attributes here: https://github.com/EnzymeAD/Enzyme.jl/blob/1829706d90c4fa2110edb8ac341a8aed68ed3f6f/src/compiler.jl#L4497
might as well also push the typetrees

wsmoses:
@vchuravy let's split this the ShiftIndicies and julia typetree into separate issues. Can you open a new issue for the latter on Enzyme.jl?

[...]

PR #1744, opened 2024-02-19
TypeTree speedup shiftindicies

wsmoses:
Fixes https://github.com/EnzymeAD/Enzyme/issues/1743

### DIFF
--- a/enzyme/Enzyme/TypeAnalysis/TypeTree.h
+++ b/enzyme/Enzyme/TypeAnalysis/TypeTree.h
@@ -965,6 +965,7 @@ public:
           chunk = dl.getPointerSizeInBits() / 8;
         }
         auto offincr = (chunk - offset % chunk) % chunk;
+        bool inserted = false;
         for (int i = offincr; i < maxSize; i += chunk) {
           next[0] = i + addOffset;
           ConcreteType prev(pair.second);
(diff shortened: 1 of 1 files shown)


wsmoses:
note I left the slow version right now to do correctness tests, which already found a bug locally (which have fixed). Going to run it on the whole test suite, then disable the check once it comes back valid.

wsmoses:
Also note this fix hopefully changes the asymptotic behavior of ShiftIndices from $O(n^2)$ to $O(n)$ [which it really should've been the whole time]

vchuravy:
Seems like a speedup from 220s to 180s, but `TypeTree::insert` and `TypeTree:checkedOrIn` are still very hot

Before:

image


After:

image

image

Detailed analysis:

- `checkedOrIn` ~28s (42% of compilation time)
    - 17s from update Analysis 
    - 10.5s from EnzymeMergeTypeTree
- `CanonicalizeInPlace`: 11s

ZuseZ4:
@vchuravy If already at it, could you please also profile 
https://github.com/EnzymeAD/Enzyme/pull/1641 if you have a few minutes? My server never had the permissions for vtune. My new Laptop is reasonable otherwise, so I can also set it up myself

wsmoses:
Considering I don’t see ShiftIndices more on the flame graph I’m going to
call this a win.

I’ll see what I can do about the Julia function. mergeTypeTree is the C API
and primarily used in that function so I think it’s the next most
meaningful place we should be able to reduce overhead from. Essentially
because it should be not terrible to make it called only once. CheckedOrIn
however doesn’t have any obvious speed ups I remember.

PR #1747, opened 2024-02-21
TypeAnalysis speed up Canonicalize

wsmoses:
Again doing a assert equal to old implementation first, then will remove and merge.


ZuseZ4:
improves compile time by 2x to 3.5x (impact becomes larger for bigger vectors).
Will check at some later point if it brings compile times down to something linear.

ZuseZ4 (approved):

PR #1748, opened 2024-02-21
TypeTree add fast path for checkedOrIn

wsmoses:


enzyme/Enzyme/TypeAnalysis/TypeTree.h
```diff
@@ -1239,10 +1239,37 @@ class TypeTree : public std::enable_shared_from_this<TypeTree> {
   /// changed Setting `PointerIntSame` considers pointers and integers as
   /// equivalent If this is an illegal operation, `LegalOr` will be set to false
   bool checkedOrIn(const TypeTree &RHS, bool PointerIntSame, bool &LegalOr) {
-    // TODO detect recursive merge and simplify
+    // Add fast path where nothing could change because all potentially inserted
+    // value are already in the map. Save all other ones that may need slow
+    // insertion
+    std::vector<std::pair<const std::vector<int>, ConcreteType>> todo;
```
tgymnich, line 1245:
```suggestion
    std::vector<std::pair<const std::vector<int>?, ConcreteType>> todo;
```

can we avoid a copy here?

tgymnich, line 1245:
```suggestion
    std::vector<std::pair<const std::vector<int>&, ConcreteType>> todo;
```

can we avoid a copy here?

PR #1750, opened 2024-02-21
Trucation to MPFR

ivanradanov:


ivanradanov:
@wsmoses We need to import `enzyme/Enzyme/Runtime/EnzymeMPFR.h`

enzyme/test/Integration/Truncate/truncate-all.cpp
```diff
@@ -4,6 +4,7 @@
 // Truncated
 // RUN: if [ %llvmver -ge 12 ]; then [ "$(%clang -O3 %s -S -emit-llvm -o - %newLoadClangEnzyme -S -mllvm --enzyme-truncate-all="64to32" | %lli -)" == "900000000.000000" ] ; fi
 // RUN: if [ %llvmver -ge 12 ]; then [ "$(%clang -O3 %s -S -emit-llvm -o - %newLoadClangEnzyme -S -mllvm --enzyme-truncate-all="11-52to8-23" | %lli -)" == "900000000.000000" ] ; fi
+// RUN: if [ %llvmver -ge 12 ]; then [ "$(%clang -O3 %s -S -emit-llvm -o - %newLoadClangEnzyme -S -mllvm --enzyme-truncate-all="11-52to3-7" | %lli -)" == "897581056.000000" ] ; fi
```
wsmoses, line 8: Why not have this as a check line, rather than have it do an output check in bash cmd line?

enzyme/Enzyme/Runtime/EnzymeMPFR.h
```diff
@@ -0,0 +1,122 @@
+//===- EnzymeMPFR.h - MPFR wrappers ---------------------------------------===//
```
wsmoses, line 1: Yeah it would be nice to have this in a file, but still get imported via tablegen, if you have a chance to set that up

ivanradanov: I could not make cmake properly handle the dependency between the included file and the .td so I inlined it the same way as the other headers.

ivanradanov:
Are the julia tests flaky?

enzyme/CMakeLists.txt
```diff
@@ -265,6 +267,11 @@ string(REPLACE "};\n}" "};\n}}" INPUT_TEXT "${INPUT_TEXT}")
 string(REPLACE "const SCEV* S;\n};\n" "const SCEV* S;\n};\n}\n" INPUT_TEXT "${INPUT_TEXT}")
 endif()
 
+find_library(MPFR_LIB_PATH mpfr)
```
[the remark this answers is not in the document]
ivanradanov: It is optional, this just checks whether we found the library and enables the MPFR test only if we found it.

enzyme/Enzyme/Runtime/EnzymeMPFR.h
ivanradanov: After hacking around with custom header importing I am thinking that the cleanest solution is for us to properly provide header file installation and ship them, what do you think?

Another option is to compile the mpfr wrappers in a static library and ship that.

enzyme/CMakeLists.txt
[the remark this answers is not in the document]
wsmoses: great

enzyme/Enzyme/Runtime/EnzymeMPFR.h
wsmoses: I'm okay either way. Pro of havng integrated header means that users of Enzyme as a library have definitional guarantees about semantics of the header/verion/custom syntax inside. On other hand, seeing code in include file can also be nice

wsmoses: Maybe leave it as is for now, and we can adjust later?

wsmoses (approved):

PR #1751, opened 2024-02-21
Fixed incorrect usage of llvm::Function::splice.

matinraayai:
cc: @wsmoses


ivanradanov:
`Function::splice` seems to be unavailable in llvm 11, does the current approach break it some cases?

wsmoses:
Sounds like we should ifdef the code there

matinraayai:
It appears that `getBasicBlocks` function became private in LLVM 16. As Billy said, we have to `#ifdef` the code

matinraayai:
@wsmoses I can fix the clang-format. Also I'm not sure what makes Bazel fail.

PR #1768, opened 2024-02-28
Improve handling of c++ attrs

wsmoses:
Now should work with templates

### DIFF
--- a/enzyme/Enzyme/PreserveNVVM.cpp
+++ b/enzyme/Enzyme/PreserveNVVM.cpp
@@ -368,6 +368,106 @@ bool preserveNVVM(bool Begin, Function &F) {
       "__enzyme_register_derivative";
   constexpr static const char splitderivative_handler_name[] =
       "__enzyme_register_splitderivative";
+
+  if (Begin)
+    if (GlobalVariable *GA =
+            F.getParent()->getGlobalVariable("llvm.global.annotations")) {
+      if (GA->hasInitializer()) {
+        auto AOp = GA->getInitializer();
+        // all metadata are stored in an array of struct of metadata
+        if (ConstantArray *CA = dyn_cast<ConstantArray>(AOp)) {
+          // so iterate over the operands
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/PreserveNVVM.cpp
```diff
@@ -368,6 +368,106 @@ bool preserveNVVM(bool Begin, Function &F) {
...
+
+                    Function *Func = dyn_cast<Function>(Val);
+                    GlobalVariable *Glob = dyn_cast<GlobalVariable>(Val);
+
+                    if (AS == "enzyme_inactive" && Func && Func == &F) {
+                      Func->addAttribute(AttributeList::FunctionIndex,
+                                         Attribute::get(Func->getContext(),
+                                                        "enzyme_inactive"));
+                      changed = true;
+                      changed |= preserveLinkage(Begin, F);
```
tgymnich, line 409: true v x => true

enzyme/Enzyme/PreserveNVVM.cpp
```diff
@@ -368,6 +368,106 @@ bool preserveNVVM(bool Begin, Function &F) {
...
+      if (GA->hasInitializer()) {
+        auto AOp = GA->getInitializer();
+        // all metadata are stored in an array of struct of metadata
+        if (ConstantArray *CA = dyn_cast<ConstantArray>(AOp)) {
+          // so iterate over the operands
+          SmallVector<Constant *, 1> replacements;
+          for (Value *CAOp : CA->operands()) {
+            // get the struct, which holds a pointer to the annotated function
+            // as first field, and the annotation as second field
+            if (ConstantStruct *CS = dyn_cast<ConstantStruct>(CAOp)) {
```
tgymnich, line 384: Can we reduce the indentation level here with something like this
```
ConstantStruct *CS = dyn_cast<ConstantStruct>(CAOp
if (!CS)
   continue;
```

same for the following lines.

wsmoses:
superceeded by upstream

PR #1769, opened 2024-02-28
[mlir] Define generic set/map lattices for dataflow analyses

pengmai:
For context, the conceptual `AliasClassSet`, `AliasClassLattice`, and `PointsToSets` data structures were immensely useful in implementing activity annotations. The "`lattice of sets that might be undefined or unknown`" and "`map of attributes to those sets`" are exactly the kind of data structures that activity annotations uses to propagate these relative "origins" both forward and backward, in an intraprocedural analysis.

This PR aims to be an intermediate step towards merging in activity annotations by breaking out the re-usable logic into abstract classes.

### DIFF
--- a/enzyme/Enzyme/MLIR/Analysis/CMakeLists.txt
+++ b/enzyme/Enzyme/MLIR/Analysis/CMakeLists.txt
@@ -2,6 +2,7 @@ add_mlir_library(MLIREnzymeAnalysis
   ActivityAnalysis.cpp
   DataFlowAliasAnalysis.cpp
   DataFlowActivityAnalysis.cpp
+  DataFlowLattice.cpp
 
   DEPENDS
   MLIRAutoDiffTypeInterfaceIncGen
--- a/enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
+++ b/enzyme/Enzyme/MLIR/Analysis/DataFlowActivityAnalysis.cpp
@@ -490,7 +490,7 @@ std::optional<Value> getCopySource(Operation *op) {
 /// If the classes are undefined, the callback will not be called at all.
 void forEachAliasedAlloc(const AliasClassLattice *ptrAliasClass,
                          function_ref<void(DistinctAttr)> forEachFn) {
-  (void)ptrAliasClass->getAliasClassesObject().foreachClass(
+  (void)ptrAliasClass->getAliasClassesObject().foreachElement(
       [&](DistinctAttr alloc, enzyme::AliasClassSet::State state) {
         if (state != enzyme::AliasClassSet::State::Undefined)
           forEachFn(alloc);
@@ -636,7 +636,7 @@ public:
           continue;
         auto *argAliasClasses = getOrCreateFor<AliasClassLattice>(block, arg);
         ChangeResult changed =
-            argAliasClasses->getAliasClassesObject().foreachClass(
+            argAliasClasses->getAliasClassesObject().foreachElement(
                 [lattice](DistinctAttr argAliasClass,
                           enzyme::AliasClassSet::State state) {
                   if (state == enzyme::AliasClassSet::State::Undefined)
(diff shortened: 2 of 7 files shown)


enzyme/Enzyme/MLIR/Analysis/Lattice.h
```diff
@@ -0,0 +1,364 @@
...
+  /// is *not* a classical singleton, there are other ways of obtaining an
+  /// "unknown" alias set.
+  static const SetLattice<ValueT> &getUnknown() { return unknownSet; }
+
+  bool operator==(const SetLattice<ValueT> &other) const {
+    assert(isCanonical() && other.isCanonical());
+    return state == other.state && llvm::equal(elements, other.elements);
+  }
+
+  void print(llvm::raw_ostream &os) const {
```
tgymnich, line 136:
```suggestion
  LLVM_DUMP_METHOD void print(llvm::raw_ostream &os) const {
```

Add LLVM_DUMP_METHOD so that this method does not get stripped for debug builds.

wsmoses (approved):

enzyme/Enzyme/MLIR/Analysis/Lattice.cpp
```diff
@@ -0,0 +1,44 @@
...
+bool enzyme::sortAttributes(Attribute a, Attribute b) {
+  auto distinctA = dyn_cast<DistinctAttr>(a);
+  auto distinctB = dyn_cast<DistinctAttr>(b);
+  if (distinctA && distinctB) {
+    auto strA = dyn_cast_if_present<StringAttr>(distinctA.getReferencedAttr());
+    auto strB = dyn_cast_if_present<StringAttr>(distinctB.getReferencedAttr());
+    if (strA && strB)
+      return strA.strref() < strB.strref();
+  }
+  // If there's no string to compare, sort them arbitrarily.
```
ftynse, line 42: How about printing attributes to a `raw_string_ostream` and comparing those?
Also, `&a` gives you the address of the local variable in this function, which is the same all the time.

enzyme/Enzyme/MLIR/Analysis/Lattice.h
```diff
@@ -0,0 +1,364 @@
...
+  /// is considered to map to an uninitialized set.
+  DenseMap<KeyT, SetLattice<ElementT>> map;
+
+private:
+  Attribute serializeMapOfSetsNaive(MLIRContext *ctx) const {
+    SmallVector<Attribute> pointsToArray;
+
+    for (const auto &[srcClass, destClasses] : map) {
+      SmallVector<Attribute, 2> pair = {srcClass};
+      SmallVector<Attribute, 5> aliasClasses;
```
ftynse, line 338: Nit: don't chose specific number of stack elements unless you have a strong reason to.

enzyme/Enzyme/MLIR/Analysis/Lattice.h
```diff
@@ -0,0 +1,364 @@
...
+      if (destClasses.isUnknown()) {
+        aliasClasses.push_back(StringAttr::get(ctx, "unknown"));
+      } else if (destClasses.isUndefined()) {
+        aliasClasses.push_back(StringAttr::get(ctx, "undefined"));
+      } else {
```
ftynse, line 343: Can we unify these string representations with what is used above (e.g., `"<unknown>"`) by putting the strings in some named constant?

enzyme/Enzyme/MLIR/Analysis/Lattice.h
```diff
@@ -0,0 +1,364 @@
...
+        llvm::sort(aliasClasses, sortAttributes);
+      }
+      pair.push_back(ArrayAttr::get(ctx, aliasClasses));
+      pointsToArray.push_back(ArrayAttr::get(ctx, pair));
+    }
+    llvm::sort(pointsToArray, [&](Attribute a, Attribute b) {
+      auto arrA = cast<ArrayAttr>(a);
+      auto arrB = cast<ArrayAttr>(b);
+      return sortAttributes(arrA[0], arrB[0]);
+    });
```
ftynse, line 356: There's no guarantee the array is non-empty. We could rather sort entire arrays lexicographically.

ftynse (approved):
LGTM in general, but I'd like to make sorting more reliable even if it costs us more. Otherwise we are likely to have schroedinbugs in printing.

PR #1782, opened 2024-03-07
Fix bazel build for macos

wsmoses:
Macos Bazel build didn't like us calling something AliasAnalysis.h and MLIR also having the same leading to nonsense issues that was strictly easier for us to rename to DataFlowAliasAnalysis anyways


pengmai (approved):

pengmai:
I think the `CMakeLists.txt` file needs updating as the .cpp got renamed

PR #1787, opened 2024-03-09
Bazel Cache

tgymnich:


wsmoses:
So the bazel builds are always the latest llvm commit so I'm not sure how much the gain of caching will be vs the cache size

tgymnich:
We still get the benefits of incremental builds of both Enzyme and LLVM.

 2m 29s vs. 2h 13m 45s

seems worth it to me.

wsmoses:
Oh that's very substantial then, definitely should merge.

wsmoses (approved):

Issue #1785, opened 2024-03-08
New C++ interface with lambda


jandrej:
Reproducer https://fwd.gymni.ch/y9blF5

tgymnich:
`error: /enzymeroot/enzyme/utils:221:63: in function main i32 (i32, ptr): Enzyme: Cannot cast __enzyme_autodiff primal argument 6, found   %263 = load i32, ptr %this1.i248, align 4, !dbg !40982, type i32 - to arg 1 ptr
`

Could you also post some C++ code for this. This seems like a frontend error (wrong enzyme invocation).

jandrej:
The only code I can share is a similar MFE https://fwd.gymni.ch/2szOgk

PR #1789, opened 2024-03-11
cpp interface improvements

samuelpmish:
working on some fixes and improvements for the new cpp interface:

- fixing inconsistent `size_t` vs `std::size_t` usage
- use automatically generated ctors for `enzyme::Active`, `enzyme::Const`, ... , to support brace-initialization
- add CTAD guides for C++17 and later (so users don't have to explicitly name the argument types)
- added a new directory in tests for c++ interface examples and unit tests with CTest support (WIP)
- disambiguate `forward_by_reference` implementation (see https://github.com/EnzymeAD/Enzyme/issues/1785)

I think @wsmoses also mentioned something about casting the function pointers to the appropriate types-- can you clarify what you want to happen and I'll add a test for it.

I'm also seeing an error in the return type deduction for Forward mode, so I'll look into that.

### DIFF
--- a/enzyme/Enzyme/Clang/include_utils.td
+++ b/enzyme/Enzyme/Clang/include_utils.td
@@ -28,6 +28,9 @@ Return __enzyme_fwddiff(T...);
 
 namespace enzyme {
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wmissing-braces"
+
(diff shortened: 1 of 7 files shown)


wsmoses (commented):
Can we move the sugar tests from fwd and reverse mode integration folders here too?

PR #1798, opened 2024-03-20
ADBench ba in rust

ZuseZ4:
Under debug it fails due to TA not understanding memcpy.
Under release it works.

Reproducing:
`export ENZYME_LOOSE_TYPES=1`
As usual, afterwards just 
`cargo +enzyme run --release`.


ZuseZ4:
merged into other pr

PR #1818, opened 2024-03-31
Fix gamma assertion error

wsmoses:


ZuseZ4 (approved):

ZuseZ4:
Not 100% sure @wsmoses, but I think this is the one where you hoped it would fix the following bug. Feel free to disregard if I mixed things up.

```
rustc: /home/jeka2967/src/rust/src/tools/enzyme/enzyme/Enzyme/AdjointGenerator.h:5376: void AdjointGenerator::recursivelyHandleSubfunction(llvm::CallInst&, llvm::Function*, const std::vector<bool>&, bool, DIFFE_TYPE, bool): Assertion `subdata' failed.
```

PR #1837, opened 2024-04-15
Alternative way of bundling headers

ivanradanov:
@wsmoses what do you think?


enzyme/BUILD
```diff
@@ -170,7 +156,7 @@ cc_library(
     hdrs = glob([
         "Enzyme/*.h",
         "Enzyme/TypeAnalysis/*.h",
-    ]),
+    ]) + ["Enzyme/Clang/bundled_includes.h"],
```
ivanradanov, line 159: This header gets generated in the output dir, and the compiler does not find it, @wsmoses do you know how to provide the include path for it?

```Enzyme/Clang/EnzymeClang.cpp:40:10: fatal error: 'bundled_includes.h' file not found
   40 | #include "bundled_includes.h"
      |          ^~~~~~~~~~~~~~~~~~~~
1 error generated.
```

ivanradanov: @wsmoses sorry I think I figured it out

wsmoses (approved):

PR #1847, opened 2024-04-25
[Truncate] Fix bug with fcmp handling

ivanradanov:
We did not pass the truncated values to the runtime


wsmoses (approved):
LGTM, but maybe test?

PR #1852, opened 2024-04-28
Bazel fix include script on non outermost project

wsmoses:


ivanradanov:
Sorry for that. I am getting the same errors as in the bazel CI so I cannot check it locally though.

wsmoses:
Yeah I tested it successfully on an earlier llvm commit I have locally so just merging for now.

PR #1853, opened 2024-04-28
Add __*_finite versions of derivatives

ivanradanov:
From my understanding, the __*_finite versions of math functions are either aliases to the normal ones or faster versions that can assume no infinite operands. This automatically generates the same derivatives for them as the normal function. 

@wsmoses How do you think we should handle this? Currently it would also generate derivatives for `___nv_asin_finite` for example, which doesn't exist.

### DIFF
--- a/enzyme/test/Enzyme/ForwardMode/hypot.ll
+++ b/enzyme/test/Enzyme/ForwardMode/hypot.ll
@@ -8,13 +8,26 @@ entry:
   ret double %call
 }
 
+define double @tester2(double %x, double %y) {
+entry:
+  %call = tail call double @__hypot_finite(double %x, double %y)
+  ret double %call
+}
+
 define double @test_derivative(double %x, double %y) {
 entry:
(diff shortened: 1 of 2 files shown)


ivanradanov:
We could wither have separate fields for names we want to generate `__*_finite` versions for, and ones that we don't. Or we could leave it as is and assume people would not define custom functions with names such as `___nv_asin_finite`

wsmoses:
Slightly generalizing, maybe we could add a "suffixof" tablegen command. Specifically since this same dupliaction also is required for cos, cosf, cosl,etc.

So maybe something like here https://github.com/EnzymeAD/Enzyme/blob/5468479e8f3de4658205751d42a6fbe073487e2e/enzyme/Enzyme/InstructionDerivatives.td#L335

but now

`SameTypesSuffixFunc<"tanh", "cosh">` which will take the actual name of the function (tanhf), and remplace the prefix tanh with cosh.

PR #1881, opened 2024-05-14
abort compilation if cfg not found

### LINKED ISSUE #1815: abort cmake when -DLLVM_DIR is an invalid path.
Jed, I, and probably other people have more than one llvm version laying around. 
If you build Enzyme you usually want to use a specific llvm, at least when you explicitely set the path.
However, if a typo happens Enzyme will just use whatever llvm it can find and ignore the invalid value.
This leads to all kind of surprising behaviour later.
I would argue that if a user uses this flag, it cares about this llvm getting used, so we should abort if the path doesn't exist?

ZuseZ4:
If we provide a path, we want Enzyme to actually use it.
If that's not possible, rather error than taking a surprising other version.

closes #1815

### DIFF
--- a/enzyme/CMakeLists.txt
+++ b/enzyme/CMakeLists.txt
@@ -55,6 +55,9 @@ set(LLVM_DIR "${LLVM_ABSOLUTE_DIR}" CACHE FILEPATH "b" FORCE)
 
 if (EXISTS "${LLVM_DIR}/lib/cmake/llvm/LLVMConfig.cmake")
   set(LLVM_DIR "${LLVM_DIR}/lib/cmake/llvm")
+else()
+  message("LLVM Config: ${LLVM_DIR}/lib/cmake/llvm/LLVMConfig.cmake")
+  message(FATAL_ERROR "LLVM Config not found")
 endif()
 
 message("LLVM ABS DIR " ${LLVM_DIR})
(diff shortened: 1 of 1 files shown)


enzyme/CMakeLists.txt
```diff
@@ -55,6 +55,9 @@ set(LLVM_DIR "${LLVM_ABSOLUTE_DIR}" CACHE FILEPATH "b" FORCE)
 
 if (EXISTS "${LLVM_DIR}/lib/cmake/llvm/LLVMConfig.cmake")
   set(LLVM_DIR "${LLVM_DIR}/lib/cmake/llvm")
+else()
+  message("LLVM Config: ${LLVM_DIR}/lib/cmake/llvm/LLVMConfig.cmake")
+  message(FATAL_ERROR "LLVM Config not found")
```
wsmoses, line 60: Can you add some more info (e.g. LLVM Config not found at location LLVM_DIR="${LLVM_DIR}")

ZuseZ4: Looking for the LLVM config is too precise (see incorrectly failing .bc CI).
I updated it to only check if the path exists. This still catches typos.

ZuseZ4:
The mlir build now fails with the new error. Is it correctly failing?

wsmoses:
That's actually good and means it works. For some reason MLIR CI on release (but not the debug) builder on Github wasn't caching correctly.

wsmoses:
@ZuseZ4 just wanted to say what a quality of life improvement this was [just caught an mistake I had made]

PR #1889, opened 2024-05-19
Update with the file we used for benchmarking

martinjm97:


enzyme/test/Integration/Sparse/eigen_analysis.cpp
```diff
@@ -143,7 +143,7 @@ template<typename T>
 __attribute__((always_inline))
 static void gradient_ip(const T *__restrict__ pos0, const size_t num_faces, const int* faces, const T *__restrict__ x, T *__restrict__ out)
 {
-    __enzyme_autodiff<void>((void *)eigenstuffM<T>,
+    __enzyme_autodiff<void>((void *)eigenstuffL<T>,
```
wsmoses, line 146: Since this is a test can you also have a function which differentiates the other function as well

wsmoses (approved):

Issue #1904, opened 2024-05-23
C++ interface templates appear to be broken


GregTheMadMonk:
Hello! Playing with C++ interface trying to get it to work and this simple program:
```cpp
#include <vector>

#include <enzyme/enzyme>

using Flt = double;
using Vec = std::vector<Flt>;

Flt sum(const Vec& x) {
    Flt ret = 0.0;
    for (const auto xi : x) ret += xi;
    return ret;
} // <-- sum

Vec dsum(const Vec& x) {
    Vec dx(x.size());
    return enzyme::get<0>(
        enzyme::autodiff<enzyme::Forward>(
            sum,
            enzyme::DuplicatedNoNeed<const Vec&>{ x, dx }
        )
    );
} // <-- dsum

int main() {
    return 0;
}
```

doesn't compile with the following errors:

```
In file included from /media/ssd/projects/cpp/enzyme/enzyme-tests/cpp-interface-test/main.cc:3:
In file included from /enzymeroot/enzyme/enzyme:3:
/enzymeroot/enzyme/utils:68:7: error: multiple overloads of 'DuplicatedNoNeed' instantiate to the same signature 'void (const std::vector<double> &, const std::vector<double> &)'
   68 |       DuplicatedNoNeed(T&& v, T&& s) : value(v), shadow(s) {};
      |       ^
/media/ssd/projects/cpp/enzyme/enzyme-tests/cpp-interface-test/main.cc:19:13: note: in instantiation of template class 'enzyme::DuplicatedNoNeed<const std::vector<double> &>' requested here
   19 |             enzyme::DuplicatedNoNeed<const Vec&>{ x, dx }
      |             ^
/enzymeroot/enzyme/utils:67:7: note: previous declaration is here
   67 |       DuplicatedNoNeed(const T& v, const T& s) : value(v), shadow(s) {};
      |       ^
In file included from /media/ssd/projects/cpp/enzyme/enzyme-tests/cpp-interface-test/main.cc:1:
In file included from /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/14.1.1/../../../../include/c++/14.1.1/vector:86:
In file included from /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/14.1.1/../../../../include/c++/14.1.1/bits/memory_resource.h:41:
[...]

PR #1914, opened 2024-06-02
C++ interface fixes: reference argument types

GregTheMadMonk:
Related issue: #1904

There are still some fixes that need to be implemented (autodiff on a function pointer doesn't work for some reason, I'll write about it in the issue) as well as tests. Putting it here before these are finished to make sure that my assumptions about how `Duplicated` and `DuplicatedNoNeed` should work were correct and they are indeed only used with reference and pointer arguments.

* Fix move constructors
* Explicitly `impl::tuple_cat` instead of `std::tuple_cat`
* Restrict `Duplicated` and `DuplicatedNoNeed` template arguments to only allow reference and pointer types
* Properly initialize shadow in `Duplicated` and `DuplicatedNoNeed`

### DIFF
--- a/enzyme/include/enzyme/utils
+++ b/enzyme/include/enzyme/utils
@@ -41,6 +41,10 @@ namespace enzyme {
 
     template < typename T >
     struct Active{
+      static_assert(
+        !std::is_reference_v<T> && !std::is_pointer_v<T>,
+        "Reference/pointer active arguments don't make sense for AD!"
+      );
       T value;
       Active(const T& v) : value(v) {};
       Active(T&& v) : value(std::move(v)) {};
@@ -66,25 +70,15 @@ namespace enzyme {
             DupArg(T&& v, T&& s) : value(v), shadow(s) {};
             DupArg(const DupArg<T>&) = default;
             DupArg(DupArg<T>&&) = default;
-
-            __attribute__((always_inline))
-            T valueExp() const { return this->value; }
-            __attribute__((always_inline))
-            T shadowExp() const { return this->shadow; }
         };
 
         template < typename T >
         struct DupArg<T&> {
-            T& value;
-            T& shadow;
-            DupArg(T& v, T& s) : value(v), shadow(s) {}
+            T* value;
+            T* shadow;
+            DupArg(T& v, T& s) : value(&v), shadow(&s) {}
             DupArg(const DupArg<T&>&) = default;
             DupArg(DupArg<T&>&&) = default;
-
-            __attribute__((always_inline))
-            const T* valueExp() const { return &this->value; }
-            __attribute__((always_inline))
-            const T* shadowExp() const { return &this->shadow; }
         };
 
     } // <-- namespace detail
@@ -109,6 +103,15 @@ namespace enzyme {
       operator T&() { return value; }
     };
 
+    template < typename T >
+    struct Const<T&>{
+        T* value;
+        Const(T& v) : value(&v) {}
+        Const(const Const<T&>&) = default;
+        Const(Const<T&>&&) = default;
+        operator T&() { return *value; }
+    };
+
     // CTAD available in C++17 or later
     #if __cplusplus >= 201703L
       template < typename T >
@@ -126,13 +129,29 @@ namespace enzyme {
 
     template < typename T >
     struct type_info {
-      static constexpr bool is_active = false;
+      static constexpr bool is_active     = false;
+      static constexpr bool is_duplicated = false;
       using type = nodiff;
     };
 
     template < typename T >
     struct type_info < Active<T> >{
-        static constexpr bool is_active = true;
+        static constexpr bool is_active     = true;
+        static constexpr bool is_duplicated = false;
+        using type = T;
+    };
+
+    template < typename T >
+    struct type_info < Duplicated<T> >{
+        static constexpr bool is_active     = false;
+        static constexpr bool is_duplicated = true;
+        using type = T;
+    };
+
+    template < typename T >
+    struct type_info < DuplicatedNoNeed<T> >{
+        static constexpr bool is_active     = false;
+        static constexpr bool is_duplicated = true;
         using type = T;
     };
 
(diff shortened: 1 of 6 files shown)


GregTheMadMonk:
```
my assumptions about how Duplicated and DuplicatedNoNeed should work were correct
```
I've just realized that they are wrong. I assumed that `Duplicated[NoNeed]` only need to work with reference/pointer arguments, but now I realize that regular values my need to be passed in forward mode.

Should there maybe be another class made, maybe `enzyme::ForwardActive`, to stress the difference between reverse mode storing the result in the shadow parameter and forward mode using the shadow parameter as a "direction" to compute the gradient?

wsmoses:
@GregTheMadMonk Duplicated / DuplicatedNoNeed permit arbitrary types in forward mode (and also you may want to do a duplicated<int> if say that represents mutable memory in some way e.g. ptrtoint.

It may be worthwhile splitting Duplicated -> Dual in forward mode and we can then add the duplicated restriction in reverse mode -- but that should likely come in a later PR.

wsmoses:
@GregTheMadMonk bumping this, could you remove the duplicated error for now [and add a test for reference]

GregTheMadMonk:
Yes, sorry, I've encountered a weird issue when trying to make it work with lambdas where I somehow managed to modify the headers so that Enzyme doesn't (seemingly) even start a pass and leaves `__enzyme_autodiff` call as is and then some other stuff kept me distracted for a while. Finishing it rn, so next week (or even tomorrow maybe) I can isolate only duplicated args part and update the PR.

GregTheMadMonk:
Sorry for the delay, here is the code.

The code from #1904 is in `gh_issue_1904.cpp` - the test fails only with `-O0`, I've left it in so that it can be looked into later. All other optimizer levels work fine, `dup_args.cpp` passes on all levels.

I've kept the restriction on pointer/reference types only in `Duplicated`/`DuplicatedNoNeed`, but only for reverse mode, and I've put it in `static_assert` with an explaining message.

Some other stuff about C++ interface that I'll be looking into next:
* ~Should Active/Const reference type arguments also get an update?~ Yes, I've just realized they should. I'll add commits here while
* How to make lambdas/other function-like objects work (they obviously are intended to work, but don't as far as I can tell)

enzyme/include/enzyme/utils
```diff
@@ -43,37 +43,67 @@ namespace enzyme {
...
+        template < typename T >
+        struct DupArg<T&> {
+            T& value;
+            T& shadow;
+            DupArg(T& v, T& s) : value(v), shadow(s) {}
+            DupArg(const DupArg<T&>&) = default;
+            DupArg(DupArg<T&>&&) = default;
+
+            __attribute__((always_inline))
+            const T* valueExp() const { return &this->value; }
```
wsmoses, line 85: This shouldn't be const here, no?

wsmoses: It you want to keep const correctness here, maybe we should store T* for value and shadow?

enzyme/test/Integration/CppSugar/dup_args.cpp
```diff
@@ -0,0 +1,121 @@
...
+#if 0
+    {
+        /*
+         * This will not work in the C++ interface because it is also not allowed
+         * with the C-style __enzyme_autodiff call
+         */
+        const float dfdx = enzyme::get<0>(
+            enzyme::get<0>(
+                enzyme::autodiff<enzyme::Reverse>(
+                    g, enzyme::Active<const float&>{ x }
```
wsmoses, line 41: So here I think it would be easiest to just disallow reference active's

enzyme/test/Integration/CppSugar/dup_args.cpp
```diff
@@ -0,0 +1,121 @@
...
+                )
+            )
+        );
+        APPROX_EQ(dfdx, etalon, 1e-10);
+    }
+#endif
+#if 0
+    {
```
wsmoses, line 49: We should be able to make a separate file and do verify errors here. See this for example: https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/test/Integration/ReverseMode/err_empty.c

enzyme/include/enzyme/utils
GregTheMadMonk: Yeah `T* value` would be much simpler as it would do the same thing that `valueExp` and `shadowExp` are doing anyway...

As for the `const`... I believe it doesn't matter a lot at this point where the `const` is because the next destination for these pointers is `__enzyme_autodiff` and by the time it reaches IR the semantic value of `const` will be already lost...

wsmoses: That's true, but if Duplicated gets used semantically for how we define custom c++ derivative rules we want users to be able to modify things appropriately.

enzyme/test/Integration/CppSugar/dup_args.cpp
GregTheMadMonk: But `Const`s should have a reference specification like `Duplicated`/`DuplicatedNoNeed`, right?

wsmoses: Yup!

enzyme/include/enzyme/utils
GregTheMadMonk: That would be a bit tricky to do. Let's take `enzyme::Duplicated<const float&>` for example. What should it's `shadow` type be? In reverse mode it will get modified by the AD, so its const-correct type is `float&`. On the contrary, forward mode will only read the value from that address, so the user should be able to bind a `const` to that reference, which means that the shadow in this case needs to be a `const float&`.

Unless I'm missing something here, there is a contradiction here in what `Reverse` and `Forward` mode want from the shadow parameter in `Duplicated` which can't be solved unless there is a separate class for the `Forward` mode specifically

wsmoses: Yeah we really should fork off duplicated in forward mode to just be Dual and have the semantic changes.

In any case yeah, we may need to remove the const from the primal type in the shadow.

This doesn’t need to stop the PR from getting in to begin with (already improving tuings) — but something to follow up on

wsmoses:
@GregTheMadMonk once tests pass [and if you can fork off the error checks like above], this should be good to merge!

GregTheMadMonk:
* Disabled reference/pointer `Active` specializations via a `static_assert`
* Added a reference-type specialization to `Const`
* Removed `valueExp`/`shadowExp` in favor of just using pointer members
* Removed `is_duplicated` in favor of extending `type_info` (also fixed `DuplicatedNoNeed` not having that trait)
* Added tests to changes
* Added tests that are supposed to be a compile-time error (is there a way to make them not break every time line numbers in enzyme/utils change?)

About `gh_issue_1904`: I'm not sure what to do with it. It just segfaults with `-O0` and it's not a C++-interface problem (checked with `__enzyme_autodiff` and it is the same). I could just remove the `-O0` tests but it would feel like cheating...

jandrej:
Sharing the failing code through compiler explorer https://fwd.gymni.ch/PE2qjx. This still doesn't give the correct answer.

wsmoses:
@GregTheMadMonk I've fixed an O0 issue with https://github.com/EnzymeAD/Enzyme/pull/1993. Let's see if that makes this succeed.

PR #1921, opened 2024-06-11
Update TypeAnalysis.cpp

ZuseZ4:
fixes https://github.com/EnzymeAD/rust/issues/125
(at least the Enzyme bug, TA still fails even with loose types).

Extending i128 to 64 bit doesn't really work and llvm allows to directly check for -1, so let's make it more explicit.


wsmoses (approved):
Add a test so we don’t regress?

PR #1952, opened 2024-06-30
Add batching pass

wsmoses:
cc @mofeing @avik-pal


mofeing:
I'm not sure how batching works in Enzyme, but some StableHLO ops have support for batching (e.g. `dot_general`). Should this pass modify those ops?

wsmoses:
Eventually, yes but for now this just needs to operate on scalars (to implement broadcasting properly in reactant)

PR #1962, opened 2024-07-04
Add callback inactive callback to nofrees set

rmoyard:
Hi @wsmoses !

We would like use to the latest version of Enzyme, but changes made in #1687 are not compatible with our `inactive_callback` function that we use inside an Enzyme inactive function. I have added `__catalyst_inactive_callback`  to the set of NoFrees, not sure if you are willing to accept it? I would like to avoid to turn the option that all unknown functions are marked with no free. Let me know if you see a more dynamic solution?


wsmoses:
Is there a reason you can't just add a NoFree attribute to `__catalyst_inactive_callback` directly?

rmoyard:
Thanks @wsmoses, I got confused with the LLVM MLIR dialect and adding attributes but found a solution. I will close this PR.

PR #2013, opened 2024-07-26
fix enzyme_type usage in memcpy

ZuseZ4:
test in the making

### DIFF
--- a/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
+++ b/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
@@ -3415,15 +3415,15 @@ void TypeAnalyzer::visitMemTransferCommon(llvm::CallBase &MTI) {
     }
   }
 
-  auto &dl = MTI.getParent()->getParent()->getParent()->getDataLayout();
+  auto &DL = MTI.getParent()->getParent()->getParent()->getDataLayout();
   TypeTree res = getAnalysis(MTI.getArgOperand(0))
                      .PurgeAnything()
                      .Data0()
-                     .ShiftIndices(dl, 0, sz, 0);
+                     .ShiftIndices(DL, 0, sz, 0);
   TypeTree res2 = getAnalysis(MTI.getArgOperand(1))
                       .PurgeAnything()
                       .Data0()
-                      .ShiftIndices(dl, 0, sz, 0);
+                      .ShiftIndices(DL, 0, sz, 0);
 
   bool Legal = true;
   res.checkedOrIn(res2, /*PointerIntSame*/ false, Legal);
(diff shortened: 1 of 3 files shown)


ZuseZ4:
@wsmoses Why isn't this one good enough to reproduce?
```
declare void @__enzyme_autodiff(void (ptr, ptr, i64, i1)*, ...) 

declare void @llvm.memcpy.p0.p0.i64(ptr, ptr, i64, i1)

define void @square(ptr %0, ptr %1) {
entry:
  ;call void @llvm.memcpy.p0.p0.i64(ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,0]:Float@double, [-1,8]:Float@double, [-1,16]:Float@double, [-1,24]:Float@double, [-1,32]:Float@double, [-1,40]:Float@double, [-1,48]:Float@double, [-1,56]:Float@double, [-1,64]:Float@double, [-1,72]:Float@double}" %0, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,0]:Float@double, [-1,8]:Float@double, [-1,16]:Float@double, [-1,24]:Float@double, [-1,32]:Float@double, [-1,40]:Float@double, [-1,48]:Float@double, [-1,56]:Float@double, [-1,64]:Float@double, [-1,72]:Float@double}" %1, i64 "enzyme_type"="{[-1]:Integer}" 80, i1 false)
  ;call void @llvm.memcpy.p0.p0.i64(ptr align 8 %0, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,0]:Float@double, [-1,8]:Float@double, [-1,16]:Float@double, [-1,24]:Float@double, [-1,32]:Float@double, [-1,40]:Float@double, [-1,48]:Float@double, [-1,56]:Float@double, [-1,64]:Float@double, [-1,72]:Float@double}" %1, i64 "enzyme_type"="{[-1]:Integer}" 80, i1 false)
  ;call void @llvm.memcpy.p0.p0.i64(ptr align 8 %0, ptr align 8 "enzyme_type"="{[-1]:Pointer, [-1,0]:Float@double, [-1,8]:Float@double, [-1,16]:Float@double, [-1,24]:Float@double, [-1,32]:Float@double, [-1,40]:Float@double, [-1,48]:Float@double, [-1,56]:Float@double, [-1,64]:Float@double, [-1,72]:Float@double}" %1, i64 80, i1 false)
  call void @llvm.memcpy.p0.p0.i64(ptr align 8 %0, ptr align 8 %1, i64 80, i1 false)
  ret void
}

define void @dsquare(ptr %x, ptr %dx, ptr %y, ptr %dy) {
entry:
  tail call void (void (ptr, ptr)*, ...) @__enzyme_autodiff(void (ptr, ptr)* nonnull @square, ptr %x, ptr %dx, ptr %y, ptr %dy)
  ret void
}
```
https://fwd.gymni.ch/SzQLez

wsmoses:
https://fwd.gymni.ch/pjnAmb

ZuseZ4:
@wsmoses your fixed testcases fails both before and after this PR:
```
opt ../test/TypeAnalysis/cpy.ll -load-pass-plugin=/home/manuel/prog/Enzyme/enzyme/build/Enzyme/LLVMEnzyme-17.so -passes="print-type-analysis" -type-analysis-func=cpy -enzyme-strict-aliasing=0 -S
cpy - {} |{[-1]:Integer}:{} {[-1]:Pointer}:{} {[-1]:Pointer}:{} 
i1 %c: {[-1]:Integer}
ptr %0: {[-1]:Pointer}
ptr %1: {[-1]:Pointer}
```
(and the corresponding autodiff test also fails with can't deduce type). Why is Enzyme still not picking it up here?

Based on some debugging, `DifferentiableMemCopyFloats` doesn't even seem to get called, so we added the code in the wrong place.

ZuseZ4:
@wsmoses As discussed, moved it to MemTransfercommon. 
I also create the right TT through parsing and use those to update the previous TT for each arg. 
However, passing those new TT to UpdateAnalysis doesn't seem to have any effect (aka changes won't get stored), 
can you tell why? I annotated the code with some dbg info.

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -3450,19 +3450,48 @@ void TypeAnalyzer::visitMemTransferCommon(llvm::CallBase &MTI) {
...
-#else
-  for (unsigned i = 2; i < MTI.getNumArgOperands(); ++i)
-#endif
-  {
-    updateAnalysis(MTI.getArgOperand(i),
-                   TypeTree(BaseType::Integer).Only(-1, &MTI), &MTI);
+  for (unsigned i = 0; i < 2; ++i) {
+    // if (MTI.getArgOperand(i) != origArg)
+    //   continue;
+    if (!MTI.getAttributes().hasParamAttr(i, "enzyme_type")) {
```
wsmoses, line 3456: This is wrong you're updating the first two args to be int if they have no type attr

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -3450,19 +3450,48 @@ void TypeAnalyzer::visitMemTransferCommon(llvm::CallBase &MTI) {
...
+      updateAnalysis(MTI.getArgOperand(i),
```
wsmoses, line 3457: The changes to TA should be reverrted and this should already be handled here: https://github.com/EnzymeAD/Enzyme/blob/9d6a86b46086a2009725f8bfb9af89e7dc7168f6/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L1242

The problem in the test case won't be fixed by TA [since it was written to mean TA wasn't enough for -- but the change needs to happen in visitMemTransfer to also parse the arg TT's and or it in

PR #2022, opened 2024-07-29
Add Rust Eprintln support

ZuseZ4:
The cxa_atexit is unneccessary, when trying to go that path I also had to add 
```
                         "pthread_key_delete",
                         "pthread_key_create",
```
and then ran into the issue that an atomic instruction wasn't supported by Enzyme.
Just recognizing eprintln on a higher level seems to be much simpler (and is probably slightly faster too).


ZuseZ4:
I can't provide an llvm reproducer, since despite my efforts there still are differences in the compilation pipelines between Rust-Enzyme and Rust->LLVM-IR ->(enzyme+opt). Thus I'm still getting this when going through opt. Or maybe it's not demanding correctly when used as a opt pass? Either way, since we don't have Rust integration tests here, I'll merge the minimal Rust test on the Rust side instead.
```
error: <unknown>:0:0: in function preprocess_eprintfunc void (): Enzyme: No create nofree of unknown value
@__cxa_thread_atexit_impl = extern_weak global i8
 at context:   call void @_ZN3std2io5stdio7_eprint17h557bd237e67376ddE(ptr noalias nocapture noundef nonnull align 8 dereferenceable(48) %1) #111
```

PR #2024, opened 2024-07-29
use correct NoFree list

ZuseZ4:
Looks like during (manual) upstreaming I added it to the wrong NoFree list, so my test was still not passing after updating the Enzyme submodule. This one should be the right noFree list.


wsmoses:
Okay, in that case revert the other one and I'm going to also insist a test be added then.

wsmoses (approved):

PR #2025, opened 2024-07-30
initial rust mpi support

ZuseZ4:


ZuseZ4:
We need to teach Enzyme to recognize
```
fn main( ) {
%5 = load ptr %MPI_SUM
%6 = call diffe_test_mpi_call(..., %5)
}

test_mpi_call(... ptr %5) {
%7 = call MPI_Allreduce(..., %5)
}
```
Because right now Enzyme yells about unknown mpi_allreduce op %5, I assume we already do that to e.g. recognize enyzme_const.

wsmoses:
See my comment on the other thread, but no that should not be the way it is handled. Speciically if it is a generic reduction op, the derivative code can similarly be arbitrary

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -145,9 +145,10 @@ class AdjointGenerator : public llvm::InstVisitor<AdjointGenerator> {
         C = CE->getOperand(0);
       }
       if (auto GV = dyn_cast<GlobalVariable>(C)) {
-        if (GV->getName() == "ompi_mpi_double") {
+        auto name = GV->getName();
```
wsmoses, line 148: add test cases for these

ZuseZ4: Can you provide me a link for where we read outside ptr values (and if it's the right approach)?
otherwise the testcase won't pass, see my error description above.

ZuseZ4: Done. I'll update it or add a second one once we found a solution for the reduction operator.

enzyme/test/Enzyme/ReverseMode/mpi_rust.ll
```diff
@@ -0,0 +1,115 @@
...
+define hidden noundef "enzyme_type"="{[-1]:Float@double}" double @_ZN10dot_enzyme12dot_parallel17h7dfcd86d9e8c176bE(ptr noalias nocapture noundef readonly align 8 dereferenceable(16) "enzyme_type"="{[-1]:Pointer}" %0, ptr noalias nocapture noundef nonnull readonly align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %1, i64 noundef "enzyme_type"="{[-1]:Integer}" %2, ptr noalias nocapture noundef nonnull readonly align 8 "enzyme_type"="{[-1]:Pointer, [-1,-1]:Float@double}" %3, i64 noundef "enzyme_type"="{[-1]:Integer}" %4, ptr noundef "enzyme_type"="{[0]:Pointer}" %5) unnamed_addr #1 personality ptr @rust_eh_personality {
+  %7 = alloca double, align 8
+  %8 = alloca double, align 8
+  call void @llvm.lifetime.start.p0(i64 8, ptr nonnull %8)
+  %9 = alloca double, align 8
+  store double 1.000, ptr %8, align 8
+  call void @llvm.lifetime.start.p0(i64 8, ptr nonnull %7)
+  store double 0.000000e+00, ptr %7, align 8
+  tail call void @llvm.experimental.noalias.scope.decl(metadata !7)
+  %10 = load ptr, ptr @RSMPI_DOUBLE, align 8, !noalias !10, !noundef !13
```
wsmoses, line 23: This is unused atm so this won't test any thing. Can you make this a minimal runnable case. Maybe a different mpi fn?

Also get rid of the other stuff like enzyme_type etc

ZuseZ4: @jedbrown I can't run things, can you test if some other function works on this branch?

enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -1150,7 +1150,8 @@ void AdjointGenerator::handleMPI(llvm::CallInst &call, llvm::Function *called,
           C = CE->getOperand(0);
         }
         if (auto GV = dyn_cast<GlobalVariable>(C)) {
-          if (GV->getName() == "ompi_mpi_op_sum") {
+          if (GV->getName() == "ompi_mpi_op_sum" ||
+              GV->getName() == "RSMPI_SUM") {
```
jedbrown, line 1154: I don't know how to resolve this.
```
source_id: DefId(0:12 ~ dot_enzyme[f53a]::dot_parallel)
num_fnc_args: 5
input_activity.len(): 5
error: <unknown>:0:0: in function preprocess__ZN10dot_enzyme12dot_parallel17hd37f1f8a2c8de07dE double (ptr, ptr, i64, ptr, i64): Enzyme: cannot compute with global variable that doesn't have marked shadow global
@RSMPI_SUM = external local_unnamed_addr global ptr
```

This is the relevant code: https://github.com/rsmpi/rsmpi/commit/840e01c9f772806a7b48fd82fcacba062a2c61eb#diff-9a676b0d0c142cd1e89e8174ddb007db982d8602bd374a04e40e9f6a421acaebR216-R228

Run with
```console
$ RUSTFLAGS='-Z unstable-options' cargo +enzyme r --example=dot_enzyme --release
```

ZuseZ4: Me neither, but can you share the module you got from ENZYME_OPT=1? Then I can experiment around to see if I find the right changes.

wsmoses: @jedbrown you need to add RSMPI_SUM to the ActivityAnalysis.cpp code [the message (poorly) warns that the global variable is differentiable, but Enzyme is unable to determine a differentiable version of the global. Of course it makes no sense to differentiate wrt MPI_SUM so we can mark that in ActivityAnalysis.cpp]

jedbrown: Thanks. Why didn't `ompi_mpi_op_sum` need to be there?

wsmoses: I think one has an extra pointer indirection causing a load which needs to be analyzed

jedbrown: We're now at this error.
```
source_id: DefId(0:12 ~ dot_enzyme[8922]::dot_parallel)
num_fnc_args: 5
input_activity.len(): 5
error: <unknown>:0:0: in function preprocess__ZN10dot_enzyme12dot_parallel17hf4631abf802872b2E double (ptr, ptr, i64, ptr, i64): Enzyme:  call:   %23 = call noundef i32 @MPI_Allreduce(ptr noundef nonnull %7, ptr noundef nonnull %6, i32 noundef 1, ptr noundef %9, ptr noundef %10, ptr noundef %22) #111, !noalias !800
 unhandled mpi_allreduce op:   %10 = load ptr, ptr @RSMPI_SUM, align 8, !noalias !803, !noundef !806
```

wsmoses: yeah in this case, now the issue is that the op of MPI_allreduce cannot be detected (the earlier check if the argument was a literal global no longer applies, since this is a load of RSMPI_SUM). Changing the MPI_Allreduce check to consider something along the lines of:

```
if (LI = dyn_cast<LoadInst>(...)))
  if (auto GV = dyn_cast<GlobalVariable>(LI->getPointerOperand()))
    if (GV->getName() == "RSMPI_SUM")
      legal = true;
```

jedbrown: I pushed that (assuming it was the correct place), but now I have
```
source_id: DefId(0:12 ~ dot_enzyme[42dc]::dot_parallel)
num_fnc_args: 5
input_activity.len(): 5
error: <unknown>:0:0: in function preprocess__ZN10dot_enzyme12dot_parallel17h2f3ed146b457ca09E double (ptr, ptr, i64, ptr, i64): Enzyme: cannot compute with global variable that doesn't have marked shadow global
@RSMPI_COMM_WORLD = external local_unnamed_addr global ptr
```

wsmoses: okay that's the same as the first issue [a global which cannot be proven non-differentiable]. ActivityAnalysis.cpp is again the right place to add that.

jedbrown: Thanks for the hand-holding. I think this output is correct now.
```console
$ RUSTFLAGS='-Z unstable-options' cargo +enzyme mpirun -n 2 --example=dot_enzyme --release | sort
[0] bx: [0.0, 2.0, 4.0, 6.0, 8.0], by: [0.0, 2.0, 4.0, 6.0, 8.0]
[0] local: 30
[1] bx: [200.0, 202.0, 204.0, 206.0, 208.0], by: [20.0, 22.0, 24.0, 26.0, 28.0]
[1] local: 6130
global: 6160
global: 6160
```

jedbrown: Note that `--release` is required, otherwise I see
```
invertedPointers:
   invertedPointers[ptr %0] =   <badref> = load ptr, ptr %4, align 8
end invertedPointers
  <badref> = load ptr, ptr %4, align 8
rustc: /home/jed/src/rust-enzyme/src/tools/enzyme/enzyme/Enzyme/GradientUtils.cpp:8489: virtual void InvertedPointerVH::deleted(): Assertion `0 && "erasing something in invertedPointers map"' failed.
error: could not compile `mpi` (example "dot_enzyme")
```

wsmoses:
I don’t think there’s a test in this PR, no? (From my earlier comment which doesn’t look resolved the llvm test doesn’t test this?)

jedbrown:
Is it appropriate to add a test that pulls in rsmpi? That would require building a Rust toolchain using this Enzyme and then `cargo test` (which would build rsmpi, assuming an MPI implementation is available somewhere in the CI environment). I'm supposing that "soon", people will be getting their Enzyme-enable Rust from rustup, and that will increase interest in support for differentiable rsmpi, etc.

Is there a stand-alone test we could use here instead?

ZuseZ4:
As a small note, I wouldn't recommend building rustc anymore, rustc is providing dist(ributable) builds.
Enzyme could download those, rm libenzyme.so and replace it by one that was just build on the latest commit.
The compiler explorer is already using the dist artifacts (for now I still create them manually), but I will deprecate those soon in favour of the rustc ones: https://github.com/EnzymeAD/rust/releases

wsmoses:
yeah we can pull in something. probably I'd recommend we make a separate repo for rust specific CI jobs (like how julia CI is inside of Enzyme.jl).

Before then, we should get these tests fixed up and merged ofc (which it looks like is just addressing some simplification comments above). Something like https://github.com/EnzymeAD/Enzyme/blob/8371758747eec72371626e719499d7bddc22dcad/enzyme/test/Enzyme/ReverseMode/mpi.ll#L26 but ofc with rsmpi (so we're not checking all of the other misc instructions)

PR #2028, opened 2024-08-01
add flag to surpress apple's -bundle flag

ZuseZ4:
Tested with another apple user. Apparently https://github.com/EnzymeAD/Enzyme/pull/1683 wasn't wrong after all, it just needs to be limited to apple and not run on Linux. Part of https://github.com/EnzymeAD/rust/issues/145 and the previous https://github.com/EnzymeAD/rust/pull/73 attempt.

cc @tgymnich, do you think that's something that can just be turned on by default for all Apple cases, or is it only good for the Rust Apple case, where we want to link libLLVMEnzyme-17.dylib in?


ZuseZ4:
Closed by Tim's solution using an existing flag in https://github.com/EnzymeAD/rust/pull/150

PR #2030, opened 2024-08-02
Memtransfer: consider explicit TA from dst ptr

wsmoses:


ZuseZ4:
Fixes the testcase and allows the Rust side now runs into a bug due to Enzyme not supporting llvm's add with overflow intrinsic, but that probably means the memcpy bug is now fixed, so let's merge it as a step forward.

ZuseZ4 (approved):

PR #2035, opened 2024-08-09
Update benchmarks

ZuseZ4:
Splitting this out of the rust-benchmarks to make it easier to upstream rust stuff on it's own.

closes #1414


enzyme/benchmarks/ReverseMode/ba/Makefile.make
```diff
@@ -16,8 +19,8 @@ clean:
 	opt $^ -o $@ -S
 	#opt $^ -O2 -o $@ -S
 
-ba.o: ba-opt.ll
-	clang++ -O2 $^ -o $@ $(BENCHLINK)
+ba.o: ba.cpp
+	clang++ $(LOAD) $(BENCH) ba.cpp -I /usr/include/c++/11 -I/usr/include/x86_64-linux-gnu/c++/11 -O2 -o ba.o -lpthread $(BENCHLINK) -lm -L /usr/lib/gcc/x86_64-linux-gnu/11
```
wsmoses, line 23: So as is this prevents from running the original enzyme ablation hypotheses, can you retain a way to do so?

ZuseZ4: Does the lstm example work for you? Then I can move all over to that.

enzyme/benchmarks/ReverseMode/lstm/Makefile.make
```diff
@@ -1,23 +1,23 @@
...
+	rm -f *.ll *.o results.json
 
 %-unopt.ll: %.cpp
-	clang++ $(BENCH) $^ -O2 -fno-vectorize -fno-slp-vectorize -ffast-math -fno-unroll-loops -o $@ -S -emit-llvm
-	#clang++ $(BENCH) $^ -O1 -Xclang -disable-llvm-passes -fno-vectorize -fno-slp-vectorize -ffast-math -fno-unroll-loops -o $@ -S -emit-llvm
+	clang++ $(BENCH) $^ -pthread -O2 -fno-vectorize -fno-slp-vectorize -ffast-math -fno-unroll-loops -o $@ -S -emit-llvm
 
 %-raw.ll: %-unopt.ll
-	opt $^ $(LOAD) -enzyme -o $@ -S
+	opt $^ $(LOAD) -o $@ -S
```
wsmoses, line 14: this should probably be restored to run Enzyme here, no?

ZuseZ4: I don't think so, check

> config.substitutions.append(('%enzyme', ('-enzyme' if int(config.llvm_ver) < 16 else '-passes="enzyme"')))

ZuseZ4:
@wsmoses 
```

CMake Error at /usr/lib/llvm-16/lib/cmake/llvm/AddLLVM.cmake:965 (add_executable):
  Target "enzyme-tblgen" links to target "zstd::libzstd_shared" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
Call Stack (most recent call first):
  /usr/lib/llvm-16/lib/cmake/llvm/TableGen.cmake:146 (add_llvm_executable)
  tools/enzyme-tblgen/CMakeLists.txt:9 (add_tablegen)

```
I won't look into such cmake issues, but if you're ok with emerging it despite those (or fixing them yourself), I can fix the unsupported flag combinations.

wsmoses:
Looks like an issue in setup for the script, can you copy whatever config from other CI ensures zlib is installed?

ZuseZ4:
The current status is that no benchmark CI is working: https://github.com/EnzymeAD/Enzyme/actions/runs/10607142537/job/29399074787?pr=2056
With this PR at least 4 Benchmarks are run under llvm-16, so I'd recommend to merge it.
Happy to disable the still broken benchmarks, so someone else can pick them up once they have time.

ZuseZ4:
lmk how to disable the benchmarks I haven't touched and which are thus still broken and this is good to merge and should even pass CI for 15 and 16.

enzyme/benchmarks/ReverseMode/adbench/gmm.h
```diff
@@ -168,8 +168,8 @@ int main(const int argc, const char* argv[]) {
     std::vector<std::string> paths;// = { "1k/gmm_d10_K100.txt" };
 
     getTests(paths, "data/1k", "1k/");
-    getTests(paths, "data/2.5k", "2.5k/");
-    getTests(paths, "data/10k", "10k/");
+    //getTests(paths, "data/2.5k", "2.5k/");
```
wsmoses, line 171: Why is this commented out?

wsmoses (commented):
So as is this removes the functionality for running the enzyme ablation hypothesis (by using the unified clang command to simplify rust linking).

can you add say another target for the reference pipeline which runs no optimization before enzyme AD?

enzyme/benchmarks/ReverseMode/adbench/gmm.h
ZuseZ4: Some of the test files for some benchmarks were commented out before due to runtime. 
I can drop this change here, but I'd suggest that we cut the number of tests that we run in CI
to keep CI time reasonable, and once we fixed all tests make it mandatory, to prevent this from regressing again.

ZuseZ4:
It .. doesn't, can you check again, or point out where exactly it uses ClangEnzyme? :D
This should be cosmetic and just makes sure that these tests here at least build again.
It still uses LLVM-Enzyme for opt. 
I sometimes define `LOADCLANG="%loadClangEnzyme"`, because the RustBench will use this for optimal C++ perf, but it's not used in this PR. I can remove it from this PR and reintroduce it in the next PR that includes Rust benches if you want?

wsmoses:
Ah it looks like it was just in a second makefile (I thought it was a sed script which modified this in place): https://github.com/EnzymeAD/Enzyme/blob/87ea5eb39315262930af8c5f01874921482484fb/enzyme/benchmarks/ReverseMode/ba/Makefile.makeafter#L13

Yeah this one should be modified too or ideally merged into the same makefile

enzyme/benchmarks/ReverseMode/adbench/gmm.h
```diff
@@ -168,8 +168,8 @@ int main(const int argc, const char* argv[]) {
...
+    //getTests(paths, "data/10k", "10k/");
```
wsmoses, line 172: Let's try to get this merged.

Here, can you use an environmental or command line variable to determine whether to run large tests or not?

wsmoses (approved):

enzyme/benchmarks/ReverseMode/ba/Makefile.make
```diff
@@ -3,21 +3,23 @@
...
-	opt $^ $(LOAD) $(ENZYME) -o $@ -S
+	$(OPT) $^ $(LOAD) $(ENZYME) -o $@ -S
 
 %-opt.ll: %-raw.ll
-	opt $^ -o $@ -S
+	$(OPT) $^ -o $@ -S
 
 ba.o: ba-opt.ll
-	clang++ $(BENCH) -pthread -O2 $^ -I /usr/include/c++/11 -I/usr/include/x86_64-linux-gnu/c++/11 -O2 -o $@ $(BENCHLINK) -lpthread -lm -L /usr/lib/gcc/x86_64-linux-gnu/11
+	$(CXX) $(BENCH) -pthread -O2 $^ -I /usr/include/c++/11 -I/usr/include/x86_64-linux-gnu/c++/11 -O2 -o $@ $(BENCHLINK) -lpthread -lm -L /usr/lib/gcc/x86_64-linux-gnu/11
```
jedbrown, line 22: Is this change welcome here? I'm running like `CXX=/home/jeka2967/src/rust/build/host/llvm/bin/clang++ OPT=/home/jeka2967/src/rust/build/host/llvm/bin/opt ninja bench-gmm-reverse` to use the same toolchain for Rust and C++. In principle this information could be based on `LLVM_DIR` in the CMake, but I don't know how to access that from this makefile and I think this flexibility is desirable.

ZuseZ4: CXX could cause issues with unhelpful error messages if people accidentally have GCC behind CXX, plus your call command is extremely verbose. I have 5 different  (incompatible) LLVM versions so I don't want to use export and right now running `ninja bench-gmm-reverse` seems to always pick the right one, so I'm hesitant to regress the usability so much. So if anything I would just change it to use the LLVM_DIR from cmake to automatically use the right one more confidently and leave it unchanged till then. I'll update the other benchmarks to include the env variable and update the Makefile.makeafter, unless you beat me to it.

jedbrown: If it's working for you now, then nothing will change with this patch. Perhaps the benchmarking system is setting `PATH` so that `clang++` is pulled from `$LLVM_DIR/bin`? I saw a failure due to what looked like mismatch between LLVM used to build Enzyme and the default `clang++` so I put this in and it passed. I didn't dig to see if that was really the cause.

wsmoses: so the path should be automatically set to point to the directory containing Enzyme's LLVM binaries. So it should always point to the right things, assuming there is something named clang there.

If we were to do this change, probably we should set it to the autopopulated path to llvm clang directly.
For example https://github.com/EnzymeAD/Enzyme/blob/fffdd2c49cd342d2ba1d4dff08bcd21db1390e58/enzyme/benchmarks/lit.site.cfg.py.in#L49 or https://github.com/EnzymeAD/Enzyme/blob/fffdd2c49cd342d2ba1d4dff08bcd21db1390e58/enzyme/test/lit.site.cfg.py.in#L71

wsmoses: here is where the path is set: https://github.com/EnzymeAD/Enzyme/blob/fffdd2c49cd342d2ba1d4dff08bcd21db1390e58/enzyme/benchmarks/lit.cfg.py#L37

jedbrown: Okay, thanks for pointing out where this is handled. I'm not reproducing the problem I saw earlier and `readelf -p .comment ba.o` reports the Rust toolchain (based on my `LLVM_DIR`) so it seems to be working. I'll drop this commit since it seems unnecessary.

enzyme/benchmarks/ReverseMode/adbench/gmm.h
jedbrown: This has been replaced by the `BENCH_LARGE` environment variable so I'll resolve this thread.

jedbrown:
I rebased this branch and will let CI run.

jedbrown:
I see the Bazel job is [failing on main](https://github.com/EnzymeAD/Enzyme/actions/runs/11584501877) too.

PR #2056, opened 2024-08-28
Update benchmark.yml

ZuseZ4:


ZuseZ4:
@tgymnich Do you know why CI still isn't running this?

tgymnich:
I don't think there are any openstack20 runners left. Try openstack22.

ZuseZ4:
@tgymnich Doesn't seem to help much, any other ideas? Iirc there were so discussions about only running them on releases, since they were so slow, could that be an issue?

tgymnich:
Workflow was disabled (in the GitHub UI under actions).

ZuseZ4:
Is there a reason why we were benchmarking debug builds? Can I remove that option?

wsmoses:
they currently all fail, if you can get them running before merging

wsmoses:
@ZuseZ4 following up from slack convo, I don't see any of the tests here as passing?

wsmoses:
In any case feel free to drop LLVM's before 15 for benchmarking, as we're going to do that shortly regardless

ZuseZ4:
closing this since the current code has too many bugs, in favor of #2035 where I have more fixes

PR #2059, opened 2024-08-30
MLIR: broadcast complex add and conj with linalg.map

Pangoraw:
`complex.add` and `complex.conj` do not broadcast like `arith.add` so when creating an add or conj or for a tensor of complex number, it needs element-wise broadcasting.

This currently implements broadcasting with a `linalg.map` which does not seem ideal. Maybe enzyme could emit `enzyme.add` instead and then consumers could lower to their own dialect when possibl. For example, `enzyme.add` to `stablehlo.add` in [Enzyme-JAX](https://github.com/EnzymeAD/Enzyme-jax) and still have the current lowering as a default? That would remove the need for arith raising in the case of Enzyme-JAX.


wsmoses:
at least for stablehlo I thought this was handled (indeed by having intermediate IR which was invalid, but then was fixed up in a subsequent pass).

Do you have an example which fails presently?

That said, I'm all for a better design. cc @ftynse for thoughts on.

Pangoraw:
> at least for stablehlo I thought this was handled

Trying again, i actually forgot to disable verification between each pass and run arith-raising. This is not needed.

PR #2062, opened 2024-08-30
Add header files to ClangEnzyme target

jandrej:


enzyme/Enzyme/CMakeLists.txt
```diff
@@ -208,6 +208,13 @@ target_compile_options(ClangEnzymeFlags INTERFACE "-fplugin=$<TARGET_FILE:ClangE
 add_custom_target(ClangEnzymeDummy "" DEPENDS ClangEnzyme-${LLVM_VERSION_MAJOR})
 add_dependencies(ClangEnzymeFlags ClangEnzymeDummy)
 
+set_property(TARGET ClangEnzyme-${LLVM_VERSION_MAJOR}
+    PROPERTY PUBLIC_HEADER
```
wsmoses, line 212: can we use a glob here to not need changing if another file is changed?

enzyme/Enzyme/CMakeLists.txt
```diff
@@ -208,6 +208,13 @@ target_compile_options(ClangEnzymeFlags INTERFACE "-fplugin=$<TARGET_FILE:ClangE
...
+    ${PROJECT_SOURCE_DIR}/include/enzyme/enzyme
```
wsmoses, line 213: also can we add a warning to the top of the file that changes in these installed files will not change the versions used by enzyme (which are internal to the compiler)

enzyme/Enzyme/CMakeLists.txt
jandrej: Sure

enzyme/Enzyme/CMakeLists.txt
jandrej: > to the top of the file

you mean the top of the line or the top of the include files?

jandrej:
We probably also need to change the capitalization to be consistent with the files. There's a clash between `enzyme` and `Enzyme` as the top directory at the moment.

enzyme/Enzyme/CMakeLists.txt
wsmoses: to the top of the include files

wsmoses:
> We probably also need to change the capitalization to be consistent with the files. There's a clash between `enzyme` and `Enzyme` as the top directory at the moment.

I don't quite follow here, the installed file should be in some /include dir no?

jandrej:
> > We probably also need to change the capitalization to be consistent with the files. There's a clash between `enzyme` and `Enzyme` as the top directory at the moment.
> 
> I don't quite follow here, the installed file should be in some /include dir no?

Yes they are in `include/Enzyme/$HEADER` while the internal files include from `include/enzyme/$HEADER`.

wsmoses:
Oh I see what you mean. Maybe we can move these files to
include/runtime/enzyme.

It doesn’t need to be on a valid include path anyways since clang auto
includes it.

So the question is just making sure the tools are happy.

If that doesn’t work, then yeah let’s find a good way to fix this properly

jandrej:
It depends on what you want it to look like from the source level where the files are included. Doing
```
#include <enzyme/enzyme>
```
seems like a clean solution to me (which would require only renaming the install directory)

wsmoses:
Isn’t that what the user presently needs to do, or am I
misremembering/misunderstanding?

jandrej:
They should, but no headers were installed previously. Also if they are installed they're in `Enzyme/enzyme` which isn't portable in terms of capitalization.

wsmoses:
we don't presently install any "Enzyme" (the library) headers, iirc -- so can we not have that directory, and only the runtime one?

jandrej:
To clarify, current main branch has this https://github.com/EnzymeAD/Enzyme/blob/b006eb21388dc1f0c13e7f4584847805767f95aa/enzyme/Enzyme/CMakeLists.txt#L214-L218 where `PUBLIC_HEADER DESTINATION` is wrong. The changes in this PR don't address this yet.

wsmoses (approved):
sgtm

jandrej:
ping @wsmoses

ZuseZ4:
@wsmoses that won't work, since there are outdated required checks that won't run. You'll need to merge manually.

jandrej:
ping @wsmoses

PR #2074, opened 2024-09-10
Set minimum LLVM to 15+

wsmoses:


vchuravy:
Need to also update https://github.com/EnzymeAD/Enzyme/blob/d6d36d6cabb37888fadae5ab1bde116ba6b0ec1d/.packaging/build_tarballs.jl#L170

https://github.com/JuliaRegistries/General/pull/114874

PR #2085, opened 2024-09-26
Add debug info to fprt runtime calls

ivanradanov:


enzyme/Enzyme/Runtimes/FPRT/GarbageCollection.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include <array>
+#include <cstdio>
+#include <cstdlib>
+#include <iostream>
+#include <list>
+#include <llvm/Support/ErrorHandling.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#define ENZYME_ENABLE_GARBAGE_COLLECTION
```
wsmoses, line 39: not necessary pre pldi, but llvm / enzyme doesn't need to be built on a system with mpfr at the moment [and thus CI fails]

PR #2114, opened 2024-10-09
[mlir] Summary-based activity analysis

pengmai:


enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
```diff
@@ -0,0 +1,1319 @@
...
+void printSetLattice(const enzyme::SparseSetLattice<ValueT> &setLattice,
+                     raw_ostream &os) {
+  if (setLattice.isUnknown()) {
+    os << "Unknown Origin";
+  } else if (setLattice.isUndefined()) {
+    os << "Undefined Origin";
+  } else {
+    os << "size: " << setLattice.getElements().size() << ":\n";
+    for (auto element : setLattice.getElements()) {
+      os << "  " << element << "\n";
```
ftynse, line 35: Nit: it would be helpful to have these sorted in a consistent way, e.g., by printing each element to a string via `raw_string_ostream` first, then sorting a vector of strings before printing them to `os`.

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
```diff
@@ -0,0 +1,1319 @@
...
+  return isa<LLVM::FMulOp, LLVM::FAddOp, LLVM::FDivOp, LLVM::FSubOp,
+             LLVM::FNegOp, LLVM::FAbsOp, LLVM::SqrtOp, LLVM::SinOp, LLVM::CosOp,
+             LLVM::Exp2Op, LLVM::ExpOp, LLVM::LogOp, LLVM::InsertValueOp,
+             LLVM::ExtractValueOp, LLVM::BitcastOp, LLVM::SelectOp>(op);
+}
+
+static bool isNoOp(Operation *op) {
+  return isa<LLVM::NoAliasScopeDeclOp, LLVM::LifetimeStartOp,
+             LLVM::LifetimeEndOp, LLVM::AssumeOp>(op);
+}
```
ftynse, line 87: I suspect we have a copy of this elsewhere and would rather have one definition that is used in both, e.g., by having a static inline function in a header. Otherwise we run the risk of forgetting to update one of the copies and wondering why it doesn't work.

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
```diff
@@ -0,0 +1,1319 @@
...
+  return success();
+}
+
+void enzyme::ForwardActivityAnnotationAnalysis::processMemoryRead(
+    Operation *op, Value address, ArrayRef<ForwardOriginsLattice *> results) {
+  auto markResultsUnknown = [&]() {
+    for (ForwardOriginsLattice *result : results) {
+      propagateIfChanged(result, result->markUnknown());
+    }
+  };
```
ftynse, line 139: Nit: this is the same lambda as above, promote to a class function?

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
```diff
@@ -0,0 +1,1319 @@
...
+//   auto strB = dyn_cast_if_present<StringAttr>(distinctB.getReferencedAttr());
+//   if (pseudoA && pseudoB) {
+//     return std::make_pair(pseudoA.getArgNumber(), pseudoA.getDepth()) <
+//            std::make_pair(pseudoB.getArgNumber(), pseudoB.getDepth());
+//   } else if (strA && strB) {
+//     return strA.strref() < strB.strref();
+//   }
+//   // Order pseudo/origin classes before fresh classes
+//   return (pseudoA || originA) && !(pseudoB || originB);
+// }
```
ftynse, line 429: dead code

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
```diff
@@ -0,0 +1,1319 @@
...
+  }
+  propagateIfChanged(lattice, changed);
+}
+
+std::optional<Value> getStored(Operation *op);
+std::optional<Value> getCopySource(Operation *op);
+
+LogicalResult enzyme::DenseActivityAnnotationAnalysis::visitOperation(
+    Operation *op, const ForwardOriginsMap &before, ForwardOriginsMap *after) {
+  join(after, before);
```
ftynse, line 454: Comment, no need for action: this is one of the reasons why the analysis may be inefficient. Specifically, this join subsumes `propagateIfChanged`, and we will call it again below. This may schedule two updates when one would have been sufficient.

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
```diff
@@ -0,0 +1,1319 @@
...
+      while (depth > 0) {
+        AliasClassSet next;
+        if (current.isUndefined()) {
+          // TODO: what should be done here?
+          // Activity annotations requires converged pointer info. If we have
+          // incomplete points-to info, we can't currently tell if it's because
+          // the points-to info hasn't _yet_ been computed (in which case we
+          // bail out here expecting to be called again with more complete
+          // points-to info), or if the points-to info has converged, this
+          // signifies reading from uninitialized memory.
```
ftynse, line 646: Leverage undefined behavior. The compiler assumes undefined behavior, such as reading from uninitialized memory, never happens.

ftynse (approved):
Looks reasonable overall, please give @wsmoses  a chance to take a look

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
```diff
@@ -0,0 +1,1319 @@
...
+  auto funcOp = cast<FunctionOpInterface>(arg.getOwner()->getParentOp());
+  auto origin = ArgumentOriginAttr::get(FlatSymbolRefAttr::get(funcOp),
+                                        arg.getArgNumber());
+  return propagateIfChanged(
+      lattice, lattice->join(ForwardOriginsLattice::single(lattice->getAnchor(),
+                                                           origin)));
+}
+
+/// True iff all results differentially depend on all operands
+// TODO: differential dependency/activity interface
```
wsmoses, line 75: So we do now have such an interface: https://github.com/EnzymeAD/Enzyme/blob/716d674d012834e6a1da0ad5febf9194e20aa868/enzyme/Enzyme/MLIR/Analysis/ActivityAnalysis.cpp#L1001

wsmoses (commented):
Could a relevant test be added to make sure it doesn’t break?

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
pengmai: I realized that the consistent order via sorting was already implemented for the `serialize` method. The `print` method isn't actually used for any tests (it's just required by the framework), so we can re-use the `serialize` method and print out the serialized Attributes rather than duplicating the sorting logic.

wsmoses:
@pengmai I'm going to merge for now to avoid rebase issues, but comments above should be addressed subsequently

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
pengmai: I've replaced this function with using the `ActivityOpInterface`.

enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
pengmai: I think getting the `ADDataFlowOpInterface` to work with the set of fully active ops that the benchmarks require will be a nontrivial amount of work (possibly involving tablegen changes), so I think it'd make sense to leave that for a future PR.

PR #2124, opened 2024-10-18
Fixes warnings and adds missing header guards

MattBolitho:
This change should produce no warnings when building with the `all` target with `clang-19`

### DIFF
--- a/enzyme/tools/enzyme-tblgen/blas-tblgen.cpp
+++ b/enzyme/tools/enzyme-tblgen/blas-tblgen.cpp
@@ -739,7 +739,7 @@ void emit_extract_calls(const TGPattern &pattern, raw_ostream &os) {
 }
 
 // Will be used by Julia
-SmallString<80> ValueType_helper(const TGPattern &pattern, ssize_t actPos,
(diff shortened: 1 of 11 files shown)


enzyme/tools/enzyme-tblgen/blas-tblgen.cpp
```diff
@@ -739,7 +739,7 @@ void emit_extract_calls(const TGPattern &pattern, raw_ostream &os) {
 }
 
 // Will be used by Julia
-SmallString<80> ValueType_helper(const TGPattern &pattern, ssize_t actPos,
+SmallString<80> ValueType_helper(const TGPattern &pattern, size_t actPos,
```
wsmoses, line 742: This isn't valid since actpos could be negative

wsmoses (commented):
Generally looks good, but fix the ssize_t part

enzyme/tools/enzyme-tblgen/blas-tblgen.cpp
MattBolitho: Seemed to not flag after reverting

wsmoses (approved):

PR #2127, opened 2024-10-21
mlir: Func call reverse diff

Pangoraw:
prompted by https://github.com/EnzymeAD/Reactant.jl/pull/171#issuecomment-2420942682.

We can now run enzyme-batch -> enzyme-wrap without inlining.

```mlir
// ./bazel-bin/enzymexlamlir-opt test/lit_tests/batchtests/autodiff.mlir --enzyme-batch --enzyme-wrap="infn=main retTys=enzyme_active argTys=enzyme_active mode=ReverseModeCombined" --canonicalize --remove-unnecessary-enzyme-ops --arith-raise --cse --canonicalize --arith-raise --enzyme-simplify-math --mlir-print-ir-before=enzyme-batch
// -----// IR Dump Before BatchPass (enzyme-batch) //----- //
module {
  func.func @f(%arg0: tensor<f32>) -> tensor<f32> {
    %0 = stablehlo.multiply %arg0, %arg0 : tensor<f32>
    return %0 : tensor<f32>
  }
  func.func @main(%arg0: tensor<10xf32>) -> tensor<10xf32> {
    %0 = enzyme.batch @f(%arg0) {batch_shape = array<i64: 10>} : (tensor<10xf32>) -> tensor<10xf32>
    return %0 : tensor<10xf32>
  }
}


module {
  func.func @f(%arg0: tensor<f32>) -> tensor<f32> {
    %0 = stablehlo.multiply %arg0, %arg0 : tensor<f32>
    return %0 : tensor<f32>
  }
  func.func private @batched_f(%arg0: tensor<10xf32>) -> tensor<10xf32> {
    %0 = stablehlo.multiply %arg0, %arg0 : tensor<10xf32>
    return %0 : tensor<10xf32>
  }
  func.func @main(%arg0: tensor<10xf32>, %arg1: tensor<10xf32>) -> tensor<10xf32> {
    %cst = arith.constant dense<0.000000e+00> : tensor<10xf32>
    %0 = call @batched_f(%arg0) : (tensor<10xf32>) -> tensor<10xf32>
    %1 = stablehlo.add %arg1, %cst : tensor<10xf32>
    %2 = call @diffebatched_f(%arg0, %1) : (tensor<10xf32>, tensor<10xf32>) -> tensor<10xf32>
    %3 = stablehlo.add %2, %cst : tensor<10xf32>
    return %3 : tensor<10xf32>
  }
  func.func private @diffebatched_f(%arg0: tensor<10xf32>, %arg1: tensor<10xf32>) -> tensor<10xf32> {
    %cst = arith.constant dense<0.000000e+00> : tensor<10xf32>
    %0 = stablehlo.add %arg1, %cst : tensor<10xf32>
    %1 = stablehlo.multiply %0, %arg0 : tensor<10xf32>
    %2 = stablehlo.add %1, %cst : tensor<10xf32>
    %3 = stablehlo.add %2, %1 : tensor<10xf32>
    return %3 : tensor<10xf32>
  }
}

```

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
@@ -29,10 +29,10 @@ namespace {
 #include "Implementations/FuncDerivatives.inc"
 } // namespace
 
-static std::optional<func::FuncOp> getContainingFunction(Operation *orig) {
+static std::optional<mlir::FunctionOpInterface> getContainingFunction(Operation *orig) {
   Operation *parent;
   while (parent = orig->getParentOp()) {
-    if (auto func = dyn_cast<func::FuncOp>(parent)) {
+    if (auto func = dyn_cast<mlir::FunctionOpInterface>(parent)) {
       return std::optional(func);
     }
   }
@@ -55,7 +55,7 @@ public:
 
     auto parent = getContainingFunction(orig);
     if (parent.has_value() &&
-        callOp.getCallee() == parent.value().getSymName()) {
+        callOp.getCallee() == parent.value().getNameAttr()) {
       // TODO: Recursion chains
       orig->emitError() << "could not emit adjoint of recursive call at: "
                         << *orig << "\n";
--- a/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
@@ -68,13 +68,20 @@ public:
     auto narg = orig->getNumOperands();
     auto nret = orig->getNumResults();
 
-    std::vector<DIFFE_TYPE> RetActivity(nret, DIFFE_TYPE::OUT_DIFF);
+    std::vector<DIFFE_TYPE> RetActivity;
+    for (auto res : callOp.getResults()) {
+      RetActivity.push_back(gutils->isConstantValue(res)
+                                ? DIFFE_TYPE::CONSTANT
+                                : DIFFE_TYPE::OUT_DIFF);
+    }
 
     std::vector<DIFFE_TYPE> ArgActivity;
     for (auto arg : callOp.getOperands()) {
-      ArgActivity.push_back(gutils->isConstantValue(arg)
-                                ? DIFFE_TYPE::CONSTANT
-                                : DIFFE_TYPE::OUT_DIFF);
+      ArgActivity.push_back(
+          gutils->isConstantValue(arg) ? DIFFE_TYPE::CONSTANT
+          : arg.getType().cast<AutoDiffTypeInterface>().isMutable()
+              ? DIFFE_TYPE::DUP_ARG
+              : DIFFE_TYPE::OUT_DIFF);
     }
 
     std::vector<bool> volatile_args(narg, true);
@@ -98,6 +105,8 @@ public:
     }
 
     for (auto result : callOp.getResults()) {
+      if (gutils->isConstantValue(result))
+        continue;
       revArguments.push_back(gutils->diffe(result, builder));
     }
 
--- a/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
@@ -29,7 +29,8 @@ namespace {
 #include "Implementations/FuncDerivatives.inc"
 } // namespace
 
-static std::optional<mlir::FunctionOpInterface> getContainingFunction(Operation *orig) {
+static std::optional<mlir::FunctionOpInterface>
+getContainingFunction(Operation *orig) {
   Operation *parent;
   while (parent = orig->getParentOp()) {
     if (auto func = dyn_cast<mlir::FunctionOpInterface>(parent)) {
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
```diff
@@ -29,9 +29,117 @@ namespace {
...
+      return failure();
+    }
+
+    Operation *callee = symbolTable.lookup(callOp.getCallee());
+    auto fn = cast<FunctionOpInterface>(callee);
+
+    auto narg = orig->getNumOperands();
+    auto nret = orig->getNumResults();
+
+    std::vector<DIFFE_TYPE> RetActivity(nret, DIFFE_TYPE::OUT_DIFF);
```
wsmoses, line 71: This should probably be populated with the activity results by calling isConstantValue on all the result operands

enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
```diff
@@ -29,9 +29,117 @@ namespace {
...
+
+    std::vector<DIFFE_TYPE> ArgActivity;
+    for (auto arg : callOp.getOperands()) {
+      ArgActivity.push_back(gutils->isConstantValue(arg)
+                                ? DIFFE_TYPE::CONSTANT
+                                : DIFFE_TYPE::OUT_DIFF);
```
wsmoses, line 77: Active or dup can be determined by casting to the AutoDiffTypeInterface and asking if the type is mutable

enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
```diff
@@ -29,9 +29,117 @@ namespace {
...
+    }
+
+    std::vector<bool> volatile_args(narg, true);
+    std::vector<bool> returnShadow(narg, false);
+    std::vector<bool> returnPrimal(nret, false);
```
wsmoses, line 82: We should probably default to always returning the primal, no?

enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
```diff
@@ -29,9 +29,117 @@ namespace {
...
+
+    auto type_args = gutils->TA.getAnalyzedTypeInfo(fn);
+
+    bool freeMemory = true;
+    size_t width = 1;
+
+    auto revFn = gutils->Logic.CreateReverseDiff(
```
wsmoses, line 89: Long term We probably need to do the same augmented forward pass and separate reverse pass that enzyme llvm does atm .

if we assume everything in the function is read only and only return active not duplicated results this is fine

wsmoses (approved):
LGTM but some misc fix ups and also the assumptions of the current implementation should throw errors if not met (we have a isReadOnly for example)

enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
Pangoraw: ah right, the caching at the primal call site is not enough if the arguments are mutable. I could not make an example input with memrefs (something about invertPointerM). So I added an error if any of the result or arg of the initial call is mutable.

enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
wsmoses: Yeah we would only need for fused forward and reverse passes or handling of mutation, we can defer that for later here

enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
```diff
@@ -29,9 +29,139 @@ namespace {
 #include "Implementations/FuncDerivatives.inc"
 } // namespace
 
+static std::optional<func::FuncOp> getContainingFunction(Operation *orig) {
+  Operation *parent;
```
wsmoses, line 33: Can this be a function interface instead?

enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
```diff
@@ -29,9 +29,140 @@ namespace {
...
+                                         SmallVector<Value> caches) const {
+    DerivativeMode mode = DerivativeMode::ReverseModeGradient;
+
+    SymbolTable symbolTable = SymbolTable::getNearestSymbolTable(orig);
+
+    func::CallOp callOp = cast<func::CallOp>(orig);
+
+    auto parent = getContainingFunction(orig);
+    if (parent.has_value() &&
+        callOp.getCallee() == parent.value().getNameAttr()) {
```
wsmoses, line 59: Ah I see this use, this is super fragile. For example it won’t work if you have a calls b calls a.

I think this is okay to merge as is but adding proper recursion support shouldn’t be bad. We have it already for forward enzymemlir which basically just requires having a cache in enzymelogic asking if we’ve already made this derivative function

Pangoraw: yeah any cycle in the call graph will not be caught here. also it would be nice if we could somehow store the symbol corresponding to the derivative of that call (with corresponding activities) to call the same reverse function in subsequent calls.

wsmoses: Yeah enzymelogic should do this for forwards (and also llvm) already

Pangoraw: I will look at implementing something like ForwardCachedFunctions for reverse in a follow-up

PR #2133, opened 2024-10-24
mlir: cache and reuse reverse funcs

Pangoraw:


enzyme/Enzyme/MLIR/Interfaces/EnzymeLogicReverse.cpp
```diff
@@ -189,13 +189,28 @@ FunctionOpInterface MEnzymeLogic::CreateReverseDiff(
     llvm_unreachable("Differentiating empty function");
   }
 
+  MReverseCacheKey tup = {
+      fn,        retType,
+      constants, returnPrimals,
```
wsmoses, line 194: you should also have returnShadows, reeMemory, volatile_args as part of the key

wsmoses (approved):

PR #2142, opened 2024-11-02
mlir: Modify `BatchOpInterface` to allow generating more than one operation

Pangoraw:
This PR modifies the `BatchOpInterface` to enable generating multiple operations and then map from the original results to new values. To do this, `batchSizes` support is removed in the original Enzyme MLIR cloning and batch cloning is implemented on its own.

See https://github.com/EnzymeAD/Enzyme-JAX/pull/149#issuecomment-2450019084 for more context.


wsmoses (approved):

wsmoses:
@Pangoraw I think you need to fix the clang-format for it to pass

PR #2169, opened 2024-11-17
Adds initial CMake presets configuration

### LINKED ISSUE #2162: Use of CMake Presets
I was wondering if consideration had been given to the use of [CMake Presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html). They improve a few aspects of the CMake experience like editor integration and sharing prefab configurations which users can extend with their own custom setups. For instance, I think it could be useful for #1607,  where it seems like MSVC specific configurations would be useful.

For some demonstration, I have [a branch in my fork](https://github.com/MattBolitho/Enzyme/tree/cmake-presets) where they work to a basic extent with parity with the current configuration options from the Enzyme root CMake project. It needs some tidying, but hopefully it is illustrative enough of the general ideas.

MattBolitho:
Adds CMake presets that have parity with the existing configuration. This will hopefully make it easier for people to extend/customize Enzyme builds on their platform.

No real changes to the build system are expected because the presets mirror the existing configuration. It's hopefully something that can be refined and extended in future, but works fine for the sake of an initial PR to make it available for basic builds.

Closes #2162

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ enzyme/benchmarks/ReverseMode/*/results.txt
 enzyme/benchmarks/ReverseMode/*/results.json
 .cache
 CMakeUserPresets.json
+/out
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ enzyme/benchmarks/ReverseMode/*/*.exe
 enzyme/benchmarks/ReverseMode/*/results.txt
 enzyme/benchmarks/ReverseMode/*/results.json
 .cache
+CMakeUserPresets.json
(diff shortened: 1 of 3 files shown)


ZuseZ4:
Would you mind to also add a note about these presets here (or in one of the related pages), so users are aware of this option? 
https://enzyme.mit.edu/Installation
Docs live in the EnzymeAD/www repo.

MattBolitho:
> Would you mind to also add a note about these presets here (or in one of the related pages), so users are aware of this option? 
> https://enzyme.mit.edu/Installation
> Docs live in the EnzymeAD/www repo.

Great idea and thanks for the info. I will try to get around to it soon!

PR #2181, opened 2024-11-28
Batched autodiff

jumerckx:
Added some type conversions to tensor types if `width != 1`. The simple test case seems correct now.
Corresponding Enzyme-JAX PR: https://github.com/EnzymeAD/Enzyme-JAX/pull/197

### DIFF
--- a/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
+++ b/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
@@ -194,17 +194,17 @@ def GenericAdjointOp : Enzyme_Op<"genericAdjoint", [AttrSizedOperandSegments]> {
 
 def BroadcastOp : Enzyme_Op<"broadcast"> {
   let description = [{
-  Broadcast the operand by adding an extra dimension the frond with a size equal to the width attribute.
-  For scalar operands, a one-dimensional ranked tensor is created.
+  Broadcast the operand by adding extra dimensions with sizes provided by the `shape` attribute to the front.
+  For scalar operands, ranked tensor is created.
 
-  NOTE: Only works for scalars and *ranked* tensors for now.
+  NOTE: Only works for scalar and *ranked* tensor operands for now.
   }];
 
-  let arguments = (ins AnyType:$input, I64Attr:$width);
+  let arguments = (ins AnyType:$input, DenseI64ArrayAttr:$shape);
   let results = (outs AnyRankedTensor:$output);
 
   let builders = [
-    OpBuilder<(ins "Value":$input, "int64_t":$width)>
+    OpBuilder<(ins "Value":$input, "llvm::SmallVector<int64_t>":$shape)>
   ];
 }
 
--- a/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
+++ b/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
@@ -204,7 +204,7 @@ def BroadcastOp : Enzyme_Op<"broadcast"> {
   let results = (outs AnyRankedTensor:$output);
 
   let builders = [
-    OpBuilder<(ins "Value":$input, "llvm::SmallVector<int64_t>":$shape)>
+    OpBuilder<(ins "Value":$input, "ArrayRef<int64_t>":$shape)>
   ];
 }
 
--- a/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
+++ b/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
@@ -192,4 +192,20 @@ def GenericAdjointOp : Enzyme_Op<"genericAdjoint", [AttrSizedOperandSegments]> {
   
 }
 
+def BroadcastOp : Enzyme_Op<"broadcast"> {
+  let description = [{
+  Broadcast the operand by adding an extra dimension the frond with a size equal to the width attribute.
+  For scalar operands, a one-dimensional ranked tensor is created.
+
+  NOTE: Only works for scalars and *ranked* tensors for now.
+  }];
+
+  let arguments = (ins AnyType:$input, I64Attr:$width);
+  let results = (outs AnyRankedTensor:$output);
+
+  let builders = [
+    OpBuilder<(ins "Value":$input, "int64_t":$width)>
+  ];
+}
+
 #endif // ENZYME_OPS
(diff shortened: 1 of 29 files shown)


enzyme/Enzyme/MLIR/Interfaces/CloneFunction.cpp
```diff
@@ -27,7 +27,11 @@ getFunctionTypeForClone(mlir::FunctionType FTy, DerivativeMode mode,
   for (auto &&[Ty, returnPrimal, returnShadow, activity] : llvm::zip(
            FTy.getResults(), returnPrimals, returnShadows, ReturnActivity)) {
     if (returnPrimal) {
-      RetTypes.push_back(Ty);
+      if (width != 1) {
+        RetTypes.push_back(mlir::RankedTensorType::get({width}, Ty));
```
wsmoses, line 31: This shouldn’t need changing since the primal is always unmodified, only Derivatives are changed (and we should be pushing the getshadow types for those below)

jumerckx: Oh, then I'm confused of what batched autodiff is.
How should my testcase change?

jumerckx: Nvm, it clicked. It's just the shadow that's batched 😅

wsmoses: so here's an example from llvm vector mode for example: https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/test/Enzyme/ForwardModeVector/add.ll

wsmoses: tho perhaps mul will be more illustrative, https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/test/Enzyme/ForwardModeVector/mul.ll (and obviously feel free to look at any/all of the other examples

jumerckx:
I haven't yet fully made the changes in enzyme-tblgen.cpp, and either way this just works for the simple test case.
But I added the following manually in `ArithDerivatives.inc`.
```cpp
mlir::Value itmp = ({
  // Computing MulFOp
  auto fwdarg_0 = dif;
  auto fwdarg_1 = gutils->getNewFromOriginal(op->getOperand(1));
  if (gutils->width != 1)
  {
    fwdarg_1 = builder.create<tensor::SplatOp>(
        op.getLoc(),
        mlir::RankedTensorType::get({gutils->width},
                                    fwdarg_1.getType()),
        fwdarg_1);
  }
  builder.create<arith::MulFOp>(op.getLoc(), fwdarg_0, fwdarg_1);
});
```

But this is the MLIR code that is generated for this simple test:
```mlir
  func.func private @fwddiffe2square(%arg0: f64, %arg1: tensor<2xf64>) -> tensor<2xf64> {
    %splat = tensor.splat %arg0 : tensor<2xf64>
    %0 = arith.mulf %arg1, %splat : tensor<2xf64>
    %splat_0 = tensor.splat %arg0 : tensor<2xf64>
    %1 = arith.mulf %arg1, %splat_0 : tensor<2xf64>
    %2 = arith.addf %0, %1 : tensor<2xf64>
    %3 = arith.mulf %arg0, %arg0 : f64
    return %2 : tensor<2xf64>
  }
```

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -192,4 +192,20 @@ def GenericAdjointOp : Enzyme_Op<"genericAdjoint", [AttrSizedOperandSegments]> {
...
 
+def BroadcastOp : Enzyme_Op<"broadcast"> {
+  let description = [{
+  Broadcast the operand by adding an extra dimension the frond with a size equal to the width attribute.
+  For scalar operands, a one-dimensional ranked tensor is created.
+
+  NOTE: Only works for scalars and *ranked* tensors for now.
+  }];
+
+  let arguments = (ins AnyType:$input, I64Attr:$width);
```
wsmoses, line 203: To support the Enzyme.batch [which is a bit more general since it takes a shape, not just a single int], do we want to make this a vararg of i64 ?

wsmoses (approved):

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -192,4 +192,20 @@ def GenericAdjointOp : Enzyme_Op<"genericAdjoint", [AttrSizedOperandSegments]> {
...
+  For scalar operands, ranked tensor is created.
+
+  NOTE: Only works for scalar and *ranked* tensor operands for now.
+  }];
+
+  let arguments = (ins AnyType:$input, DenseI64ArrayAttr:$shape);
+  let results = (outs AnyRankedTensor:$output);
+
+  let builders = [
+    OpBuilder<(ins "Value":$input, "llvm::SmallVector<int64_t>":$shape)>
```
wsmoses, line 207: can you change this to ArrayRef<int64_t> (which is a more general base class)

enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -191,3 +191,23 @@ LogicalResult BatchOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
 }
+
+//===----------------------------------------------------------------------===//
+// BroadcastOp
+//===----------------------------------------------------------------------===//
+
+void BroadcastOp::build(OpBuilder &builder, OperationState &result, Value input, llvm::SmallVector<int64_t> shape) {
+  auto shapeAttr = builder.getDenseI64ArrayAttr(shape);
+  RankedTensorType output;
+  // TODO: support things other than scalars and ranked tensors, maybe reuse getShadowType here?
```
wsmoses, line 202: yeah long term we should probably just use https://github.com/EnzymeAD/Enzyme/blob/534a28595a9bf1f41d4a8222615f00e36611dc0f/enzyme/Enzyme/MLIR/Interfaces/AutoDiffTypeInterface.td#L59

so essentially do

auto ty = input.getType();
for (auto s : reverse(shape)) {
  ty = ty.cast<AutodiffTypeInterface>().getShadowType(s)
}

wsmoses: but for now this is fine

wsmoses (approved):

wsmoses:
fix the format/etc then I think this is good to go!

jumerckx:
@wsmoses I started looking into control flow (`cf` and `scf`) but can move these commits out if you'd like to merge without those changes first.

wsmoses:
That’s totally fine/go for it, but also feel free to merge whenever it’s in a good state

PR #2189, opened 2024-12-06
custom derivative doc prototype

ZuseZ4:


enzyme/Enzyme/CApi.cpp
```diff
@@ -340,6 +340,21 @@ void EnzymeRegisterAllocationHandler(char *Name, CustomShadowAlloc AHandle,
     };
 }
 
+
+/// This is the main entry point to register a custom derivative for language frontends.
```
wsmoses, line 344: I probably wouldn't put this here, can you put the docs [here](https://github.com/EnzymeAD/Enzyme/blob/31efe088a2c5d82f9f48323fee6e01fb3ecd5ef0/enzyme/Enzyme/GradientUtils.h#L87)  and cross reference/link it from here?

enzyme/Enzyme/CApi.cpp
```diff
@@ -340,6 +340,21 @@ void EnzymeRegisterAllocationHandler(char *Name, CustomShadowAlloc AHandle,
...
+/// It should be prefered over trying to register custom-derivatives in the llvm-ir module.
```
wsmoses, line 345: I wouldn't say necessarily that it's preferred, but it's certainly much more general. It supports arbitrary LLVM code injection, with full access to all Enzyme and LLVM analyses (such as activity analysis, differential use analysis, alias analysis, etc)

enzyme/Enzyme/CApi.cpp
```diff
@@ -340,6 +340,21 @@ void EnzymeRegisterAllocationHandler(char *Name, CustomShadowAlloc AHandle,
...
+/// The main reason is that these rules can handle non-default activity cases, e.g.
+/// a function call where a pointer or a float scalar is marked as const.
+/// To get a better low-level understanding, the code in AdjointGenerator.h can be read.
+///
+/// This Function will only handle using ReverseMode AD (either split or combined).
```
wsmoses, line 350: currently this only supports split mode. Well kind of. A combined mode call will call the "forward pass" generator from here and then also the "reverse pass" generator as well. This is also the case for  split mode as well. Of course with the full functionality of gradientutils/etc available, one could check if they are in combined mode by checking if gutils->mode == ReverseModeCombined.

enzyme/Enzyme/CApi.cpp
```diff
@@ -340,6 +340,21 @@ void EnzymeRegisterAllocationHandler(char *Name, CustomShadowAlloc AHandle,
...
+/// As a high-level example, assume we want to register a custom derivative for a vector resize function.
```
wsmoses, line 351: I don't think this is a good example (far too complicated, and I'm not confident your algirthm for the derivative is correct). Perhaps explain with a simple pow(x, y) example where x or y could be differentiable or not and you want to handle both easily.

enzyme/Enzyme/CApi.cpp
```diff
@@ -363,6 +378,13 @@ void EnzymeRegisterCallHandler(char *Name,
   };
 }
 
+/// This is the main entry point to register a custom derivative for language frontends.
```
wsmoses, line 381: Same comments as above

enzyme/Enzyme/CApi.cpp
ZuseZ4: Can you please actually explain why it is not correct?

enzyme/Enzyme/CApi.cpp
ZuseZ4: Can you please answer how the last three arguments should be used?

ZuseZ4:
Also, let us not loose the original focus due to which I started looking into documenting this. Under release mode, high-level vec resize functions disappear, as visible here: https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=a68343446d1e5d79183603a1eb3b089f
You'll see a call to `alloc::raw_vec::RawVecInner<A>::reserve::do_reserve_and_handle` which takes the original pointer and the original len of the primal vec. I *think* gutils can give us the shadow pointer, but not the shadow len, correct?
In that case, it seems hardly possible to define a custom-derivative for it (since it would be a lot of effort to find the shadow len). So before continuing here, I would need to mark one of the more high-level functions as inline-never?

enzyme/Enzyme/GradientUtils.h
```diff
@@ -79,6 +79,13 @@ extern llvm::StringMap<std::function<llvm::Value *(
     shadowHandlers;
 
 class DiffeGradientUtils;
+
+/// This is the main entry point to register a custom derivative for language frontends.
```
wsmoses, line 83: I wouldn't say the main way, but perhaps "how to register custom derivatives programmatically"?

enzyme/Enzyme/GradientUtils.h
```diff
@@ -79,6 +79,13 @@ extern llvm::StringMap<std::function<llvm::Value *(
...
+/// It is more general, and therefore recommended over trying to register
```
wsmoses, line 84: I wouldn't go so far as to recommending it quite yet, it is a fair bit more complex after all

enzyme/Enzyme/GradientUtils.h
```diff
@@ -79,6 +79,13 @@ extern llvm::StringMap<std::function<llvm::Value *(
     shadowHandlers;
 
 class DiffeGradientUtils;
+
+/// This is the main entry point to register a custom derivative for language frontends.
+/// It is more general, and therefore recommended over trying to register 
+/// custom-derivatives in the llvm-ir module. Examples on why it is more general include custom-derivatives
+/// for non-default activity cases (e.g. a function call where a pointer or a float scalar is marked as const).
```
wsmoses, line 86: oh can you actually write out an example here where that function call is handled using the gutils api? I think that would be helpful for people

enzyme/Enzyme/GradientUtils.h
```diff
@@ -79,6 +79,13 @@ extern llvm::StringMap<std::function<llvm::Value *(
...
+/// It also allows using Enzyme and LLVM analysis, e.g. activity analysis, differential use analysis, alias analysis.
+/// To get a better low-level understanding, the documentation in CApi.cpp can be read.
 extern llvm::StringMap<std::pair<
```
wsmoses, line 89: Can you give a go explaining that there are two handlers here, an augmented forward pass handler and a reverse pass handler, explaining what they mean, and what the arguments to each are (if you don't know leave an empty space and we can fill it in on GH)

enzyme/Enzyme/GradientUtils.h
```diff
@@ -79,6 +79,13 @@ extern llvm::StringMap<std::function<llvm::Value *(
     shadowHandlers;
 
 class DiffeGradientUtils;
+
+/// This is the main entry point to register a custom derivative for language frontends.
+/// It is more general, and therefore recommended over trying to register 
+/// custom-derivatives in the llvm-ir module. Examples on why it is more general include custom-derivatives
+/// for non-default activity cases (e.g. a function call where a pointer or a float scalar is marked as const).
+/// It also allows using Enzyme and LLVM analysis, e.g. activity analysis, differential use analysis, alias analysis.
+/// To get a better low-level understanding, the documentation in CApi.cpp can be read.
```
wsmoses, line 88: no, the docs should be here not CApi.cpp. CApi.cpp should point users here

wsmoses:
> Also, let us not loose the original focus due to which I started looking into documenting this. Under release mode, high-level vec resize functions disappear, as visible here: https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=a68343446d1e5d79183603a1eb3b089f You'll see a call to `alloc::raw_vec::RawVecInner<A>::reserve::do_reserve_and_handle` which takes the original pointer and the original len of the primal vec. I _think_ gutils can give us the shadow pointer, but not the shadow len, correct? In that case, it seems hardly possible to define a custom-derivative for it (since it would be a lot of effort to find the shadow len). So before continuing here, I would need to mark one of the more high-level functions as inline-never, until we're done with AD?

The shadow and primal should be structurally equivalent so the primal len == shadow len

wsmoses:
> > Also, let us not loose the original focus due to which I started looking into documenting this. Under release mode, high-level vec resize functions disappear, as visible here: https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=a68343446d1e5d79183603a1eb3b089f You'll see a call to `alloc::raw_vec::RawVecInner<A>::reserve::do_reserve_and_handle` which takes the original pointer and the original len of the primal vec. I _think_ gutils can give us the shadow pointer, but not the shadow len, correct? In that case, it seems hardly possible to define a custom-derivative for it (since it would be a lot of effort to find the shadow len). So before continuing here, I would need to mark one of the more high-level functions as inline-never, until we're done with AD?
> 
> The shadow and primal should be structurally equivalent so the primal len == shadow len

If you want you can relax this requirement and manually

ZuseZ4:
> > 
> 
> The shadow and primal should be structurally equivalent so the primal len == shadow len

Hmm, I used to allow the shadow to be larger than the primal, but I guess this is a valid motivation to enforce equality, at least for now. So you think this is still high-level enough? Asking since you told me that realloc is too low and that I should write higher-level rules.

enzyme/Enzyme/GradientUtils.h
ZuseZ4: For simplicity I've done the example for fwd-mode only, but I can check for the reverse mode.
I've also added what Enzyme generates on it's own as a reference, since that's what we want to generate, but
it makes the example larger. I think it can help people who're not too familiar with llvm's builder to visualize, but I can also delete it.

enzyme/Enzyme/CApi.cpp
```diff
@@ -340,6 +340,9 @@ void EnzymeRegisterAllocationHandler(char *Name, CustomShadowAlloc AHandle,
     };
 }
 
+
+/// This is the entry point to register reverse-mode custom derivatives programmatically.
+/// A more detailed documentation is available in GradientUtils.h
```
wsmoses, line 345: can you refer to the doxygen variable of note in gradientutils.h

enzyme/Enzyme/GradientUtils.h
```diff
@@ -79,13 +79,81 @@ extern llvm::StringMap<std::function<llvm::Value *(
...
+///
+/// The custom rule for this function could be:
+/// customCallHandlers["my_pow"] = [](llvm::IRBuilder<> &Builder, llvm::CallInst *CI, GradientUtils &gutils, llvm::Value *&dcall, llvm::Value *&normalReturn, llvm::Value *&shadowReturn) {
+///  auto x = CI->getArgOperand(0);
+///  auto y = CI->getArgOperand(1);
+///  auto xprime = gutils.getNewFromOriginal(x);
+///  auto yprime = gutils.getNewFromOriginal(y);
+///  bool is_x_active = !gutils.isConstantValue(x);
+///  bool is_y_active = !gutils.isConstantValue(y);
+///  normalreturn = Builder.CreateCall(Intrinsic::pow, {x, y});
```
wsmoses, line 118: this needs the be passed in `Value* args[2] = {x, y}` or something iirc?

enzyme/Enzyme/GradientUtils.h
```diff
@@ -79,13 +79,81 @@ extern llvm::StringMap<std::function<llvm::Value *(
...
+///     // if y were inactive, this would be conceptually equivalent to generating
+///     // define internal double @fwddiffetester(double %x, double %"x'", double %y) #1 {
+///     //   %0 = fsub fast double %y, 1.000000e+00
+///     //   %1 = call fast double @llvm.pow.f64(double %x, double %0)
+///     //   %2 = fmul fast double %y, %1
+///     //   %3 = fmul fast double %"x'", %2
+///     //   ret double %3
+///     // }
+///  }
+///  if (is_y_active) {
```
wsmoses, line 134: just for fun can you test out this code and confirm it works?

ZuseZ4: I'm sorry, but that's too much of an ask. Would you be ok to lower your standards and just merge it as is,
I'm happy to add a warning that this is un-tested pseudo code.
If you think of all the attempts like the new website which got rolled back and died, or all the previous attempts to document Enzyme I think it would be much more valuable to have *some* docs for this over having none at all.
I have already been sitting down with another person using enzyme and we had to reverse-engineer this together, so it's not like I'm the only user of this.

wsmoses: just checking that the thing described in the docs actually runs feels like the minimum bar we should have here since otherwise someone could read the doc and get confused when it doesn't work.

this PR has been open for a few months and isn't in any rush, happy to wait until you have time to test it

PR #2200, opened 2024-12-18
mlir: Add Enzyme ops removal on structured control flow

Pangoraw:
TODO:
 - [x] scf: support non-constant iterations `Cache<f32> -> tensor<?xf32>`.
 - [x] scf: push/pop only once if a value is pushed multiple times.
 - [x] Cache of tensor (nested for).
 - [x] passes: add option in enzyme pass to try to remove enzyme ops after generating the function. This should help with higher order diff. ref #2214.
 - [ ] scf: graph min-cut.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
@@ -28,6 +28,13 @@ namespace {
 
 static mlir::Type batchType(mlir::Type type, int64_t width) {
   if (width > 1 || ShapedType::isDynamic(width)) {
+    if (auto TT = dyn_cast<mlir::TensorType>(type)) {
+      SmallVector<int64_t> shape;
+      shape.reserve(TT.getShape().size());
+      shape.push_back(width);
+      shape.append(TT.getShape().begin(), TT.getShape().end());
+      return TT.clone(shape);
+    }
     return RankedTensorType::get({width}, type);
   }
   return type;
--- a/enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
@@ -27,17 +27,18 @@ using namespace mlir::enzyme;
 namespace {
 
 static mlir::Type batchType(mlir::Type type, int64_t width) {
-  if (width > 1 || ShapedType::isDynamic(width)) {
-    if (auto TT = dyn_cast<mlir::TensorType>(type)) {
-      SmallVector<int64_t> shape;
-      shape.reserve(TT.getShape().size() + 1);
-      shape.push_back(width);
-      shape.append(TT.getShape().begin(), TT.getShape().end());
-      return TT.clone(shape);
-    }
-    return RankedTensorType::get({width}, type);
+  if (width == 1)
+    return type;
+
+  if (auto TT = dyn_cast<mlir::TensorType>(type)) {
+    SmallVector<int64_t> shape;
+    shape.reserve(TT.getShape().size() + 1);
+    shape.push_back(width);
+    shape.append(TT.getShape().begin(), TT.getShape().end());
+    return TT.clone(shape);
   }
-  return type;
+
+  return RankedTensorType::get({width}, type);
 }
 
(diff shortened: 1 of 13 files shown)


enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -0,0 +1,40 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "RemovalUtils.h"
+#include "Interfaces/AutoDiffOpInterface.h"
+
+mlir::Type mlir::enzyme::CacheInfo::batchType() {
+  return mlir::enzyme::CacheInfo::batchType(mlir::ShapedType::kDynamic);
+}
+
+mlir::Type mlir::enzyme::CacheInfo::batchType(int64_t dim) {
```
wsmoses, line 16: so there is already an Enzyme Autodiff Type interface, which should have a method for batching (and if not that would probably be the right place for this)

wsmoses (approved):

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -32,126 +35,499 @@ using namespace mlir::enzyme;
...
-        // There is an assumption here that there is only regions that branch to
-        // the successor. Specifically, otherwise we would need to
-        // gutils->addToDiffe select (if came from that result)
-        for (auto &&[prev, post] : llvm::zip(operandRange, resDiffes)) {
-          if (!post)
-            continue;
-          if (!gutils->isConstantValue(prev))
-            gutils->addToDiffe(prev, post, builder);
+    auto numIters = getConstantNumberOfIterations(forOp);
+    Value inductionVariable; // [0, N[ counter
```
wsmoses, line 155: nit: presumably `N]`

enzyme/Enzyme/MLIR/Interfaces/CloneFunction.cpp
```diff
@@ -232,6 +240,11 @@ FunctionOpInterface CloneFunctionWithReturns(
 
   {
     auto &blk = NewF.getFunctionBody().front();
+    if (width != 1) {
```
wsmoses, line 243: similarly this seems wrong?

enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
```diff
@@ -182,9 +184,8 @@ class ComplexTypeInterface
     return builder.create<complex::ConjOp>(loc, a)->getResult(0);
   }
 
-  Type getShadowType(Type self, unsigned width) const {
-    assert(width == 1 && "unsupported width != 1");
-    return self;
+  Type getShadowType(Type self, int64_t width) const {
+    return batchType(self, width);
```
wsmoses, line 188: in a separate different PR, it may be worthwhile switching getShadowType and the likes to take an ArrayRef<int64_t> indices to batch on (@jumerckx did something similar when adding batched differentiation broadcast earlier)

wsmoses (approved):
looks good, though there's some unrelated batch stuff here that probably shouldn't be here (maybe leftover from debugging)

Pangoraw:
> though there's some unrelated batch stuff here that probably shouldn't be here (maybe leftover from debugging)

Yeah I had cherry-picked changes from #2181 before it was merged, the diff should be more correct now that I rebased.

enzyme/Enzyme/MLIR/Interfaces/CloneFunction.cpp
```diff
@@ -27,7 +27,11 @@ getFunctionTypeForClone(mlir::FunctionType FTy, DerivativeMode mode,
   for (auto &&[Ty, returnPrimal, returnShadow, activity] : llvm::zip(
            FTy.getResults(), returnPrimals, returnShadows, ReturnActivity)) {
     if (returnPrimal) {
-      RetTypes.push_back(Ty);
+      if (width != 1) {
```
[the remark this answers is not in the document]
Pangoraw: this isn't part of this PR anymore, (rebased onto main).

enzyme/Enzyme/MLIR/Interfaces/CloneFunction.cpp
Pangoraw: this isn't part of this PR anymore, (rebased onto main).

wsmoses (approved):

PR #2206, opened 2024-12-24
tablegen: Add StaticSelect to select based on static condition

Pangoraw:
cc @mofeing, we can collaborate to integrate in your fft pr.

I am open to other naming ideas.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/Common.td
+++ b/enzyme/Enzyme/MLIR/Implementations/Common.td
@@ -96,7 +96,7 @@ class GlobalExpr<bit uses_primal, bit uses_shadow, string val> : Operation<uses_
 // The three arguments version accepts a name as a first argument
 // which is then available in the condition as a `Value` under the
 // variable `imVal`.
-class StaticSelect<string condition_> : Operation</*usesPrimal*/0, /*usesShadow*/0, /*usesCustom*/1> {
+class StaticSelect<string condition_> : Operation</*usesPrimal*/0, /*usesShadow*/0, /*usesCustom*/0> {
   string condition = condition_;
 }
 
--- a/enzyme/Enzyme/MLIR/Implementations/Common.td
+++ b/enzyme/Enzyme/MLIR/Implementations/Common.td
@@ -90,7 +90,13 @@ class GlobalExpr<bit uses_primal, bit uses_shadow, string val> : Operation<uses_
   string value = val;
 }
 
-class StaticIf<bit uses_primal, bit uses_shadow, string condition_> : Operation<uses_primal, uses_shadow> {
+// Class for a dag operator that generates either a or b
+// It can then be used with a two or three arguments.
+// The two arguments version is (StaticSelect a, b)
+// The three arguments version accepts a name as a first argument
+// which is then available in the condition as a `Value` under the
+// variable `imVal`.
+class StaticSelect<string condition_> : Operation</*usesPrimal*/0, /*usesShadow*/0, /*usesCustom*/1> {
   string condition = condition_;
 }
 
@@ -103,13 +109,14 @@ class Inst<string mnemonic, string dialect_, string postop_="">  : Operation</*p
 def Op {
 }
 
-def SelectIfActive : Operation</*primal*/0, /*shadow*/0, /*custom*/1> {
+def SelectIfActive : StaticSelect<"!gutils->isConstantValue(imVal)">;
 
-}
-
-def SelectIfComplex : Operation</*primal*/1, /*shadow*/0, /*custom*/0> {
-
-}
+def SelectIfComplex : StaticSelect<[{
+  auto ty = imVal.getType();
+  ty.isa<ComplexType>() ||
+    ty.isa<TensorType>() &&
+    ty.cast<TensorType>().getElementType().isa<ComplexType>();
+}]>;
 
 class ConstantFP<string val, string dialect_, string op_, string type_=""> : Operation</*primal*/0, /*shadow*/0> {
   string value = val;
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/MLIR/Implementations/Common.td
```diff
@@ -86,10 +86,14 @@ def DiffeRet : DiffeRetIndex<[-1]>;
 def Shadow : Operation</*primal*/0, /*shadow*/1> {
 }
 
-class GlobalExpr<bit uses_primal, bit uses_shadow, string val> : Operation<uses_primal, uses_shadow>{
+class GlobalExpr<bit uses_primal, bit uses_shadow, string val> : Operation<uses_primal, uses_shadow> {
   string value = val;
 }
 
+class StaticIf<bit uses_primal, bit uses_shadow, string condition_> : Operation<uses_primal, uses_shadow> {
```
wsmoses, line 93: I think probably we should remove the explicit needs primal and shadow, and instead directly integrate this into the use analysis (lookup SelectIfActive in this case we wouldn’t do the check on if active but check on whatever the string is)

Pangoraw: Looking into it, I could not find the custom use analysis for SelectIfActive, it seems that instead SelectIfActive has no analysis:

```cpp
       SmallVector<bool> cachedArguments(Operation *op,
                                 MGradientUtilsReverse *gutils) const {
         SmallVector<bool> toret(op->getNumOperands(), false);
         for (size_t idx=0; idx<op->getNumOperands(); idx++) {
            bool used = false;
            // Rule (Op ?:$x, ?:$y)
            // Arg 0 : (SelectIfActive ?:$x, (CMul (DiffeRet), ?:$y), ?:$x)
            // Arg 1 : (DiffeRet)
            toret[idx] = used;
         }
         return toret;
       }
```
And it currently works because either it is used in the Forward diff which seem to  rely on the analysis or the primals are used outside of the SelectIfActive node.

It seems that the assert here does not trigger due to the way enzyme-tblgen might be built?

https://github.com/EnzymeAD/Enzyme/blob/1ca3ceb12723b7246fe19016bca6e91fba37f1c7/enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp#L1208

So really, it should base its use on its arguments.

wsmoses: Ah yeah we should add support there (actually for both static if and if active). The reason we never hit that branch is because use analysis is auto generated from reverse mode rules and I think we only use select if active in forward.

However here we really want to add the correct diff use analysis, since then we can for example use static if to define rules for reduce and static if it’s a known max min or add etc

wsmoses: But also it should do so recursively for the operands of the select if active (depending on if active)

Pangoraw: Right, the same analysis need to be used for SelectIfComplex as well (or it should be replaced with an instance of StaticIf)

wsmoses: I mean ideally we can define selectifcomplex in tablegen with selectif

Pangoraw:
I am trying to implement SelectIfActive based on StaticSelect as well but it looks like it has special things for vectors?

wsmoses:
ah yeah that one might be hard. Yeah the return of handle is whether or not the value is batched already (e.g. for batched vector mode). The right thing to do is probably as follows.

Emit both the lhs and the rhs. If they have the same vector mode set (either both not vectored or both vectored), that's fine and return it.

If one has it set, upgrade the other one to vector mode (e.g. do that insertvalue related stuff), and return vector = true.

wsmoses:
cc @jumerckx this will probably also need an MLIR version too at one point (which presumably calls the new broadcast op)

enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
```diff
@@ -1203,9 +1166,27 @@ void handleUse(
   bool usesShadow = Def->getValueAsBit("usesShadow");
   bool usesCustom = Def->getValueAsBit("usesCustom");
 
-  // We don't handle any custom primal/shadow
-  (void)usesCustom;
-  assert(!usesCustom);
+  // This only concerns instances of StaticSelect for now
+  if (usesCustom) {
```
wsmoses, line 1170: very minor but can we explicitly check opName == "StaticSelect" || Def->isSubClassOf("StaticSelect") and leave this assertion as is?

enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
```diff
@@ -1203,10 +1166,30 @@ void handleUse(
   bool usesShadow = Def->getValueAsBit("usesShadow");
   bool usesCustom = Def->getValueAsBit("usesCustom");
 
-  // We don't handle any custom primal/shadow
   (void)usesCustom;
   assert(!usesCustom);
 
+  if (Def->isSubClassOf("StaticSelect")) {
```
wsmoses, line 1172: these should be in the other order the assertion come after the handler, and the handler should also return to ensure the error isn't hit

Pangoraw: ah right, I moved the assert after and set StaticSelect to have useCustom

wsmoses: I realized a bug in the useCustom analysis logic that was easier to just code up so I went ahead and fixed it

PR #2222, opened 2025-01-16
Cache batched functions and recursively batch.

jumerckx:
Fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/239

I should still check what happens with recursive functions though.


wsmoses (approved):
lgtm but add a test?

jumerckx:
Should be good now

wsmoses (approved):

PR #2223, opened 2025-01-16
add user data to EnzymeRegisterCallHandler

martinjrobins:
This adds two items to the C API:
1. a `void *data` arg to `EnzymeRegisterCallHandler` so users can supply their own data to the callbacks
2. `EnzymeRegisterCallHandler` and `EnzymeGradientUtilsNewFromOriginal` added to the C API header

My particular motivation for 1. is that I need access to the LLVM module pointer in my reverse callback, so the data arg allows me to pass this in. I know the LLVM C++ API has `getParent` that can be used to get the module from the original call instruction, but I'm using the LLVM C API from rust, which as far as I can see does not have this functionality. In any case, I figure it would be generally useful for users to pass arbitrary data to these callbacks.

### DIFF
--- a/enzyme/Enzyme/CApi.cpp
+++ b/enzyme/Enzyme/CApi.cpp
@@ -350,8 +350,8 @@ void EnzymeRegisterCallHandler(const char *Name,
     LLVMValueRef normalR = wrap(normalReturn);
     LLVMValueRef shadowR = wrap(shadowReturn);
     LLVMValueRef tapeR = wrap(tape);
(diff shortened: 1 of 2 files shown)


wsmoses:
https://llvm.org/doxygen/IR_2Core_8cpp_source.html#l01940 this should do that

martinjrobins:
yea, you are right, thanks! Ok I guess I just need 2. then. I'll open another PR with just that

PR #2229, opened 2025-01-22
Drive enzyme ops removal using pattern rewriter

Pangoraw:
This modifies the interface to notify of IR mutations via the pattern rewriter to prevent error during iteration.
Also fixes bugs in the scf.for implementation of the interface which resulted in non-deterministic test failures.

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/RemoveUnusedEnzymeOps.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/RemoveUnusedEnzymeOps.cpp
@@ -462,12 +462,17 @@ struct RemoveUnusedEnzymeOpsPass
 
     applyPatterns(op);
 
-    // TODO: we should use the result here.
+    bool failed = false;
     op->walk([&](FunctionOpInterface func) {
       PostOrderWalkDriver driver(func);
       driver.initializeWorklist();
-      (void)driver.processWorklist();
+      failed |= driver.processWorklist().failed();
     });
+
+    if (failed) {
+      signalPassFailure();
+      return;
+    }
 
     applyPatterns(op);
   }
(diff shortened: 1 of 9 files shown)


enzyme/Enzyme/MLIR/Passes/RemoveUnusedEnzymeOps.cpp
```diff
@@ -290,13 +291,25 @@ struct InitSimplify : public OpRewritePattern<enzyme::InitOp> {
...
 };
 
+struct EnzymeOpsRemoverOpPattern
+    : public OpInterfaceRewritePattern<enzyme::EnzymeOpsRemoverOpInterface> {
+  using OpInterfaceRewritePattern<
+      enzyme::EnzymeOpsRemoverOpInterface>::OpInterfaceRewritePattern;
+
+  LogicalResult matchAndRewrite(enzyme::EnzymeOpsRemoverOpInterface iface,
+                                PatternRewriter &rewriter) const final {
+    return iface.removeEnzymeOps(rewriter);
```
wsmoses, line 301: High level comment then we can merge: It's critical that we make sure that whatever region op we simplify doesn't have its interface called if there are interior ops for push/pop/set/get. Where is that check happening?

For example we could do something here that says 
```if (iface.walk(op)[]{ return isa enzyme push/pop op; })) return failure();
```

before calling iface.removeEnzymeOps

Pangoraw: So currently, implementers of the interface should make sure of that but that is not ideal.

I think in this case we want to process ops in post-order (like the walk did before) and newly created ops should be processed only if they are after in the post-order.

I will try to implement something like this.

wsmoses:
@Pangoraw this good to go?

Pangoraw:
Yes, this is ok to merge from my POV. I updated the corresponding Enzyme-JAX PR as well.

PR #2231, opened 2025-01-30
[bazel] Fully replace gentbl with gentbl_cc_library

rupprecht:
Enzyme uses tablegen generators from both llvm/tblgen.bzl and mlir/tblgen.bzl.
This consolidates usage so that we only use one of them (from MLIR).
Later we may delete the llvm/tblgen.bzl entirely.

This is mostly mechanical, but as a side effect, we need to adjust `includes`
to pick up `bundled_includes.h`. It was probably relying on an `includes`
config from the llvm/tblgen.bzl implementation that is no longer being provided.

See also https://github.com/llvm/llvm-project/pull/124996.


rupprecht:
CI appears to be failing due to https://github.com/llvm/llvm-project/commit/29441e4f5fa5f5c7709f7cf180815ba97f611297 now. Likely needs an update in Enzyme code, although I'm not familiar w/ that change, so I don't have a fix to suggest. I was able to test this locally by pinning LLVM to an older revision.

If someone can land a change to fix Enzyme w/ LLVM from trunk, I can push another empty commit to rerun CI.

wsmoses:
nocapture fix landed

PR #2245, opened 2025-02-15
recognizing ENZYME_RUNPASS for lib/LLVM Enzyme builds

ZuseZ4:
I noticed that the ENZYME_RUNPASS variable is ignored, when you just pass it as `-DENZYME_RUNPASS` to cmake.
This might also affect your `BUILD` file, idk.
With these changes it now does run the augmentedPassBuilder.

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -3658,8 +3658,6 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
   PB.registerFullLinkTimeOptimizationEarlyEPCallback(loadLTO);
 }
 
-// A version used by Rust-Enzyme, which does not require cmake adjustments.
-// Rust just always sets augment to true.
 extern "C" void registerEnzyme2(llvm::PassBuilder &PB, bool augment = false) {
   if (augment) {
     augmentPassBuilder(PB);
@@ -3696,8 +3694,6 @@ extern "C" void registerEnzyme2(llvm::PassBuilder &PB, bool augment = false) {
       });
 }
 
-// A version used by ClangEnzyme and LLDEnzyme, allowing to
-// hardcode the usage of augmentPassBuilder at build time.
 extern "C" void registerEnzyme(llvm::PassBuilder &PB) {
 #ifdef ENZYME_RUNPASS
   registerEnzyme2(PB, /*augment*/ true);
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -3658,7 +3658,8 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
   PB.registerFullLinkTimeOptimizationEarlyEPCallback(loadLTO);
 }
 
-extern "C" void registerEnzymeAndPassPipeline(llvm::PassBuilder &PB, bool augment = false) {
+extern "C" void registerEnzymeAndPassPipeline(llvm::PassBuilder &PB,
+                                              bool augment = false) {
   if (augment) {
     augmentPassBuilder(PB);
   }
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -3658,7 +3658,7 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
   PB.registerFullLinkTimeOptimizationEarlyEPCallback(loadLTO);
 }
 
-extern "C" void registerEnzyme2(llvm::PassBuilder &PB, bool augment = false) {
+extern "C" void registerEnzymeAndPassPipeline(llvm::PassBuilder &PB, bool augment = false) {
   if (augment) {
     augmentPassBuilder(PB);
   }
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -3658,10 +3658,12 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
   PB.registerFullLinkTimeOptimizationEarlyEPCallback(loadLTO);
 }
 
-extern "C" void registerEnzyme(llvm::PassBuilder &PB) {
-#ifdef ENZYME_RUNPASS
-  augmentPassBuilder(PB);
-#endif
+// A version used by Rust-Enzyme, which does not require cmake adjustments.
+// Rust just always sets augment to true.
+extern "C" void registerEnzyme2(llvm::PassBuilder &PB, bool augment = false) {
+  if (augment) {
+    augmentPassBuilder(PB);
+  }
   PB.registerPipelineParsingCallback(
       [](llvm::StringRef Name, llvm::ModulePassManager &MPM,
          llvm::ArrayRef<llvm::PassBuilder::PipelineElement>) {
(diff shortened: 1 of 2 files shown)


wsmoses (changes requested):
This should be disabled for llvmenzyme thougj? The flag says that the enzyme pass should be injected to be run by default in the pass pipeline

ZuseZ4:
What is the difference between libEnzyme-X.so and LLVMEnzyme-19.so? I thought they would have just different linking behaviour, but the same functionality? We use libEnzyme because it fixed some linking issues I can't remember, so I don't mind only updating it and removing the LLVMEnzyme one (?)

wsmoses:
LLVMEnzyme is an llvm pass plugin, libenzyme is a dynamic library you can link into programs that aren't llvm's opt

ZuseZ4:
Fair, I vaguely had in mind that at some point I also linked against LLVMEnzyme when using the Enzyme API, but I might have confused something.

wsmoses (changes requested):
I'm still not understanding why this is needed here

ZuseZ4:
I want Enzymes opt pass to run befire autodiff, as it is supposed to improve performance. Right now we already set `-DENZYME_RUNPASS` when calling cmake, but the value is ignored and does not cause Enzyme to run the optimization pass. It only affects libEnzyme, as lld and clangenzyme have it hardcoded. I could also hardcode it for libEnzyme, but this solution allows other users to run the autodiff pass without the opt pass.

ZuseZ4:
https://github.com/EnzymeAD/Enzyme/blob/3ea5fe2619a18569b88fff278e2b5b8b30073f64/enzyme/Enzyme/Enzyme.cpp#L3662

wsmoses:
Why not just run augmentPassBuilder?

ZuseZ4:
Why have RustEnxyme differ from ClangEnzyme or LLDEnzyem?

wsmoses:
The use you linked is specific to llvm pass loading which doesn't make sense for API users, since its better form to build the pass manager directly (like with the function I mentioned)

ZuseZ4:
This has all been discussed in the zulip thread where I cc'ed you and pinged you a few times, so if you have extra knowledge it would be great if you could answer there. There I had asked which functions I should use in which order, and this is what a few other devs and I came up with.

From what we saw, we will always run `augmentPassBuilder` and then `registerEnzyme` to then run the `enzyme` pass.
We will never just run `augmentPassBuilder` or the enzyme pass without the other. So if there is a performance benefit to run augmentPassBuilder separately, it would be good if you could give an explanation on how it's supposed to be used and what is the difference over just enabling it here.

ZuseZ4:
Ok, wild guess now, but in an attempt to understand the difference, could it be that you missed all the messages/pings where I/Rust people asked about the Enzyme pass, e.g. [here](https://rust-lang.zulipchat.com/#narrow/channel/187780-t-compiler.2Fwg-llvm/topic/Enzyme.20Pass.20integration.2E).
Rust moved over to a pass based Enzyme approach various months ago, hence all my question about the enzyme pass.
If you missed all of those this discussion makes more sense to me and hopefully now to you.

wsmoses:
I'm still a bit confused how you're intending to use this. In any case, perhaps you can make a second function void augmentAndRegisterEnzyme(PB, bool shouldAugment) which contains the current definition, and then is true if the current Enzyme_runpass is set. That way you can just call the api function without setting a compile time var

ZuseZ4:
The intention is to have Rust differ as little as possible from Clang/LLD-Enzyme, since all three are pass based, as opposed to Julia which is Api based., Clang/LLD Enzyme both first run augmentPassBuilder and then register Enzyme and run it. So my confusion comes from your recommendation to handle Rust and C++ different here. 
Doing it at runtime would be totally fine for me, but then I wonder why you don't do the same with the clang/lld plugin? What are you confused about?

ZuseZ4:
Also for the long term goal, the idea is to have autodiff run as the pass (e.g. because of your recommendation and because it handles higher order derivatives in the right order automatically). For things like custom-derivatives you recommended to not copy c++Enzyme, but rather link against libEnzyme, so we can use the more advanced registerCustomDerivative functions, since they are more granular than the metadata approach.

So my understanding is that clang will in the future move to the api for custom derivatices, Enzyme.jl will move to the pass for autodiff (but keep using the api for things like custom derivatives) and then all 3 frontends will behave identical. Since Rust is coming in later, I directly try to do the right approach for each feature. Does that help?

ZuseZ4:
@wsmoses adding the runtime variant and updated the old one to forward, to make sure they can never get out of sync.

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -3658,10 +3658,12 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
   PB.registerFullLinkTimeOptimizationEarlyEPCallback(loadLTO);
 }
 
-extern "C" void registerEnzyme(llvm::PassBuilder &PB) {
-#ifdef ENZYME_RUNPASS
-  augmentPassBuilder(PB);
-#endif
+// A version used by Rust-Enzyme, which does not require cmake adjustments.
+// Rust just always sets augment to true.
+extern "C" void registerEnzyme2(llvm::PassBuilder &PB, bool augment = false) {
```
wsmoses, line 3663: can we come up with a better name, like registerEnzymeAndPassPipeline

and we shouldn't need two versions of this if we have the optional arg?

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -3658,10 +3658,12 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
   PB.registerFullLinkTimeOptimizationEarlyEPCallback(loadLTO);
 }
 
-extern "C" void registerEnzyme(llvm::PassBuilder &PB) {
-#ifdef ENZYME_RUNPASS
-  augmentPassBuilder(PB);
-#endif
+// A version used by Rust-Enzyme, which does not require cmake adjustments.
```
wsmoses, line 3661: this comment isn't needed

enzyme/Enzyme/Enzyme.cpp
ZuseZ4: Yep, that was my question:
> So my confusion comes from your recommendation to handle Rust and C++ different here.

and 

> Doing it at runtime would be totally fine for me, but then I wonder why you don't do the same with the clang/lld plugin? 

Shall I then remove ENZYME_RUNPASS completely and update the clang/lld usages to set it to true?

ZuseZ4: We can not I guess, seems like the registration process has a fixed interface?
Feel free to push your own changes and merge it.

```
/home/manuel/prog/Enzyme/enzyme/Enzyme/Enzyme.cpp:3704:59: error: cannot initialize a member subobject of type 'void (*)(PassBuilder &)' with an lvalue of type 'void (llvm::PassBuilder &, bool)': different number of parameters (1 vs 2)
 3704 |   return {LLVM_PLUGIN_API_VERSION, "EnzymeNewPM", "v0.1", registerEnzymeAndPassPipeline};
      |                                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

wsmoses: okay in that case just give the functions more reasonable names then registerEnzyme and registerEnzymeAndPassPipeline (being the false / true version), one calling the other

ZuseZ4: done

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -3694,6 +3694,14 @@ extern "C" void registerEnzyme(llvm::PassBuilder &PB) {
       });
 }
 
+extern "C" void registerEnzyme(llvm::PassBuilder &PB) {
+#ifdef ENZYME_RUNPASS
+  registerEnzyme2(PB, /*augment*/ true);
```
wsmoses, line 3699: this clearly won't compile

wsmoses (approved):

PR #2246, opened 2025-02-18
fix-crash-by-coalese

minansys:
Fix the typo


wsmoses (commented):
do you have a test for this by chance (this certainly seems like we should fix this though)

Issue #2251, opened 2025-02-22
Incorrect range propagation since LLVM20


ZuseZ4:
`Attribute 'range(i32 0, 2)' applied to incompatible type!`, resulting in `ERROR: function failed verification (2)`

It seems like LLVM20 is more likely to use the range metadata, which I encountered in most non-trivial programs.

The problem seems to be that range attributes are kept by Enzyme, even if the augmented function has a more complex return type.
e.g. the primal returns `i32`, while the `fakeaugmented` one returns `{ ptr, i32 }`. For the later, range isn't valid.
 
The compiler Explorer is broken, so I'm adding a testfile here. 
Based on LLVM commit https://github.com/rust-lang/llvm-project/tree/92e80685d0d5dcea3ccf321995c43b72338639c6

Instructions to reproduce:
`myopt file.ll --load-pass-plugin=<path>/build/Enzyme/LLVMEnzyme-20.so -passes='enzyme' -enzyme-loose-types -S`

[file.ll.txt](https://github.com/user-attachments/files/18925797/file.ll.txt)

ZuseZ4:
llvm-reduce gives

```llvm
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

define void @foo(ptr %0, ptr %1, ptr %2, ptr %3, ptr %4, ptr %5) personality ptr null {
"_ZN5alloc7raw_vec20RawVecInner$LT$A$GT$15try_allocate_in17hd1f0620d66c47f1bE.exit":
  %6 = call i32 @bar(ptr %0)
  ret void
}

define void @caller() {
  tail call void (...) @__enzyme_autodiff(ptr @foo, metadata !"enzyme_dup", ptr null, ptr null, metadata !"enzyme_const", ptr null, metadata !"enzyme_const", ptr null, metadata !"enzyme_const", ptr null, metadata !"enzyme_const", ptr null, metadata !"enzyme_dup", ptr null, ptr null)
  ret void
}

declare void @__enzyme_autodiff(...)

define range(i32 0, 2) i32 @bar_loop(ptr %0) personality ptr null {
  ret i32 0
}

define i32 @bar(ptr %0) personality ptr null {
  %2 = call i32 @bar_loop(ptr %0)
  ret i32 0
}

; uselistorder directives
[...]

PR #2253, opened 2025-02-26
Drop range

ZuseZ4:
fixes https://github.com/EnzymeAD/Enzyme/issues/2251
The test fails without and passes with this fix.


wsmoses (approved):

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -1721,6 +1721,9 @@ void clearFunctionAttributes(Function *f) {
     f->removeRetAttr(Attribute::Alignment);
   }
   Attribute::AttrKind attrs[] = {
+#if LLVM_VERSION_MAJOR >= 20
+    Attribute::Range,
```
wsmoses, line 1725: can you check which version of llvm introduced the range attr (and confirm it makes your number here and in the test)

wsmoses: [my lazy trick for this is removing the range, and seeing what the first version which compiles]

ZuseZ4: updated, 18 failed, 19 worked.

ZuseZ4:
@wsmoses Seems like automerge still doesn't work due to the remaining failing tests, so you still have to manually merge it.

ZuseZ4:
@wsmoses I think you still need to merge this manually.

wsmoses:
@tgymnich it looks like the merge is getting a 404 when doing the ci-bench query, if that rings any bells

tgymnich:
@wsmoses thats the continuous benchmarking hosted at https://bench.enzyme.csail.mit.edu/superset. I restarted the container to hopefully fix the 404. If this is no longer actively used we can also shut this down altogether and remove the build step.

PR #2265, opened 2025-03-14
EnzymeMLIR ProbProg Sample Op

sbrantq:
Assuming we only sample once by directly calling distribution function for now, so sampler data structure is omitted


sbrantq:
Refactoring things for a while...

sbrantq:
Replaced by PR #2325

PR #2271, opened 2025-03-29
Fwddiff ret_activity canonicalizations

vimarsh6739:
Simplifies return activity whenever legal. 

- [x] dup --> dupnoneed
- [x] dup --> const
- [x] dupnoneed --> constnoneed
- [x] const --> constnoneed

tagging @jhueckelheim


enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -169,6 +168,272 @@ ForwardDiffOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
+
+    return FwdRetConverter(uop, rewriter, Activity::enzyme_dup, prologue,
+                           epilogue);
+  }
+};
+
+class FwdRetDupNNToConstNN final : public OpRewritePattern<ForwardDiffOp> {
+public:
+  using OpRewritePattern<ForwardDiffOp>::OpRewritePattern;
+
```
wsmoses, line 362: can we simplify this by having one pattern that does all of these, and not needing to create all the separate prologue/epilogues?

enzyme/Enzyme/MLIR/enzymemlir-translate/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ target_link_libraries(enzymemlir-translate
   MLIRSPIRVDialect
   MLIRTranslateLib
   MLIRSupport
+  "ZLIB::ZLIB"
```
wsmoses, line 22: is this needed?

vimarsh6739: cmake build fails without that, so for now I just added it as a quick fix

enzyme/Enzyme/MLIR/Dialect/Ops.cpp
vimarsh6739: We could do that, but i.m.o. the epilogue part becomes hairy (especially if we want to handle `dup` --> `constnoneed` - you are eliminating 2 return values as compared to 1). All transforms here strictly reduce number of args by 1. 

I'll still give it a go.

PR #1883, opened 2024-05-15
error if using Enzyme with gcc/g++

ZuseZ4:
It would be nicer to check if we actually have a LLVM based compiler, but I am worried of some vendor specific compilers having awkward names. So only check for GNU, since that's 100% unsupported. This should still cover almost all cases and prevents issues like this:
closes #1871

Tested locally:
```sh
➜  build git:(recognize-unsupported-gnu) CC=/usr/bin/gcc cmake .. -GNinja  -DEnzyme_DIR=/u/drehwald/prog/Enzyme/enzyme/build/ -DCMAKE_LINKER=lld
project languages: C
C compiler: GNU
CMake Error at /u/drehwald/prog/Enzyme/enzyme/build/EnzymeConfigVersion.cmake:8 (message):
  Enzyme requires an LLVM based C compiler.
Call Stack (most recent call first):
  CMakeLists.txt:5 (find_package)


-- Configuring incomplete, errors occurred!
➜  build git:(recognize-unsupported-gnu) ..
➜  test_find_package git:(recognize-unsupported-gnu) rm -rf build && mkdir build && cd build                                                                
➜  build git:(recognize-unsupported-gnu) CC=clang cmake .. -GNinja  -DEnzyme_DIR=/u/drehwald/prog/Enzyme/enzyme/build/ -DCMAKE_LINKER=lld
-- The C compiler identification is Clang 16.0.5
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /u/drehwald/prog/llvm16/build/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
found dir /u/drehwald/prog/Enzyme/enzyme/build/
found 1
imported targets LLVMEnzyme-16;ClangEnzyme-16;ClangEnzymeFlags;LLDEnzyme-16;LLDEnzymeFlags;LLDEnzymeAssumeUnknownNoFree;LLDEnzymeLooseTypeFlags;LLDEnzymePrintTypeFlags;LLDEnzymePrintFlags;LLDEnzymeNoStrictAliasingFlags
-- Configuring done (0.5s)
-- Generating done (0.0s)
-- Build files have been written to: /u/drehwald/prog/Enzyme/enzyme/test/test_find_package/build
```

PR #2272, opened 2025-04-01
Allow suppression of incompatible compiler warning

dime10:
As discussed [here](https://github.com/EnzymeAD/Enzyme/pull/1883#issuecomment-2762636340), Enzyme can be compiled with non-LLVM compilers, as long as it is loaded/linked into LLVM tools (clang, opt, ...).

This PR adds a CMake option to suppress the compiler warning introduced in #1883.


ZuseZ4:
cc @tgymnich, not sure why Integration tests aren't running here, I approved the workflow.

tgymnich:
@ZuseZ4 I guess this must have been a planned brownout of ubuntu 20.04 runners from GitHub. Here is the fix: https://github.com/EnzymeAD/Enzyme/pull/2262. PR is not merged because we still need to decide wether we want to keep C/C++ CI as required or only run C/C++ CI when a relevant file changes (see paths in PR). Both do not work together since the Required CI pass may never run if you don't touch any of the files define in the workflow.

ZuseZ4 (approved):

ZuseZ4:
@tgymnich I removed the path but CI there was blocked too, so I'm unable to merge it, sorry.
However it seems like the 20.04 runners here came back at some point.
So to not have this PR blocked too long and since no one seems to object I've added it to the merge queue, so they don't have to deal with extra warnings.

PR #2285, opened 2025-04-28
don't access LLVMEnzyme on apple builds if LLVM_PLUGINS isn't set

ZuseZ4:
Upstreaming of https://github.com/rust-lang/enzyme/pull/4
Incorrect guard caused a build failure in https://github.com/rust-lang/rust/pull/140064#issuecomment-2836408078
(There might be another fix needed, but this seems uncontroversial?)

### DIFF
--- a/enzyme/Enzyme/CMakeLists.txt
+++ b/enzyme/Enzyme/CMakeLists.txt
@@ -155,9 +155,9 @@ endif()
 
 if (APPLE)
 # Darwin-specific linker flags for loadable modules.
+if (ENZYME_ENABLE_PLUGINS)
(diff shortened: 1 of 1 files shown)


wsmoses (approved):

ZuseZ4:
C++ Integration test failures are unrelated, e.g.
```
error: <unknown>:0:0: in function preprocess__ZNK5Eigen13DefaultDevice6memcpyEPvPKvm void (ptr, ptr, ptr, i64): Enzyme: Cannot deduce type of copy   call void @llvm.memcpy.p0.p0.i64(ptr align 1 %1, ptr align 1 %2, i64 %3, i1 false) #22
```
Required CI LLVM won't run due to being too outdated, so you'll need to manually merge.

All apple runners which execute this passed.
@wsmoses

PR #2288, opened 2025-04-29
fix apple builds against a no-plugin LLVM

ZuseZ4:
This (in combination with yesterdays fix) let's me build rustc on apple.
Without this fix I get
```
-- BCPass-20 ignored -- Loadable modules not supported on this platform.
CMake Error at BCLoad/CMakeLists.txt:108 (target_include_directories):
  target_include_directories called with non-compilable target type
```

Used rust commands
`./configure --enable-llvm-link-shared --enable-llvm-enzyme --release-channel=nightly --enable-llvm-assertions --enable-clang --enable-lld --enable-option-checking --enable-ninja --disable-docs`
Not used flag: ` --enable-llvm-plugins `

I don't know under which flags exactly we should disable it.
if APPLE AND NO_PLUGINS?

### DIFF
--- a/enzyme/BCLoad/CMakeLists.txt
+++ b/enzyme/BCLoad/CMakeLists.txt
@@ -14,6 +14,7 @@ if (APPLE)
     set(BC_LOAD_FLAGS2 "${BC_LOAD_FLAGS2} --sysroot=${CMAKE_OSX_SYSROOT}")
 endif()
 
+if (ENZYME_ENABLE_PLUGINS)
 if (NOT ("${BC_LOAD_HEADER}" STREQUAL ""))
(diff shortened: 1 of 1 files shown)


wsmoses:
yeah this is definitely wrong, probably go inside of BCLoad itself and only build the plugin if plugins are enabled, like presumably already done within Enzyme

ZuseZ4:
But should this do anything if only libEnzyme is build?

ZuseZ4:
Relevant test run (I'd wait 30 min for it before merging here): https://github.com/rust-lang/rust/pull/140064#issuecomment-2839686585

ZuseZ4:
ok @wsmoses the rust failure is unrelated, so I'd still merge this since it fixes at least a local mac build.

wsmoses (approved):

PR #2292, opened 2025-04-30
Fwd canonicalizer, fix const noneed

wsmoses:


enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -218,6 +218,14 @@ class FwdRetOpt final : public OpRewritePattern<ForwardDiffOp> {
     for (auto [idx, act] : llvm::enumerate(retActivity)) {
       auto iattr = cast<ActivityAttr>(act);
       auto val = iattr.getValue();
+
+      // const_noneed does not have a value associated with it
+      // so we can't index into outputs.
+      if (val == Activity::enzyme_constnoneed) {
```
vimarsh6739, line 224: Instead of an explicit if, we can just delete the `push_back` inside the switch

vimarsh6739: Also need to check it during the new op construction

vimarsh6739: Oh I see, you are doing this before any indexing

PR #2302, opened 2025-05-04
libLLVM.a doesn't exist, so how enzyme is linking LLVM

ZuseZ4:
There are a couple of bugs in both Enzyme and LLVM, so it took me a while to figure this one out.

1) linking against "LLVM" as a constant string only works if there's some type of libLLVM.something.
We only have a libLLVM.so if we build LLVM as a dynamically linked library, which rust does e.g. for x86-64.
There is no libLLVM.a equivalent otherwise, just a lot of individual archives. 

2) In theory, the right thing to fix this would be
```cmake
llvm_map_components_to_libnames(llvm_libs all)
target_link_libraries(Enzyme-${LLVM_VERSION_MAJOR} ${llvm_libs})
```
However, "all" isn't being handled correctly, see https://github.com/llvm/llvm-project/issues/46347
It seems to be a common experience that some cmake macros are unreliable, so I was recommended
to just call the llvm-config binary.

3)
```
    execute_process(COMMAND ${LLVM_TOOLS_BINARY_DIR}/llvm-config --libs all
                OUTPUT_VARIABLE llvm_libraries)
```
Trying to also add the --ldflags fails the subsequent parsing:
```
CMake Error:
  Running

   '/usr/bin/ninja' '-C' '/home/manuel/prog/rust-new/build/x86_64-unknown-linux-gnu/enzyme/build' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:606: expected '=', got identifier

  -lLLVMWindowsManifest -lLLVMXRay -lLLVMLibDriver -lLLVMDlltoolDriver -lL...

                        ^ near here
```
so we have to add the `-L somedir`  in yet another command.

4) We therefore call
```
    set_target_properties(Enzyme-${LLVM_VERSION_MAJOR} PROPERTIES
        LINK_FLAGS "`${LLVM_TOOLS_BINARY_DIR}/llvm-config --ldflags`")
```
which correctly gives e.g. on my system `-L/home/manuel/prog/rust-new/build/x86_64-unknown-linux-gnu/llvm/lib`, where all the libLLVMSomething.a files are located.

So this still seems messy, but at least it should be more often correct than the current code. I'd be happy if someone else can fix it in upstream.

### DIFF
--- a/enzyme/Enzyme/CMakeLists.txt
+++ b/enzyme/Enzyme/CMakeLists.txt
@@ -145,7 +145,20 @@ if (${ENZYME_EXTERNAL_SHARED_LIB})
     add_dependencies(Enzyme-${LLVM_VERSION_MAJOR} BlasDeclarationsIncGen)
     add_dependencies(Enzyme-${LLVM_VERSION_MAJOR} BlasTAIncGen)
     add_dependencies(Enzyme-${LLVM_VERSION_MAJOR} BlasDiffUseIncGen)
-    target_link_libraries(Enzyme-${LLVM_VERSION_MAJOR} LLVM)
+
+    # This would be the desired way to link against LLVM components,
+    # however this function is bugged and does not work with `all`, see:
+    # https://github.com/llvm/llvm-project/issues/46347
+    #llvm_map_components_to_libnames(llvm_libs all)
+    # Therefore, manually invoke llvm-config
+    execute_process(COMMAND ${LLVM_TOOLS_BINARY_DIR}/llvm-config --libs all
+                OUTPUT_VARIABLE llvm_libraries)
+    string(STRIP "${llvm_libraries}" llvm_libraries)
+    # In theory, adding --libs should also add all the -l flags,
+    # but it isn't picked up correctly by clang, so we call target_link_libraries
+    set_target_properties(Enzyme-${LLVM_VERSION_MAJOR} PROPERTIES
+        LINK_FLAGS "`${LLVM_TOOLS_BINARY_DIR}/llvm-config --ldflags`")
+    target_link_libraries(Enzyme-${LLVM_VERSION_MAJOR} ${llvm_libraries})
     install(TARGETS Enzyme-${LLVM_VERSION_MAJOR}
         EXPORT EnzymeTargets
         LIBRARY DESTINATION lib COMPONENT shlib


ZuseZ4:
Also there is at least a 4th bug, but I think that one is on the rust side, where we at times pass the wrong llvm_dir (in cases like PGO builds, where we have multiple).

ZuseZ4:
[out.ll.txt](https://github.com/user-attachments/files/20029663/out.ll.txt)

ZuseZ4:
@wsmoses I don't understand how the julia build environment changes, especially the last commit was just a guess (since I don't know why zlib was previously commented out). However, it works. I'd probably try to again remove 
```
        llvm_map_components_to_libnames(llvm_librariess Passes Support)
        target_link_libraries(Enzyme-${LLVM_VERSION_MAJOR} ${llvm_librariess})
```
once CI passed, given that I suspect that the ZLIB removal was the likely culprit for the build failure. Then, we'd be fully relying on llvm-config.

enzyme/Enzyme/CMakeLists.txt
```diff
@@ -8,8 +8,10 @@ endif()
 
 get_target_property(TBL_LINKED_LIBS LLVMSupport INTERFACE_LINK_LIBRARIES)
 if (NOT TBL_LINKED_LIBS)
+message(STATUS "No TBL_LINKED_LIBS found")
 else()
-list(REMOVE_ITEM TBL_LINKED_LIBS "ZLIB::ZLIB")
+message(STATUS "TBL_LINKED_LIBS (test): ${TBL_LINKED_LIBS}")
```
wsmoses, line 13: I'm fairly certain this was intentionally removed due to a bug in LLVM's packaging of the dylib

ZuseZ4: How do you work around it, since linking failed on the Julia side without this line.
Also, any other requests before landing this?

wsmoses: let's do these things one at a time to confirm they're fine.

Can you leave this as a non-functional change (aka keep old zlib behavior) first, but add the non dylib linking, and then a follow up for zlib change?

enzyme/Enzyme/CMakeLists.txt
```diff
@@ -145,7 +147,36 @@ if (${ENZYME_EXTERNAL_SHARED_LIB})
...
+
+    if (${LLVM_BUILD_LLVM_DYLIB})
+        target_link_libraries(Enzyme-${LLVM_VERSION_MAJOR} LLVM)
+    else()
+        # This would be the desired way to link against LLVM components,
+        # however this function is bugged and does not work with `all`, see:
+        # https://github.com/llvm/llvm-project/issues/46347
+        #llvm_map_components_to_libnames(llvm_libraries Passes)
+        # Therefore, manually invoke llvm-config
+        if (EXISTS "${LLVM_TOOLS_BINARY_DIR}/llvm-config")
```
wsmoses, line 159: so the macos ci is hitting this print, which clearly it shouldn't be
```
-- TBL_LINKED_LIBS (test): m;ZLIB::ZLIB;LLVMDemangle
-- Using llvm-config from /Users/runner/.julia/artifacts/1344b3dbe1edda5d63922107ed3d06d2b436d713/tools
-- Linking against LLVM libraries: -lLLVMWindowsManifest -lLLVMWindowsDriver -lLLVMXRay -lLLVMLibDriver -lLLVMDlltoolDriver -lLLVMCoverage -lLLVMLineEditor -lLLVMX86TargetMCA -lLLVMX86Disassembler -lLLVMX86AsmParser -lLLVMX86CodeGen -lLLVMX86Desc -lLLVMX86Info -lLLVMAVRDisassembler -lLLVMAVRAsmParser -lLLVMAVRCodeGen -lLLVMAVRDesc -lLLVMAVRInfo -lLLVMBPFDisassembler -lLLVMBPFAsmParser -lLLVMBPFCodeGen -lLLVMBPFDesc -lLLVMBPFInfo -lLLVMWebAssemblyDisassembler -lLLVMWebAssemblyAsmParser -lLLVMWebAssemblyCodeGen -lLLVMWebAssemblyDesc -lLLVMWebAssemblyUtils -lLLVMWebAssemblyInfo -lLLVMOrcJIT -lLLVMMCJIT -lLLVMJITLink -lLLVMInterpreter -lLLVMExecutionEngine -lLLVMRuntimeDyld -lLLVMOrcTargetProcess -lLLVMOrcShared -lLLVMDWP -lLLVMDebugInfoGSYM -lLLVMOption -lLLVMObjectYAML -lLLVMObjCopy -lLLVMMCA -lLLVMMCDisassembler -lLLVMLTO -lLLVMPasses -lLLVMCFGuard -lLLVMCoroutines -lLLVMObjCARCOpts -lLLVMipo -lLLVMVectorize -lLLVMLinker -lLLVMInstrumentation -lLLVMFrontendOpenMP -lLLVMFrontendOpenACC -lLLVMExtensions -lLLVMDWARFLinker -lLLVMGlobalISel -lLLVMMIRParser -lLLVMAsmPrinter -lLLVMSelectionDAG -lLLVMCodeGen -lLLVMIRReader -lLLVMAsmParser -lLLVMInterfaceStub -lLLVMFileCheck -lLLVMFuzzMutate -lLLVMTarget -lLLVMScalarOpts -lLLVMInstCombine -lLLVMAggressiveInstCombine -lLLVMTransformUtils -lLLVMBitWriter -lLLVMAnalysis -lLLVMProfileData -lLLVMSymbolize -lLLVMDebugInfoPDB -lLLVMDebugInfoMSF -lLLVMDebugInfoDWARF -lLLVMObject -lLLVMTextAPI -lLLVMMCParser -lLLVMMC -lLLVMDebugInfoCodeView -lLLVMBitReader -lLLVMFuzzerCLI -lLLVMCore -lLLVMRemarks -lLLVMBitstreamReader -lLLVMBinaryFormat -lLLVMTableGen -lLLVMSupport -lLLVMDemangle
-- Linking against LLVM ldflags: -L/Users/runner/.julia/artifacts/1344b3dbe1edda5d63922107ed3d06d2b436d713/lib -Wl,-search_paths_first -Wl,-headerpad_max_install_names
```

PR #2303, opened 2025-05-05
tblgen: Auto-generate ActivityOpInterface

### LINKED ISSUE #2301: MLIR: autogenerate activity interface from argument spec
e.g. select being inactive, (expr), (expr), should let us define the activity interface to register arg 0 as inactive

Pangoraw:
Closes #2301.

Example of the generated interface:
```cpp
struct GatherOpActivity : 
                   public ActivityOpInterface::ExternalModel<GatherOpActivity, stablehlo::GatherOp> {
  bool isInactive(mlir::Operation *) const { return false; }
  bool isArgInactive(mlir::Operation *op, size_t arg) const {
    if (arg == 1) { return true; }
    return false;
  }
};
struct SelectOpActivity : 
                   public ActivityOpInterface::ExternalModel<SelectOpActivity, stablehlo::SelectOp> {
  bool isInactive(mlir::Operation *) const { return false; }
  bool isArgInactive(mlir::Operation *op, size_t arg) const {
    if (arg == 0) { return true; }
    return false;
  }
};

### DIFF
--- a/enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
+++ b/enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
@@ -2597,6 +2597,49 @@ static void emitDerivatives(const RecordKeeper &recordKeeper, raw_ostream &os,
   }
 
   if (intrinsic == MLIRDerivatives) {
+    SmallVector<bool> hasActivity(patterns.size(), false);
+    for (auto patternEn : enumerate(patterns)) {
+      size_t patternIdx = patternEn.index();
+      auto &pattern = patternEn.value();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by wsmoses** [comment]
sadly:
```
/home/runner/work/Enzyme/Enzyme/enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp:2729:32: error: ‘zip_equal’ was not declared in this scope
 2729 |     for (auto [pattern, act] : zip_equal(patterns, hasActivity)) {
 ```
 
we need to use regular zip as it doesn't exist in LLVM15

wsmoses:
or alternatively version gate it for LLVM > 15, since MLIR is only supported on latest llvm

PR #2328, opened 2025-05-29
feat: op to detach tensor [MLIR]

avik-pal:


enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -313,7 +313,9 @@ const std::set<Intrinsic::ID> KnownInactiveIntrinsics = {
     Intrinsic::roundeven,
     Intrinsic::lround,
     Intrinsic::llround,
+#if LLVM_VERSION_MAJOR <= 20
     Intrinsic::nvvm_barrier0,
```
wsmoses, line 317: It’s important we handle the new intrinsic, was that already in the list

wsmoses: This is equivalent to cuda sync threads for contact

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -222,4 +222,11 @@ def SampleOp : Enzyme_Op<"sample",
   }];
 }
 
+def DetachGradientOp : Enzyme_Op<"detach_gradient",
```
wsmoses, line 225: I’m not that big a fan of the name detach here, maybe block or stop

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -222,4 +222,11 @@ def SampleOp : Enzyme_Op<"sample",
...
+    [Pure, SameOperandsAndResultElementType, SameOperandsAndResultShape]> {
+  let summary = "Prevents the flow of gradients by creating a new tensor that is detached from the original tensor.";
```
wsmoses, line 227: Replace value with tensor. And maybe specify that this is an identity operation in the primal , and that this block blocks all higher order derivatives as well

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
wsmoses: Cc @vchuravy as well for name suggestions (well do the same for enzymellvm)

enzyme/Enzyme/ActivityAnalysis.cpp
avik-pal: from the PR `nvvm_barrier_cta_sync_aligned_all` seems to be the one?

wsmoses: Yeah I think so

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
avik-pal: - Pytorch -- `detach`
- ChainRules -- `ignore_derivatives`
- Jax -- `stop_gradient`

I prefer the chainrules one (suggests all derivatives including higher-order)

avik-pal: done

PR #2333, opened 2025-05-31
Fix Fwddiff canonicalization [MLIR]

vimarsh6739:


enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -261,6 +261,10 @@ class FwdRetOpt final : public OpRewritePattern<ForwardDiffOp> {
                 rewriter.getContext(),
                 mlir::enzyme::Activity::enzyme_constnoneed);
             newRetActivityArgs.push_back(new_constnn);
+          } else {
```
wsmoses, line 264: can you add a test for what was missing here/

PR #2339, opened 2025-06-06
fix: remove ignore_derivatives pass

avik-pal:
cc @wsmoses


enzyme/Enzyme/MLIR/Passes/SimplifyMath.cpp
```diff
@@ -37,13 +37,26 @@ struct ApplySimplificationPattern
   }
 };
 
+struct ApplyRemoveIgnoreDerivativePattern
```
wsmoses: can we move this to remove-enzyme-ops instead of enzyme-simplify-math?

avik-pal: done

PR #2348, opened 2025-06-12
[mlir] SCF reverse-mode derivatives

pengmai:


enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -415,6 +415,90 @@ struct ForOpEnzymeOpsRemover
...
+            if (!gutils->isConstantValue(res)) {
+              gutils->zeroDiffe(res, bodyBuilder);
+            }
+          }
+        }
+
+        auto term = oBB.getTerminator();
+
+        for (auto &&[arg, operand] :
+             llvm::zip_equal(incomingGradients, term->getOperands())) {
```
Pangoraw: IIUC, incomingGradients can be smaller than `term->getOperands()` if at least one result is inactive

pengmai: Good catch, will fix and add a test

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -419,6 +419,102 @@ struct ForOpEnzymeOpsRemover
...
+      }
+    }
+    return success(valid);
+  }
+
+  SmallVector<Value> cacheValues(Operation *op,
+                                 MGradientUtilsReverse *gutils) const {
+    auto ifOp = cast<scf::IfOp>(op);
+
+    Operation *newOp = gutils->getNewFromOriginal(op);
```
wsmoses: Nit but if all the inner operations and resulting values are inactive we can skip storing this

wsmoses (approved):

wsmoses (approved):

Issue #2194, opened 2024-12-10
MLIR: Autodiff-specific optimizations


wsmoses:
- [ ] Don't compute the derivative if it is not used in the program

PR #2349, opened 2025-06-16
[MLIR] Forward Mode arg conversion for dup -> const

vimarsh6739:
#2194

### DIFF
--- a/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
+++ b/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
@@ -256,7 +256,7 @@ public:
         auto ETintf = dyn_cast<AutoDiffTypeInterface>(ET);
 
         if (ETintf && !isMutable(ET) && ETintf.isZero(inp).succeeded()) {
-        // skip and promote to const
+          // skip and promote to const
           auto new_const = mlir::enzyme::ActivityAttr::get(
               rewriter.getContext(), mlir::enzyme::Activity::enzyme_const);
           newInActivityArgs.push_back(new_const);
--- a/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
+++ b/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
@@ -203,7 +203,6 @@ public:
 
     auto in_idx = 0;
     SmallVector<mlir::Value, 2> in_args;
-    SmallVector<Type, 2> in_ty;
     SmallVector<ActivityAttr, 2> newInActivityArgs;
     bool changed = false;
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -187,6 +175,118 @@ static inline bool isMutable(Type type) {
...
+
+      case Activity::enzyme_dupnoneed: {
+        // always pass in primal
+        in_args.push_back(inp);
+        in_idx++;
+
+        // selectively push or skip directional derivative
+        inp = uop.getInputs()[in_idx];
+        if (!isMutable(inp.getType()) &&
+            (matchPattern(inp, m_Zero()) ||
```
wsmoses: @avik-pal @Pangoraw do you remember the is zero we had set up [since this won't applies for complex as is]

wsmoses: it feels like this is worthwhile making into an interface for the type

wsmoses: see https://github.com/EnzymeAD/Enzyme/pull/2355/files#diff-0a1e6c2523b1ed10a4bfba6f1a5208ba0eb2a444446c8fea0f653bf6a810719c for an example [notably getApproxSize]

enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -187,6 +175,120 @@ static inline bool isMutable(Type type) {
...
+        // always pass in primal
+        in_args.push_back(inp);
+        in_idx++;
+
+        // selectively push or skip directional derivative
+        inp = uop.getInputs()[in_idx];
+        auto ET = inp.getType();
+        auto ETintf = dyn_cast<AutoDiffTypeInterface>(ET);
+
+        if (ETintf && !isMutable(ET) && ETintf.isZero(inp).succeeded()) {
```
wsmoses, line 233: we should probably have made isZero return a bool, not a LogicalResult

wsmoses (approved):
lgtm modulo comment about isZero interface return type

wsmoses:
also errors here are genuine:

```
/__w/Enzyme/Enzyme/Enzyme/enzyme/Enzyme/MLIR/Dialect/Ops.cpp:201:10: error: variable 'retActivity' set but not used [-Werror=unused-but-set-variable]
  201 |     auto retActivity = uop.getRetActivity();
      |          ^~~~~~~~~~~
```

vimarsh6739:
Ah that I can remove

PR #2356, opened 2025-06-20
Add C API for setting type-analysis-func via EnzymeSetCLString

KMJ-007:
This PR introduces a C API to dynamically set the type-analysis-func flag (FunctionToAnalyze).

## Changes:
- Added: EnzymeSetCLString(void* ptr, const char* val) to CApi.cpp and CApi.h
- This enables users to set string-based LLVM command-line options at runtime.

### DIFF
--- a/enzyme/Enzyme/TypeAnalysis/TypeAnalysisPrinter.cpp
+++ b/enzyme/Enzyme/TypeAnalysis/TypeAnalysisPrinter.cpp
@@ -64,10 +64,12 @@ using namespace llvm;
 #endif
 #define DEBUG_TYPE "type-analysis-results"
 
+extern "C" {
 /// Function ActivityAnalysis will be starting its run from
 llvm::cl::opt<std::string>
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/TypeAnalysis/TypeAnalysisPrinter.cpp
```diff
@@ -64,10 +64,12 @@ using namespace llvm;
 #endif
 #define DEBUG_TYPE "type-analysis-results"
 
+extern "C" {
 /// Function ActivityAnalysis will be starting its run from
 llvm::cl::opt<std::string>
     FunctionToAnalyze("type-analysis-func", cl::init(""), cl::Hidden,
```
wsmoses: can you change the name of this, perhaps EnzymeFunctionToAnalyze, if it is expected to be exported now

KMJ-007: done

wsmoses (approved):

PR #2359, opened 2025-06-26
mlir: enable mincut support for scf.for

Pangoraw:


enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -97,16 +97,20 @@ struct ForOpEnzymeOpsRemover
...
-      caches.push_back(info);
-    }
+    SmallVector<CacheInfo> caches =
+        llvm::map_to_vector(cachesMap, [](auto p) { return std::get<1>(p); });
 
     // nothing to do
     if (updatedGradients.empty() && caches.empty())
       return success();
 
+    if (forOp->hasAttr("enzyme.enable_mincut")) {
```
wsmoses, line 107: i think we should make this the default [and optionally disable]

wsmoses: we'll also want to add this to affine.for/{scf,affine}.parallel

enzyme/test/MLIR/ReverseMode/scf_for_mincut.mlir
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// CHECK:  func.func @main(%arg0: f32, %arg1: f32) -> f32 {
+// CHECK-NEXT:    %c9 = arith.constant 9 : index
+// CHECK-NEXT:    %c1 = arith.constant 1 : index
+// CHECK-NEXT:    %c10 = arith.constant 10 : index
+// CHECK-NEXT:    %c0 = arith.constant 0 : index
+// CHECK-NEXT:    %[[v0:.+]] = tensor.empty() : tensor<10xf32>
+// CHECK-NEXT:    %[[for:.+]]:2 = scf.for %arg2 = %c0 to %c10 step %c1 iter_args(%arg3 = %arg0, %arg4 = %[[v0]]) -> (f32, tensor<10xf32>) {
+// CHECK-NEXT:      %[[cache:.+]] = tensor.insert %arg3 into %arg4[%arg2] : tensor<10xf32>
```
wsmoses, line 27: we should likely prefer using a memref alloca of the correct size, store in fwd, load in reverse, and free after the for

wsmoses:
@Pangoraw I'm going to merge for now, can you address in post commit follow up

enzyme/test/MLIR/ReverseMode/scf_for_mincut.mlir
Pangoraw: it will create a cache<memref<Nxf32>> for the parent op so the alloca has to be hoisted all the way to the top (and would be invalid for upcoming autodiff deferred). IIRC, that's why I went with an immutable data structure at the time.

We can write a bufferization pass that converts this to memref once all push/pop have been removed ?

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
Pangoraw: makes sense, i will follow-up with those (i don't think we have the rev rules for these ops yet)

PR #2364, opened 2025-06-29
Reverse mode output args conversions

sbrantq:


vimarsh6739 (approved):
@wsmoses We can add interface conversion for zero checking in a separate change

PR #2373, opened 2025-07-02
[MLIR] Add `isZero` interface

vimarsh6739:


enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
```diff
@@ -134,6 +143,44 @@ class TensorTypeInterface
...
+  LogicalResult isZero(Type self, Value val) const {
+    auto tenType = cast<TensorType>(self);
+    auto ET = tenType.getElementType();
+    DenseElementsAttr eAttr;
+
+    if (!matchPattern(val, m_Constant(&eAttr)))
+      return failure();
+
+    if (eAttr.isSplat()) {
+      auto splatVal = eAttr.getSplatValue<Attribute>();
```
wsmoses, line 163: I think if it's not a splat, it would be hard for them to be all zeros (and we can ignore that branch)

enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
```diff
@@ -134,6 +143,44 @@ class TensorTypeInterface
...
+
+      if (isa<IntegerType>(ET)) {
+        return matchPattern(splatVal, m_Zero()) ? success() : failure();
```
wsmoses: we should ideally recurse here, on the elementType.

so we should also have a version of isZero that takes an attribute, not just the value

vimarsh6739:
@wsmoses can you take a look now. I think I handled all the cases

enzyme/Enzyme/MLIR/Implementations/BuiltinAutoDiffTypeInterfaceImpl.cpp
```diff
@@ -216,6 +278,33 @@ class ComplexTypeInterface
...
+    return failure();
+  }
+
+  LogicalResult isZeroAttr(Type self, Attribute attr) const {
+    auto arrayAttr = dyn_cast<ArrayAttr>(attr);
+    if (!arrayAttr || arrayAttr.size() != 2)
+      return failure();
+
+    // Check that both elements of the complex number are zero.
+    bool firstIsZero = cast<FloatAttr>(arrayAttr[0]).getValue().isZero();
```
wsmoses: you can have complex of int, this should recurse

vimarsh6739: Used recursion for both now

Issue #2367, opened 2025-07-01
[Bug] `unordered_map` access


ax3l:
I encounter an issue in [ImpactX](https://github.com/BLAST-ImpactX/impactx/pull/825) when differentiating through functions with return types of value `std::unordered_map<std::string, double>`.

I use those dictionary style types to return named tuples of ~20 values in certain physics (analysis) routines, usually then optimizing on a select few parameters in them. Unfortunately, I get some LLVM error that I do not (yet) know how to read (`error: Enzyme: Cannot deduce type of copy call`) .

Here is the full reproducer:
https://fwd.gymni.ch/DTQR3r

```c++
#include <cmath>
#include <string>
#include <unordered_map>


std::unordered_map<std::string, double>
reduced_beam_characteristics (double x, double y)
{
    std::unordered_map<std::string, double> data;

    data["alpha_x"] = x*x + y;
    data["alpha_y"] = y*y + 2.;

    return data;
}

double compute (double r)
{
    double x = std::sqrt(r);
    double y = std::log(r);

    std::unordered_map<std::string, double> const rbc =
        reduced_beam_characteristics(x, y);
    return rbc.at("alpha_x");
}

double __enzyme_autodiff(void*, double);

int main()
{
    double q1_k = 3.0;

    // normal
    double const alpha_x = compute(q1_k);

    // diff (fails)
    double ddx = __enzyme_autodiff((void*) compute, q1_k);

}
```

<details>
In file included from <source>:2:
In file included from /usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/string:40:
/usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/char_traits.h:437:33: error: Enzyme: Cannot deduce type of copy   call void @llvm.memcpy.p0.p0.i64(ptr noundef nonnull align 8 dereferenceable(7) %6, ptr noundef nonnull align 1 dereferenceable(7) @.str, i64 7, i1 false) #20, !dbg !3616
<analysis>
  %39 = load i64, ptr %38, align 8, !dbg !3931, !tbaa !3932: {[-1]:Integer}, intvals: {}
double %0: {[-1]:Float@double}, intvals: {}
  %27 = load ptr, ptr %25, align 8, !dbg !3783, !tbaa !3791: {[-1]:Pointer}, intvals: {}
  call void @_ZdlPv(ptr noundef %29) #21, !dbg !3877: {}, intvals: {}
[...]

PR #2376, opened 2025-07-07
Test: Unorderded Map Access

ax3l:
## Summary

Seeking help for the issue in #2367 by providing these unit tests.

This test demonstrates two issues with `std::unordered_map`:
- `at()` fails to compile
- `operator[]` has a memory leak

## To Do

Please feel free to push to this PR as needed :pray: I will also try to mingle around with it a bit, but am still very early in understanding what's going on :)

- [ ] add fix for compile issue with `at()`
- [ ] add fix for memory leak in `operator[]`


enzyme/test/Integration/ReverseMode/stl_umap.cpp
```diff
@@ -0,0 +1,82 @@
+// RUN: %clang++ -std=c++11 -fno-exceptions -ffast-math -O0 %s -S -emit-llvm -o - | %opt - %OPloadEnzyme %enzyme -S | %lli -
```
ax3l: `-O0` fails to compile even for `operator[]`, likely because it strumbles over the map insert that is not optimized out (which would be good to support in more general cases, too):
https://fwd.gymni.ch/R6Q2bQ

PR #2385, opened 2025-07-18
Introduce __enzyme_ignore_derivatives

vchuravy:
The goal is to provide a function `y = ignore_derivative(x)` that will detach uses of `y` from the derivative calculations for `x`

x-ref: https://github.com/EnzymeAD/Enzyme.jl/pull/2479#issuecomment-3089738818

### DIFF
--- a/enzyme/Enzyme/CallDerivatives.cpp
+++ b/enzyme/Enzyme/CallDerivatives.cpp
@@ -3659,30 +3659,6 @@ bool AdjointGenerator::handleKnownCallDerivatives(
     return true;
   }
 
-  if (funcName.contains("__enzyme_ignore_derivatives")) {
-    if (gutils->isConstantValue(&call)) {
-      eraseIfUnused(call);
-      return true;
-    }
-
-    auto ifound = gutils->invertedPointers.find(&call);
-    assert(ifound != gutils->invertedPointers.end());
-
-    auto placeholder = cast<PHINode>(&*ifound->second);
-
-    bool needShadow =
-        DifferentialUseAnalysis::is_value_needed_in_reverse<QueryType::Shadow>(
-            gutils, &call, Mode, oldUnreachable);
-    if (!needShadow) {
-      gutils->invertedPointers.erase(ifound);
-      gutils->erase(placeholder);
-      eraseIfUnused(call);
-      return true;
-    }
-
-    // TODO: How to create a shadow from nothing?
-  }
-
   if (funcName == "memcpy" || funcName == "memmove") {
     auto ID = (funcName == "memcpy") ? Intrinsic::memcpy : Intrinsic::memmove;
     visitMemTransferCommon(ID, /*srcAlign*/ MaybeAlign(1),
--- a/enzyme/Enzyme/CallDerivatives.cpp
+++ b/enzyme/Enzyme/CallDerivatives.cpp
@@ -3659,6 +3659,30 @@ bool AdjointGenerator::handleKnownCallDerivatives(
     return true;
   }
 
+  if (funcName.contains("__enzyme_ignore_derivatives")) {
+    if (gutils->isConstantValue(&call)) {
+      eraseIfUnused(call);
+      return true;
+    }
+
+    auto ifound = gutils->invertedPointers.find(&call);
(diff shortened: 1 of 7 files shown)


enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -3659,6 +3659,30 @@ bool AdjointGenerator::handleKnownCallDerivatives(
     return true;
   }
 
+  if (funcName.contains("__enzyme_ignore_derivatives")) {
+    if (gutils->isConstantValue(&call)) {
+      eraseIfUnused(call);
+      return true;
```
wsmoses: we should also add to activity analysis proper

vchuravy: Where about? I grepped for `todense` and followed those breadcrumbs.

enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -3659,6 +3659,30 @@ bool AdjointGenerator::handleKnownCallDerivatives(
...
+        DifferentialUseAnalysis::is_value_needed_in_reverse<QueryType::Shadow>(
+            gutils, &call, Mode, oldUnreachable);
+    if (!needShadow) {
+      gutils->invertedPointers.erase(ifound);
+      gutils->erase(placeholder);
+      eraseIfUnused(call);
+      return true;
+    }
+
+    // TODO: How to create a shadow from nothing?
```
vchuravy: @wsmoses this is the big unknown for me.

enzyme/Enzyme/CallDerivatives.cpp
wsmoses: Activity analysis.cpp

enzyme/Enzyme/CallDerivatives.cpp
wsmoses: Marking it inactive in activity analysis means we never will need to make a shadow

vchuravy: So I am worried about that since it will easily cause dynamic activity issues.

enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -3659,6 +3659,30 @@ bool AdjointGenerator::handleKnownCallDerivatives(
     return true;
   }
 
+  if (funcName.contains("__enzyme_ignore_derivatives")) {
```
wsmoses: I think we ought be able to remove this

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -4261,7 +4261,7 @@ bool GradientUtils::shouldRecompute(const Value *val,
         n == "tanhf" || n == "__pow_finite" ||
         n == "julia.pointer_from_objref" || startsWith(n, "enzyme_wrapmpi$$") ||
         n == "omp_get_thread_num" || n == "omp_get_max_threads" ||
-        startsWith(n, "_ZN4libm4math3log")) {
+        startsWith(n, "_ZN4libm4math3log") || "__enzyme_ignore_derivatives") {
```
vchuravy:
```suggestion
        startsWith(n, "_ZN4libm4math3log") || n == "__enzyme_ignore_derivatives") {
```

enzyme/test/Enzyme/ReverseMode/ignore_derivatives.ll
```diff
@@ -0,0 +1,55 @@
+; RUN: if [ %llvmver -lt 16 ]; then %opt < %s %loadEnzyme -enzyme -enzyme-preopt=false -S | FileCheck %s; fi
+; RUN: %opt < %s %newLoadEnzyme -passes="enzyme" -enzyme-preopt=false -S | FileCheck %s
+
+; Function Attrs: nounwind readnone uwtable
+define double @tester(double %x) {
+entry:
+  %y = tail call fast double @__enzyme_ignore_derivatives(double %x)
```
vchuravy, line 7:
```suggestion
  %y = tail call fast double @__enzyme_ignore_derivatives.fp64(double %x)
```

PR #2420, opened 2025-08-24
[MLIR] Better support for llvm.func and reverse-mode differentiation of pointers

pengmai:


enzyme/Enzyme/MLIR/Interfaces/EnzymeLogicReverse.cpp
```diff
@@ -239,7 +248,14 @@ FunctionOpInterface MEnzymeLogic::CreateReverseDiff(
         retargs.push_back(gutils->diffe(arg, builder));
       }
     }
-    builder.create<func::ReturnOp>(oBB->rbegin()->getLoc(), retargs);
+
+    if (isa<LLVM::LLVMFuncOp>(fn)) {
+      assert(retargs.size() <= 1 && "LLVM func ops assume at most 1 return; "
```
wsmoses: We can do the same trick as in llvm and make an anonymous struct of all values

wsmoses (approved):

enzyme/Enzyme/MLIR/Interfaces/GradientUtils.cpp
```diff
@@ -200,19 +200,16 @@ void mlir::enzyme::MGradientUtils::setDiffe(mlir::Value val, mlir::Value toset,
...
-    auto placeholder = found.getDefiningOp<enzyme::PlaceholderOp>();
-    invertedPointers.erase(val);
-    // replaceAWithB(placeholder, toset);
-    placeholder.replaceAllUsesWith(toset);
-    erase(placeholder);
-    invertedPointers.map(val, toset);
-    return;
-  }
+
+  assert(getShadowType(val.getType()) == toset.getType());
```
wsmoses: this is not correct, the convention for setDiffe in reverse mode is to store into the differenital alloca [see the commented out code at the bottom of the function]

wsmoses: the commented out code below should likely just be a memref store op (or llvm store op), depending on type

pengmai: Should a different function be used when setting the differential value of a pointer/memref (such that the original type and shadow type are the same)? I'm also seeing that `MDiffeGradientUtils::setDiffe` (in contrast to `MGradientUtils::setDiffe`) correctly emits an `enzyme::SetOp` when the value type isn't mutable, which looks like the store you're describing.

I needed to make this change specifically to get pointer casts and offsets (`memref.subview` and `enzymexla.pointer2memref`) working

wsmoses: yeah we can make a different helper function, setInvertedPointer or something [everywhere in enzyme proper/etc just does the same inverted pointer map rewrite by hand]

Issue #2429, opened 2025-08-31
Add CI to test rust std::autodiff


sgasho:
related: https://github.com/rust-lang/rust/issues/145899

PR #2430, opened 2025-08-31
ci: add a github action ci to test Rust std::autodiff

sgasho:
closes: #2429
related: https://github.com/rust-lang/rust/issues/145899

### DIFF
--- a/.github/workflows/enzyme-rust.yml
+++ b/.github/workflows/enzyme-rust.yml
@@ -1,0 +1,62 @@
+name: Enzyme Rust CI
+
+on: 
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  merge_group:
+
+jobs:
+  rust-autodiff:
+    name: Rust Autodiff Tests LLVM ${{ matrix.llvm }}
+    runs-on: [self-hosted, Linux, X64, 32-core]
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        llvm: ["19", "20"]
+    
+    timeout-minutes: 90
+    
+    steps:
+    - name: Install LLVM and dependencies
+      run: |
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+        sudo apt-add-repository "deb http://apt.llvm.org/`lsb_release -c | cut -f2`/ llvm-toolchain-`lsb_release -c | cut -f2`-${{ matrix.llvm }} main" || true
+        sudo apt-get install -y cmake gcc g++ llvm-${{ matrix.llvm }}-dev libomp-${{ matrix.llvm }}-dev lld-${{ matrix.llvm }} clang-${{ matrix.llvm }} libclang-${{ matrix.llvm }}-dev libeigen3-dev libboost-dev libzstd-dev
+        sudo python3 -m pip install --upgrade pip lit
+    - name: Install Rust toolchain
+      run: |
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        source ~/.cargo/env
+        rustup toolchain install nightly
+        rustup default nightly
+    - uses: actions/checkout@v4
+      name: Checkout Enzyme
+    - name: Build Enzyme
+      run: |
+        rm -rf build && mkdir build && cd build
+        cmake ../enzyme -DLLVM_EXTERNAL_LIT=`which lit` -DCMAKE_BUILD_TYPE=Release -DLLVM_DIR=/usr/lib/llvm-${{ matrix.llvm }}/lib/cmake/llvm
+        make -j `nproc` LLVMEnzyme-${{ matrix.llvm }}
+    - name: Clone and configure Rust compiler
+      run: |
+        source ~/.cargo/env
+        git clone --depth 1 --branch master https://github.com/rust-lang/rust.git
+        cd rust
+        ./configure --enable-llvm-link-shared --enable-llvm-plugins --enable-llvm-enzyme --release-channel=nightly --enable-llvm-assertions --enable-clang --enable-lld --enable-option-checking --enable-ninja --disable-docs
+    - name: Build Rust compiler stage 1
+      working-directory: rust
+      run: |
+        source ~/.cargo/env
+        ./x build --stage 1 library
+    - name: Run Rust autodiff tests
+      working-directory: rust
+      run: |
+        source ~/.cargo/env
+        ./x test --stage 1 tests/codegen-llvm/autodiff
+        ./x test --stage 1 tests/pretty/autodiff
+        ./x test --stage 1 tests/ui/autodiff
+        ./x test --stage 1 tests/ui/feature-gates/feature-gate-autodiff.rs


.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,62 @@
...
+      run: |
+        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+        sudo apt-add-repository "deb http://apt.llvm.org/`lsb_release -c | cut -f2`/ llvm-toolchain-`lsb_release -c | cut -f2`-${{ matrix.llvm }} main" || true
+        sudo apt-get install -y cmake gcc g++ llvm-${{ matrix.llvm }}-dev libomp-${{ matrix.llvm }}-dev lld-${{ matrix.llvm }} clang-${{ matrix.llvm }} libclang-${{ matrix.llvm }}-dev libeigen3-dev libboost-dev libzstd-dev
+        sudo python3 -m pip install --upgrade pip lit
+    - name: Install Rust toolchain
+      run: |
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        source ~/.cargo/env
+        rustup toolchain install nightly
```
wsmoses, line 42: until Rust adds CI for Enzyme, I think this should be fixed to a specific commit

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,62 @@
...
+  merge_group:
+
+jobs:
+  rust-autodiff:
+    name: Rust Autodiff Tests LLVM ${{ matrix.llvm }}
+    runs-on: [self-hosted, Linux, X64, 32-core]
+    
+    strategy:
+      fail-fast: false
+      matrix:
```
wsmoses, line 18: you don't need llvm versions here, right?

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,62 @@
...
+      - main
+  pull_request:
+    branches:
+      - main
+  merge_group:
+
+jobs:
+  rust-autodiff:
+    name: Rust Autodiff Tests LLVM ${{ matrix.llvm }}
+    runs-on: [self-hosted, Linux, X64, 32-core]
```
wsmoses: this is not the correct setup for the runner, see https://github.com/EnzymeAD/Enzyme/blob/main/.github/workflows/enzyme-mlir.yml for an example

.github/workflows/enzyme-rust.yml
ZuseZ4: rustc vendors llvm. It supports the last two llvm versions (so currently 20+21), but I never tested that locally.
For simplicity I'd recommend to just use the vendored (default) for now. We could later improve CI to share LLVM builds, but let's postpone that extra complexity for now.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,62 @@
...
+        rustup default nightly
+    - uses: actions/checkout@v4
+      name: Checkout Enzyme
+    - name: Build Enzyme
+      run: |
+        rm -rf build && mkdir build && cd build
+        cmake ../enzyme -DLLVM_EXTERNAL_LIT=`which lit` -DCMAKE_BUILD_TYPE=Release -DLLVM_DIR=/usr/lib/llvm-${{ matrix.llvm }}/lib/cmake/llvm
+        make -j `nproc` LLVMEnzyme-${{ matrix.llvm }}
+    - name: Clone and configure Rust compiler
+      run: |
```
wsmoses, line 61: this is never told to use the enzyme we built above?

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,62 @@
...
+        source ~/.cargo/env
+        git clone --depth 1 --branch master https://github.com/rust-lang/rust.git
+        cd rust
+        ./configure --enable-llvm-link-shared --enable-llvm-plugins --enable-llvm-enzyme --release-channel=nightly --enable-llvm-assertions --enable-clang --enable-lld --enable-option-checking --enable-ninja --disable-docs
```
wsmoses: we should also use ccache here, and cache intermediate artifacts

.github/workflows/enzyme-rust.yml
wsmoses: it needs to run on `linux-x86-n2-32`

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,98 @@
...
+          --enable-clang \
+          --enable-option-checking \
+          --enable-ninja \
+          --disable-docs \
+          --enable-ccache
+    - name: Setup ccache
+      run: |
+        echo "CCACHE_DIR=$HOME/.ccache" >> $GITHUB_ENV
+        echo "CC=ccache gcc" >> $GITHUB_ENV
+        echo "CXX=ccache g++" >> $GITHUB_ENV
```
I-Al-Istannen: Who uses this? What exactly is built using GCC? I think Rust still uses LLVM. The rust invocation will prefix its invocations with `ccache` by itself as far as I can see, thanks to the `--enable-ccache` a few lines above.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,98 @@
...
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [linux-x86-n2-32]
+    
+    timeout-minutes: 90
+
+    container:
+      image: ${{ (contains(matrix.os, 'linux') && 'ghcr.io/enzymead/reactant-docker-images@sha256:91e1edb7a7c869d5a70db06e417f22907be0e67ca86641d48adcea221fedc674' ) || '' }}
```
I-Al-Istannen, line 25: If you arbitrarily(?) pin this to a hash here, who will update it? Does it matter, if it becomes outdated? Is it picked up by dependabot?

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,98 @@
...
+    - name: Build Rust compiler stage 1
+      working-directory: rust
+      run: |
+        . ~/.cargo/env
+        ./x build --stage 1 library
+    - name: Build Enzyme with Rust's LLVM
+      run: |
+        . ~/.cargo/env
+        rm -rf build && mkdir build && cd build
+        RUST_LLVM_DIR=${GITHUB_WORKSPACE}/rust/build/x86_64-unknown-linux-gnu/llvm
```
I-Al-Istannen: I would shell-quote this. GitHub probably has no space or special chars in the checkout dir, but better be safe.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,98 @@
...
+    steps:
+    - name: Install dependencies
+      run: |
+        apt-get update
+        apt-get install -y binutils ninja-build cmake gcc g++ python3 python3-dev ccache nodejs npm
+    - name: Install Rust toolchain
+      run: |
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        . ~/.cargo/env
+        # Pin to specific nightly commit for reproducibility
```
I-Al-Istannen: What is this used for? Rustc bootstrap? Might be worth it to add a comment saying *what* is now reproducible. Enzyme itself is built against the latest rust compiler (you check out the main branch of rust-lang/rust), so I imagine it won't really help there?

.github/workflows/enzyme-rust.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     timeout-minutes: 90
 
     container:
-      image: ${{ (contains(matrix.os, 'linux') && 'ghcr.io/enzymead/reactant-docker-images@sha256:91e1edb7a7c869d5a70db06e417f22907be0e67ca86641d48adcea221fedc674' ) || '' }}
+      image: ${{ (contains(matrix.os, 'linux') && 'ghcr.io/enzymead/reactant-docker-images:main' ) || '' }}
```
wsmoses: this should use the specific hash, not main

.github/workflows/enzyme-rust.yml
sgasho: is this comment conflicting to this below?
https://github.com/EnzymeAD/Enzyme/pull/2430#discussion_r2319220575

Is it correct to understand that when a specific hash is required, we must accept the drawbacks associated with updates?

.github/workflows/enzyme-rust.yml
sgasho: I shell-quoted it in this commit!
[use quotes](https://github.com/EnzymeAD/Enzyme/pull/2430/commits/f6c53f9f0e4b451f1565ace27f39d1d9dc01cf14)

.github/workflows/enzyme-rust.yml
I-Al-Istannen: I am not involved in Enzyme, wsmoses knows enzyme a lot better :)

That were mostly some questions that the workflow should ideally answer, lest they crop up in a few weeks to months when nobody knows why things are frozen at some particular point.

Ideally there is a reason for freezing the ubuntu image at a particular point (and some mechanism to update it at some point), but that's not something I can answer you.

wsmoses: mostly security concerns for the self-hosted runners.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,89 @@
...
+    timeout-minutes: 90
+
+    container:
+      image: ${{ (contains(matrix.os, 'linux') && 'ghcr.io/enzymead/reactant-docker-images@sha256:91e1edb7a7c869d5a70db06e417f22907be0e67ca86641d48adcea221fedc674' ) || '' }}
+    
+    steps:
+    - name: Install dependencies
+      run: |
+        apt-get update
+        apt-get install -y binutils ninja-build cmake gcc g++ python3 python3-dev ccache nodejs npm
```
wsmoses: do you need all these deps?

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,89 @@
...
+        rustup toolchain install nightly
+        rustup default nightly
+    - uses: actions/checkout@v4
+      name: Checkout Enzyme
+    - name: Clone and configure Rust compiler
+      run: |
+        . ~/.cargo/env
+        git clone https://github.com/rust-lang/rust.git
+        cd rust
+        git checkout 51ff895062ba60a7cba53f57af928c3fb7b0f2f4
```
wsmoses: can you move the commit to the matrix? that'll make it easier to update, and or if in the future we want to checkout multiple and/or stable tags

ZuseZ4:
This currently builds Enzyme twice, once as part of rustc and once standalone.
Unfortunately it uses the old one from rustc for testing.
Is there a way to delete the default Enzyme submodule and re-add it using the new Enzyme source code which we have here (e.g. through ln -s or copying)? Then you can stop to build it seperately and we use the normal rust infra for building it.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,83 @@
...
+          --disable-optimize-llvm \
+          --enable-option-checking \
+          --enable-ninja \
+          --disable-docs \
+          --enable-ccache
+    - name: Cache build artifacts
+      uses: actions/cache@v4
+      with:
+        path: |
+          rust/build
```
sgasho, line 67: hmm...I think this cache has become invalid because I replaced the Enzyme submodule....

.github/workflows/enzyme-rust.yml
```diff
@@ -60,14 +60,6 @@ jobs:
...
           --disable-docs \
           --enable-ccache
-    - name: Cache build artifacts
-      uses: actions/cache@v4
-      with:
-        path: |
-          rust/build
-        key: rust-${{ matrix.commit }}-${{ matrix.os }}
-        restore-keys: |
-          rust-${{ matrix.commit }}-
```
sgasho: I don't think we can use this cache anymore because I replaced the Enzyme submodule with the source code from this repository.

ZuseZ4: The time difference from Enzyme is negligible compared to rustc or LLVM builds, so I don't think that's an issue.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,75 @@
...
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        . ~/.cargo/env
+        rustup toolchain install nightly
+        rustup default nightly
+    - uses: actions/checkout@v4
+      name: Checkout Enzyme
+    - name: Clone and configure Rust compiler
+      run: |
+        . ~/.cargo/env
+        git clone https://github.com/rust-lang/rust.git
```
wsmoses: you should use another actions/checkout for this [and can give it the relevant tag/commit]

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,79 @@
...
+        ./configure \
+          --enable-llvm-link-shared \
+          --enable-llvm-plugins \
+          --enable-llvm-enzyme \
+          --release-channel=nightly \
+          --enable-llvm-assertions \
+          --enable-clang \
+          --enable-lld \
+          --disable-optimize-llvm \
+          --enable-option-checking \
```
wsmoses, line 70: we should use the github actions cache here, but lets see if it builds first

wsmoses: otherwise ccache doesn't do anything as the run is ephemeral

ZuseZ4:
@sgasho Not sure if it's related, or needed but you can also try to do a submodule update for enzyme and commit it.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,84 @@
...
+    steps:
+    - name: Install dependencies
+      run: |
+        apt-get update
+        apt-get install -y ninja-build cmake ccache nodejs
+    - name: Cache ccache
+      uses: actions/cache@v4
+      with:
+        path: ~/.cache/ccache
+        key: ccache-${{ matrix.os }}-${{ matrix.commit }}
```
wsmoses: maybe also include rust here?

ZuseZ4:
I blessed the last Rust test, so now rust CI should pass if rerun.

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,88 @@
...
+      image: ${{ (contains(matrix.os, 'linux') && 'ghcr.io/enzymead/reactant-docker-images@sha256:91e1edb7a7c869d5a70db06e417f22907be0e67ca86641d48adcea221fedc674' ) || '' }}
+    
+    steps:
+    - name: Install dependencies
+      run: |
+        apt-get update
+        apt-get install -y ninja-build cmake ccache nodejs
+    - name: Prepare ccache directory
+      run: |
+        mkdir -p "$CCACHE_DIR"
```
sgasho, line 37: I'm not sure if I should do this. Just in case for now, though

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,88 @@
...
+    branches:
+      - main
+  merge_group:
+
+jobs:
+  rust-autodiff:
+    name: Rust Autodiff ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    env:
+      CCACHE_DIR: "${GITHUB_WORKSPACE}/.ccache"
```
giordano: You can't use shell syntax, you're getting a literal `${GITHUB_WORKSPACE}` here: https://github.com/EnzymeAD/Enzyme/actions/runs/18382389055/job/52372812771?pr=2430#step:5:10.

I recommend adding a step at the beginning of the job which runs
```sh
echo "CCACHE_DIR=${GITHUB_WORKSPACE}/.ccache" >> "${GITHUB_ENV}"
```

giordano:
I don't think the cache is really accessible for anyone.  Running `ccache --show-stats` before and after is your best bet at seeing any information.

sgasho:
Sorry, I've been a bit busy and couldn't check comments... I made a few commits

ZuseZ4:
Not sure about that part.
@I-Al-Istannen mentioned though that it would be good to move the first show stats a little bit down, to run after the `Cache ccache` step where the cache gets restored.

sgasho:
> thread 'main' (2614) panicked at src/bootstrap/src/core/config/config.rs:1768:22:
> called `Result::unwrap()` on an `Err` value: "command did not execute successfully: cd \"/__w/Enzyme/Enzyme/rust\" && \"git\" \"rev-parse\" \"HEAD^1\"\nexpected success, got: exit status: 128\n"
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

hmm....I couldn't reproduce this on my local machine using [act](https://github.com/nektos/act). 
I explicitly set fetch-depth to 0 for now.

[specify fetch-depth](https://github.com/EnzymeAD/Enzyme/pull/2430/commits/55fb6610f621bfec87ed6ab2393322d6aa8d6381)

.github/workflows/enzyme-rust.yml
```diff
@@ -0,0 +1,91 @@
...
+        mkdir -p "$CCACHE_DIR"
+    - name: Install Rust toolchain
+      run: |
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+        . ~/.cargo/env
+        rustup toolchain install nightly
+        rustup default nightly
+    - uses: actions/checkout@v4
+      name: Checkout Enzyme
+    - uses: actions/checkout@v4
```
giordano:
```suggestion
    - uses: actions/checkout@v5
      name: Checkout Enzyme
    - uses: actions/checkout@v5
```

ZuseZ4:
It looks like the cmake files alone are not enough and people are not keen on shipping additional things just for one user, so Enzyme should just build it's own LLVM. Even with a cache that barely has any cache hits, the last run was only 23 minutes once we remove cland and lld. That's already at the lower end of what the Julia runners need, so from a Rust point I'd recommend to land that and open an issue for someone to investigate why ccache has a low hitrate, to make it even faster.

wsmoses:
What was the error message when cmake was shipped?

wsmoses:
What was the error message when cmake was shipped?

ZuseZ4:
https://github.com/rust-lang/rust/pull/148027#issuecomment-3443922402, and looking at the corresponding CMake file it would have looked for all other archives next (if I read cmake correctly).

sgasho:
@giordano 
> Could you run it again? Now we can see how much faster it gets

Thanks!


The second run took 10 minutes, which was three times faster than without the cache.

スクリーンショット 2025-10-27 23 12 10

PR #2434, opened 2025-08-31
[CI] Replace `tibdex/github-app-token` with `actions/create-github-app-token`

giordano:
[`tibdex/github-app-token`](https://github.com/tibdex/github-app-token) is now archived in favour of [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token).


wsmoses (approved):

wsmoses:
We have fun stuff for a new jll too so might as well do that later today

PR #2445, opened 2025-09-11
update handling of rust (de)alloc fn to use demangled names

ZuseZ4:
Once the other Rust CI PR lands one could add some Integration test.


wsmoses:
see comment in linked issue, can we solve this purely rust side by attaching the "enzyme behaves like" (enzyme_math) attribute?

PR #2449, opened 2025-09-16
Fwddiff batching

vimarsh6739:


enzyme/Enzyme/MLIR/Passes/EnzymeBatchPass.h
```diff
@@ -111,6 +111,29 @@ LogicalResult batchOperation(
   return success();
 }
 
+template <typename T>
+FunctionOpInterface batchOperationWithoutInsertingCallOp(
```
wsmoses: this should not be necessary

vimarsh6739: yeah but for some reason undoing the first commit(which moves the batchutils stuff to a new cpp file) kind of messed up compilation

enzyme/test/MLIR/ForwardMode/batched_scalar_merge.mlir
```diff
@@ -0,0 +1,27 @@
...
+    return %y : f64
+  }
+  func.func @dsq(%x : f64, %dx1 : f64, %dx2 : f64) -> (f64,f64) {
+    %r1 = enzyme.fwddiff @square(%x, %dx1) { activity=[#enzyme<activity enzyme_dup>], ret_activity=[#enzyme<activity enzyme_dupnoneed>]} : (f64, f64) -> (f64)
+    %r2 = enzyme.fwddiff @square(%x, %dx2) { activity=[#enzyme<activity enzyme_dup>], ret_activity=[#enzyme<activity enzyme_dupnoneed>] } : (f64, f64) -> (f64)
+    return %r1,%r2 : (f64,f64)
+  }
+}
+
+// CHECK:   func.func @dsq(%[[arg0:.+]]: f64, %[[arg1:.+]]: tensor<2xf64>) -> tensor<2xf64> {
```
wsmoses: can we have a simpler test that just does the merge and not run AD?

Also we should but this in a different subfolder maybe Optimize (instead of ForwardMode)

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -0,0 +1,439 @@
...
+#include "Passes/Passes.h.inc"
+} // namespace enzyme
+} // namespace mlir
+
+namespace {
+
+struct BatchDiffPass : public enzyme::impl::BatchDiffPassBase<BatchDiffPass> {
+  void runOnOperation() override;
+
+  bool isReadOnly(Operation *op) {
```
wsmoses: you should move this to a header file, not re-implement

enzyme/test/MLIR/ForwardMode/batched_scalar_merge.mlir
vimarsh6739: yeah I can do that - I'm looking into the `isReadOnly` check because apparently `@square` fails it

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -0,0 +1,451 @@
...
+namespace enzyme {
+#define GEN_PASS_DEF_BATCHDIFFPASS
+#include "Passes/Passes.h.inc"
+} // namespace enzyme
+} // namespace mlir
+
+namespace mlir {
+namespace enzyme {
+namespace batchutils {
+
```
wsmoses, line 38: Do we really need to copy this?

vimarsh6739: not really but I can fix it up in https://github.com/EnzymeAD/Enzyme/pull/2460 (this branch is the parent to that)

PR #2460, opened 2025-09-22
Rev mode autodiff batching

vimarsh6739:


enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,52 @@
+//===- OperationUtils.cpp - Utilities for operation interfaces
```
wsmoses: can we just call this utils?

vimarsh6739: yes, and I also think it might be good to push the alias functions here

vimarsh6739: right now, they are all statically defined in the pass itself

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
+  auto T = dyn_cast<TensorType>(argTy);
+  mlir::Value out;
+  if (!T) {
+    // use tensor.from_elements
+    out = builder.create<tensor::FromElementsOp>(loc, argList);
+  } else {
+    // use tensor.concat on dim 0
+    out = builder.create<tensor::ConcatOp>(loc, /*dim*/ 0, argList);
+  }
+  return out;
```
wsmoses, line 71: separate but you currently just need these dialects registered :
```
def BatchDiffPass : Pass<"enzyme-diff-batch">{
  let summary = "Batching enzyme.fwddiff and enzyme.autodiff calls";
  let dependentDialects = [
    "arith::ArithDialect",
    "tensor::TensorDialect",
    "enzyme::EnzymeDialect",
  ];
}
```

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
-    effectInterface.getEffects(effects);
-    if (!llvm::all_of(effects, [op](const MemoryEffects::EffectInstance &it) {
-          return isa<MemoryEffects::Read>(it.getEffect());
-        })) {
-      return false;
+Value tensorizeArg(OpBuilder &builder, Location &loc,
+                   SmallVector<Value> &argList) {
+  auto argTy = argList.front().getType();
+  auto T = dyn_cast<TensorType>(argTy);
+  mlir::Value out;
```
wsmoses: this is going to break in Enzyme-JaX.

Can you instead make a new enzyme op for these operations, and then a separate lowering pass. Here we will add a lowering pass to tensor. In Enzyme-Jax we will add a lowering pass to stablehlo

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
-  if (isRecursiveContainer) {
-    for (Region &region : op->getRegions()) {
-      for (auto &block : region) {
-        for (auto &nestedOp : block)
-          if (!isReadOnly(&nestedOp))
-            return false;
+  return out;
+}
+
+SmallVector<MemoryEffects::EffectInstance> collectFnEffects(
```
wsmoses: This should go in utils

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
+        }
+      }
+    }
+
+    effectCache[fnOp] = innerEffects;
+  }
+  return effectCache[fnOp];
+}
+
+const std::set<std::string> &getNonCapturingFunctions() {
```
wsmoses: and this

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
+      "free",           "printf",       "fprintf",       "scanf",
+      "fscanf",         "gettimeofday", "clock_gettime", "getenv",
+      "strrchr",        "strlen",       "sprintf",       "sscanf",
+      "mkdir",          "fwrite",       "fread",         "memcpy",
+      "cudaMemcpy",     "memset",       "cudaMemset",    "__isoc99_scanf",
+      "__isoc99_fscanf"};
+  return NonCapturingFunctions;
+}
+
+static bool isCaptured(Value v, Operation *potentialUser = nullptr,
```
wsmoses: and this

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
+          continue;
+      }
+      return true;
+    }
+  }
+
+  return false;
+}
+
+static Value getBase(Value v) {
```
wsmoses: and this

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
+    if (auto s = v.getDefiningOp<memref::CastOp>()) {
+      v = s.getSource();
+      continue;
+    }
+    break;
+  }
+  return v;
+}
+
+static bool isStackAlloca(Value v) {
```
wsmoses: and this

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
wsmoses: ideally this actually becomes an interface we can use (otherwise the polygeist/enzymexla.pointer2memref and related won't trigger here (and we want them to trigger to support c++ code)).

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
+  return v.getDefiningOp<memref::AllocaOp>() ||
+         v.getDefiningOp<memref::AllocOp>() ||
+         v.getDefiningOp<LLVM::AllocaOp>();
+}
+
+static bool mayAlias(Value v, Value v2) {
```
wsmoses: also to move to utils

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -36,37 +42,272 @@ namespace mlir {
...
+  Value valA = a.getValue();
+  Value valB = b.getValue();
+
+  // unknown effects may always alias
+  if (!valA || !valB) {
+    return true;
+  }
+  auto valResult = mayAlias(valA, valB);
+  // query alias analysis and polygeist based alias analysis
+  auto aliasResult = aliasAnalyzer.alias(valA, valB);
```
wsmoses: @pengmai already has a much more robust alias analysis integrated here (though perhaps doesn't understand the properties above): https://github.com/EnzymeAD/Enzyme/blob/a2af0289d50f4f505f466c10ec488b4f3d7cf77e/enzyme/Enzyme/MLIR/Analysis/DataFlowAliasAnalysis.h

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.h
```diff
@@ -48,12 +51,64 @@ struct BatchDiffCacheKey {
...
+  return key;
+}
+
+Value tensorizeArg(OpBuilder &builder, Location &loc,
+                   SmallVector<Value> &argList);
+
+Value extractArg(OpBuilder &builder, Location &loc, Type &argTy, Value &val,
+                 int64_t index);
+
+SmallVector<MemoryEffects::EffectInstance> collectFnEffects(
```
wsmoses: This should be a DenseMap, not std::map

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+        auto key = pair.first;
+        auto allDiffs = pair.second;
+        if (allDiffs.size() < 2)
+          continue;
+
+        // Collect inner effects of function
+        SmallVector<MemoryEffects::EffectInstance> calleeEffects =
+            batchutils::collectFnEffects(innerEffectCache, key.function);
+
+        // TODO skip if known readonly from existing analyses
```
wsmoses: readnone, not readonly

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+                  break;
+                }
+              }
+              auto e = eff.getEffect();
+              if (found) {
+
+                // Primal value as effect
+                Value primalVal = key.inputs[argnum];
+                if (auto primalOpResult = dyn_cast<OpResult>(primalVal))
+                  callerEffects.emplace_back(MemoryEffects::EffectInstance(
```
wsmoses: this is wrong, we need to be conservative and set this equal to the default resource

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+                      eff.getEffect(), primalOpResult, eff.getResource()));
+                else if (auto primalBlockArg =
+                             dyn_cast<BlockArgument>(primalVal)) {
```
wsmoses: this is insufficient, it needs to both be a blockargument, and the owner of the arg needs to be the entry to the fn (e.g. if its an iter_arg of an scf.for this does not apply)

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+                  callerEffects.emplace_back(MemoryEffects::EffectInstance(
+                      eff.getEffect(), primalBlockArg, eff.getResource()));
+                } else {
+                  llvm_unreachable("Value is neither an argument nor a result "
```
wsmoses: we should never unreachable, we should to the conservative default resource thing

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+                }
 
-    // process map
-    for (auto mergeItr = toMerge.begin(); mergeItr != toMerge.end();
-         ++mergeItr) {
-      auto key = mergeItr->first;
-      SmallVector<enzyme::ForwardDiffOp> allOps = mergeItr->second;
-      auto width = allOps.size();
+                // Derivative effects(remain the same for fwddiff)
+                for (auto dop : allDiffs) {
```
wsmoses: can you move this loop to precompute the mapping of getPrimalArgs() to be done once, outside of the call effects.

also frankly, we should jjust add a getPrimalArgs helper function to fwddiff/autodiff ops [and we can call that from here]

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
wsmoses: we should not add this effect, we should add `MemoryEffects::EffectInstance(
                      eff.getEffect(), primalArgs[primalBlockArg.getIdx()], eff.getResource())`

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+                // function argument
+                // TODO: Handle this either as a global value, or a value which
+                // is inside of the MLIR function.
+                callerEffects.emplace_back(eff);
+              }
 
-      if (width < 2)
-        continue;
+            } else {
+              // unknown effect which isnt a value
```
wsmoses: nested ifs are hard to read. can you rewrite as

```
Value resource = eff.getValue();
// unknown effect which isnt a value
if (!resource) {
  callerEffects.emplace_back(eff);
  continue;
}

.. rest of code without extra indent/braces

```

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+              callerEffects.emplace_back(eff);
+            }
+          } else {
+            // encountered an allocate / load effect. Skip to next map entry
```
wsmoses: you mean allocate/free?

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+          // legal to merge since there is no effect overwrite
+          legalMerge = allDiffs;
+        } else {
+
+          legalMerge.emplace_back(allDiffs[0]);
+          SmallVector<MemoryEffects::EffectInstance, 4> betweenEffects;
+
+          for (auto *cur = allDiffs[0]->getNextNode(); cur != allDiffs.back();
+               cur = cur->getNextNode()) {
+            auto curOpEffects = mlir::getEffectsRecursively(cur);
```
wsmoses: you should call into the same getEffects helper function from above for consistency, namely a vendored version of https://github.com/EnzymeAD/Enzyme-JAX/blob/67dc89667de4bdf4737b577f2bb29f32ef25c1db/src/enzyme_ad/jax/Utils.cpp#L46 (which should be called by the collectFn one)

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
-            mlir::Value batchedDeriv;
-            if (!T) {
-              // use tensor.from_elements
-              batchedDeriv =
-                  builder.create<tensor::FromElementsOp>(loc, derivToBatch);
-            } else {
-              // use tensor.concat on dim 0
-              batchedDeriv = builder.create<tensor::ConcatOp>(loc, /*dim*/ 0,
-                                                              derivToBatch);
+            bool stillOk = true;
```
wsmoses: stillOk -> legal or conflicting

wsmoses: this also relies on us to define memory effect collection for forwarrddiff/etc ops, or we need to explicitly do the check for forwarddiff ops here. 

I think we probably want to special case the forward diff ops first, and if not, fallback to the generic collected effects.

Specifically if a forward op += the primal, and we call it twice, that still conflicts. But both reading the same is fine. We already computed the aliasing set of the primals above. We also need to check the shadows too (e.g. if the same shadow is +='d into)

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+              }
+            }
+            if (!stillOk) {
+              break;
             }
 
-            in_args.push_back(batchedDeriv);
-            in_idx++;
+            auto curFnOp = dyn_cast<enzyme::ForwardDiffOp>(cur);
+            if (curFnOp && (std::find(pair.second.begin(), pair.second.end(),
```
wsmoses: llvm::is_contained(pair.second, curFnOp)

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
           }
         }
 
-        // process output type
-        auto out_idx = 0;
-        for (auto [idx, ract] : llvm::enumerate(key.retActivity)) {
-          ActivityAttr iattr = ActivityAttr::get(context, ract);
+        // go ahead and actually do the merge now
+        {
+          SmallVector<enzyme::ForwardDiffOp> allOps = legalMerge;
```
wsmoses: can we just use the name legalMerge, if not we can do `SmallVector<enzyme::ForwardDiffOp> &allOps` to avoid the copy

enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.cpp
```diff
@@ -75,377 +316,798 @@ namespace {
...
+
+              auto argnum = 0;
+              bool found = false;
+              for (auto fnArg : key.function.getArguments()) {
+                if (fnArg == resource) {
+                  argnum = fnArg.getArgNumber();
+                  found = true;
+                  break;
+                }
+              }
```
wsmoses: same general comments as above^^, wrapping things into helper fns will also cut down on duplication

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -127,8 +127,17 @@ def ForwardDiffOp : Enzyme_Op<"fwddiff",
...
 
-  // Define activity optimizations as canonicalization 
   let hasCanonicalizer = 1;
+
+  let extraClassDeclaration = [{
+  ::llvm::SmallVector<::mlir::Value> getPrimalArgs() {
+    return ::mlir::enzyme::detail::filterGradInputs<ForwardDiffOp, false>(*this);
+  }
+
+  ::llvm::SmallVector<::mlir::Value> getGradArgs() {
```
wsmoses, line 137: We should likely separate this is getshadow args [aka the dup ones] as distinct from the autodiff differential return

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,374 @@
...
+            return false;
+      }
+    }
+  }
+
+  return true;
+}
+
+std::optional<SmallVector<MemoryEffects::EffectInstance>>
+collectOpEffects(Operation *rootOp) {
```
wsmoses: is there a reason for chaing this to be out of place?

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,374 @@
...
+        for (Block &block : region) {
+          for (Operation &nestedOp : block) {
+            effectingOps.push_back(&nestedOp);
+          }
+        }
+      }
+    }
+
+    if (auto effectInterface = dyn_cast<MemoryEffectOpInterface>(op)) {
+      effectInterface.getEffects(effects);
```
wsmoses: I'm not sure this is guaranteed to append into effects [hence the old code made a temporary allocation for the local effects, then appended]

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,374 @@
...
+  }
+  return effects;
+}
+
+SmallVector<MemoryEffects::EffectInstance>
+collectFnEffects(FunctionOpInterface fnOp) {
+  SmallVector<MemoryEffects::EffectInstance> innerEffects;
+  for (auto &blk : fnOp.getBlocks()) {
+    for (auto &op : blk) {
+      auto opEffects = collectOpEffects(&op);
```
wsmoses: This is wrong, if you define std::null_opt as not having effects, it needs to append all effect types here

vimarsh6739: `std::nullopt` is that effects aren't defined for the op at all, not having no effects(that's an empty vector)

vimarsh6739: It's basically a re-implementation of https://github.com/llvm/llvm-project/blob/1127dd775426238cdc57b2581fbb91a0252e06ae/mlir/lib/Interfaces/SideEffectInterfaces.cpp#L346

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
vimarsh6739: Im not following - we will eventually special case the LLVM effects and autodiff/fwddiff effects here(which Im currently handling as a special case in `pruneMemoryEffects`

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
vimarsh6739: I'll change this

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
wsmoses: if effects aren't defined at all we need to be conservative and assume all effects

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
wsmoses: the polygeist collectEffects takea a `SmallVectorImpl<MemoryEffects::EffectInstance> &` argument and returns a bool if we have to be conservative that there was an unknown op

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
vimarsh6739: I feel that might be too conservative. For example, `arith.constant` will lead to a possible `Read/Write` effect then

wsmoses: no, arith.constant has memory effects implemented via pure

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,383 @@
+//===- OperationUtils.cpp - Utilities for operation interfaces
```
wsmoses: comment should say utils

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
wsmoses: this wasn't resolved, no?

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,383 @@
...
+          }
+          if (*callee == "strlen") {
+            for (auto &arg : cop.getArgOperandsMutable()) {
+              effects.emplace_back(MemoryEffects::Read::get(), &arg,
+                                   SideEffects::DefaultResource::get());
+            }
+          }
+        }
+      } else if (auto dop = dyn_cast<AutoDiffOp>(op)) {
+        // TODO: handle inter-procedural AutoDiffOp effects. Just skip for now.
```
wsmoses: we need to equally be conservative here with the all effects added (even if marked todo)

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
vimarsh6739: Oh I was assuming this was to add the effects conservatively

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,383 @@
...
+        // obtain any result(consistent with the definition of
+        // `mlir::getEffectsRecursively`)
+        // We need to be conservative here in case the op doesn't have the
+        // interface and assume it can have any possible effect.
+        effects.emplace_back(MemoryEffects::Effect::get<MemoryEffects::Read>());
+        effects.emplace_back(
+            MemoryEffects::Effect::get<MemoryEffects::Write>());
+        effects.emplace_back(
+            MemoryEffects::Effect::get<MemoryEffects::Allocate>());
+        effects.emplace_back(MemoryEffects::Effect::get<MemoryEffects::Free>());
```
wsmoses: we should break after this, and/or return false like in https://github.com/llvm/Polygeist/blob/77c04bb2a7a2406ca9480bcc9e729b07d2c8d077/lib/polygeist/Ops.cpp#L312

otherwise we spend unnecessary time when we've already gotten the most conservative possible result

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
vimarsh6739: This would break the collection of `betweenEffects`. Should I just move this to `batchutils` then?
https://github.com/EnzymeAD/Enzyme/blob/4f1d4deec19410cc75ca05d1f6968b058ed51955/enzyme/Enzyme/MLIR/Passes/EnzymeBatchDiffPass.h#L276

vimarsh6739: I'm currently handling `fwddiff` and `autodiff` as a special case

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
wsmoses: I mean that's a separate issue from having the function signature not be:
```
bool collectEffects(Operation *op,
                    SmallVectorImpl<MemoryEffects::EffectInstance> &effects) {
```

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
wsmoses: why would it break collection?

```
      } else if (auto currBwdOp = dyn_cast<AutoDiffOp>(curr)) {
        // curr is a previously un-encountered revdiff op
        auto fnOp = dyn_cast_or_null<FunctionOpInterface>(
            symbolTable.lookupNearestSymbolFrom(currBwdOp,
                                                currBwdOp.getFnAttr()));
        if (!fnOp)
          continue;

        if (!innerEffectCache.contains(fnOp)) {
          innerEffectCache[fnOp] = oputils::collectFnEffects(fnOp);
        }

        // map to outerEffects
        betweenEffects.append(batchutils::findCallerEffects(
            currBwdOp, fnOp, innerEffectCache[fnOp]));
```

this already is special cased and doesn't call this fn.

However its definitely necessary in the case fnOp has an internal autodiff call which currently is otherwise incorrectly assumed effect-free.

vimarsh6739: Ah yes then this makes sense (I'll conservatively add all effects).

vimarsh6739:
@wsmoses can you approve this

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,400 @@
...
+
+  return true;
+}
+
+bool collectOpEffects(Operation *rootOp,
+                      SmallVector<MemoryEffects::EffectInstance> &effects) {
+  SmallVector<Operation *> effectingOps(1, rootOp);
+  bool couldCollectEffects = true;
+
+  SmallVector<MemoryEffects::EffectInstance> conservativeSet;
```
wsmoses: minor but can you not make this an array here, and instead just append them as needed at bottom

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,400 @@
...
+          }
+          if (*callee == "strlen") {
+            for (auto &arg : cop.getArgOperandsMutable()) {
+              effects.emplace_back(MemoryEffects::Read::get(), &arg,
+                                   SideEffects::DefaultResource::get());
+            }
+          }
+        }
+      } else if (auto dop = dyn_cast<AutoDiffOp>(op)) {
+        // TODO: handle inter-procedural AutoDiffOp effects. Just
```
wsmoses: don't special case the autodiff ops for now, we can come back to and split

enzyme/Enzyme/MLIR/Interfaces/Utils.cpp
```diff
@@ -0,0 +1,400 @@
...
+}
+
+SmallVector<MemoryEffects::EffectInstance>
+collectFnEffects(FunctionOpInterface fnOp) {
+  SmallVector<MemoryEffects::EffectInstance> innerEffects;
+  for (auto &blk : fnOp.getBlocks()) {
+    for (auto &op : blk) {
+      SmallVector<MemoryEffects::EffectInstance> opEffects;
+      [[maybe_unused]] bool couldCollectEffects =
+          collectOpEffects(&op, opEffects);
```
wsmoses: instead have the line after as
(void)couldCollectEffects

PR #2484, opened 2025-10-07
add external_context to TypeAnalysis

vchuravy:


enzyme/Enzyme/CApi.cpp
```diff
@@ -274,7 +274,8 @@ EnzymeTypeAnalysisRef CreateTypeAnalysis(EnzymeLogicRef Log,
                                          char **customRuleNames,
                                          CustomRuleType *customRules,
                                          size_t numRules) {
-  TypeAnalysis *TA = new TypeAnalysis(((EnzymeLogic *)Log)->PPC.FAM);
+  EnzymeLogic &logic = eunwrap(Log);
+  TypeAnalysis *TA = new TypeAnalysis(logic.PPC.FAM, logic.ExternalContext);
```
wsmoses: why not just give TA the EnzymeLogic reference?

enzyme/Enzyme/CApi.cpp
```diff
@@ -314,6 +315,10 @@ void FreeTypeAnalysis(EnzymeTypeAnalysisRef TAR) {
   delete TA;
 }
 
+void *EnzymeGetExternalContextTypeAnalysis(EnzymeTypeAnalysisRef TAR) {
```
wsmoses: then this would be get EnzymeLogic

vchuravy:
@wsmoses any ideas? Otherwise I could swap back to the variant where we just store the external context.

wsmoses:
Sorry yeah I had started looking into and fixing this (it is doable), but didn't manage to complete before going on conf travel, will try to close the loop on shortly

PR #2486, opened 2025-10-08
ProbProg: Static HMC

sbrantq:


enzyme/Enzyme/MLIR/Dialect/Dialect.cpp
sbrantq: To allow passes transform instructions that return `!enzyme.Trace`. Should I make this more restrictive?

enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -367,6 +419,34 @@ def IgnoreDerivativesOp : Enzyme_Op<"ignore_derivatives",
   }];
 }
 
+def ConcatOp : Enzyme_Op<"concat"> {
```
wsmoses, line 422: isnt this already available (cc @vimarsh6739 )

wsmoses (approved):
lgtm modulo op being added again?

PR #2497, opened 2025-10-21
Lower enzyme.concat and enzyme.extract

vimarsh6739:


enzyme/Enzyme/MLIR/Passes/Passes.td
```diff
@@ -227,6 +227,15 @@ def AddToOpToSplitPass : Pass<"add-to-op-to-split"> {
   let summary = "Duplicate linalg.generic once for each output to remove enzyme.addTo op";
 }
 
+def LegalizeBatchingPass : Pass<"enzyme-legalize-batch"> {
```
wsmoses: can you call this enzyme-batch-to-tensor or something? legalize-batch isn't instructive

vimarsh6739: sg

PR #2495, opened 2025-10-17
Add shadow filters to EnzymeOps

vimarsh6739:

PR #2498, opened 2025-10-21
[mlir] fixes with region inlining/outlining

pengmai:
The initial region outlining had a calling convention bug when mixing free variables and seeds, then #2491 introduced a calling convention bug when mixing free variables with duplicated arguments. This PR should fix all of the above.

It also fixes a symbol re-definition issue when outlining.

cc @vimarsh6739

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
@@ -58,8 +58,10 @@ static void serializeFunctionAttributes(Operation *fn,
   SmallVector<NamedAttribute> fnAttrs;
   fnAttrs.reserve(fn->getAttrDictionary().size());
   for (auto attr : fn->getAttrs()) {
-    // Don't store the function type because it may change when outlining
-    if (attr.getName() == getFunctionTypeAttrName(fn))
+    // Don't store the function type or sym_name because they may change when
+    // outlining
+    if (attr.getName() == getFunctionTypeAttrName(fn) ||
+        attr.getName() == SymbolTable::getSymbolAttrName())
       continue;
     fnAttrs.push_back(attr);
   }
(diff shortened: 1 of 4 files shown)


enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
```diff
@@ -43,6 +43,16 @@ static StringRef getFunctionTypeAttrName(Operation *operation) {
       });
 }
 
+static StringRef getSymNameAttrName(Operation *operation) {
+  return llvm::TypeSwitch<Operation *, StringRef>(operation)
```
wsmoses: Can this not be determined from either function interface or symboluseropinterface?

enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
```diff
@@ -180,28 +192,40 @@ outlineAutoDiffFunc(enzyme::AutoDiffRegionOp op, StringRef funcName,
...
+      op.getRetActivity().getAsRange<enzyme::ActivityAttr>(), [](auto attr) {
+        switch (attr.getValue()) {
+        case enzyme::Activity::enzyme_active:
+        case enzyme::Activity::enzyme_activenoneed:
+          return true;
+        default:
+          return false;
+        }
+      });
+  size_t numPrimalArgs = op.getInputs().size() - numSeeds;
```
wsmoses: Hm I don't fully understand this. Why is get activity size not equal to the number of primal args,? That should always be the case, right?

pengmai: I agree the name is confusing. `numPrimalArgs` and `primalArgs` actually refers to the number of primal arguments and any duplicated shadow args. So for the added test, we have `enzyme.autodiff_region(%x, %dx, %seed)`. Crucially, we want to insert any free variables *after* `%dx` and *before* `%seed`.

I considered renaming to `nonSeeds`, but I'm not sure.

wsmoses: Okay maybe numPrimalsAndShadows

To distinguish from seeds

enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
pengmai: I thought it couldn't, but I checked again and found the function I needed under `SymbolTable::getSymbolAttrName`

enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
```diff
@@ -180,30 +182,42 @@ outlineAutoDiffFunc(enzyme::AutoDiffRegionOp op, StringRef funcName,
...
 
   SmallVector<enzyme::Activity> argActivities =
       llvm::map_to_vector(op.getActivity().getAsRange<enzyme::ActivityAttr>(),
                           [](auto attr) { return attr.getValue(); });
-
-  size_t numPrimalArgs = op.getActivity().size();
-  SmallVector<Value> primalInputs(op.getInputs().begin(),
-                                  op.getInputs().begin() + numPrimalArgs);
-  SmallVector<Value> seedInputs(op.getInputs().begin() + numPrimalArgs,
+  size_t numSeeds = llvm::count_if(
```
vimarsh6739, line 193: I think this can be avoided once #2495 is merged. We could just call `op.getDifferentialReturns()` to get seedInputs

PR #2505, opened 2025-10-28
Fix sparse add check

wsmoses:


wsmoses:
Tangentially, @nkoukpaizan would you be interested in adding grid kit as an integration test to Enzyme proper ci?

PR #2512, opened 2025-10-30
mlir: cache shadow for llvm load/store

Pangoraw:
Similar change to #2508

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
@@ -139,7 +139,7 @@ struct LoadOpInterfaceReverse
     auto loadOp = cast<LLVM::LoadOp>(op);
     auto addr = loadOp.getAddr();
     if (!(isa<AutoDiffTypeInterface>(loadOp.getType()) &&
-        (!gutils->isConstantValue(loadOp) && !gutils->isConstantValue(addr))))
+          (!gutils->isConstantValue(loadOp) && !gutils->isConstantValue(addr))))
       return {};
     OpBuilder cacheBuilder(gutils->getNewFromOriginal(op));
     return {gutils->initAndPushCache(gutils->invertPointerM(addr, cacheBuilder),
@@ -167,7 +167,6 @@ struct StoreOpInterfaceReverse
 
     if (!gutils->isConstantValue(addr)) {
       Value addrGradient = gutils->popCache(caches.front(), builder);
-
 
       if (!iface.isMutable()) {
         if (!gutils->isConstantValue(val)) {


enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
```diff
@@ -73,6 +73,34 @@ class PointerTypeInterface
...
+
+  void createShadowValues(Operation *op, OpBuilder &builder,
+                          MGradientUtilsReverse *gutils) const {
+    auto gep = cast<LLVM::GEPOp>(op);
+    auto newGep = cast<LLVM::GEPOp>(gutils->getNewFromOriginal(op));
+    auto base = gep.getBase();
+    if (!gutils->isConstantValue(base)) {
+      auto baseShadow = gutils->invertPointerM(base, builder);
+      Operation *shadowGep = builder.clone(*newGep);
+      shadowGep->setOperand(0, baseShadow);
```
pengmai: Consider using the higher level API for `LLVM::GEPOp` rather than the raw `Operation *` API (in the unlikely event that the GEP operand order ever changes so 0 is no longer the base pointer)

```suggestion
      auto shadowGep = cast<LLVM::GEPOp>(builder.clone(*newGep));
      shadowGep.getBaseMutable().assign(baseShadow);
```

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
```diff
@@ -153,7 +185,11 @@ struct StoreOpInterfaceReverse
 
   SmallVector<Value> cacheValues(Operation *op,
                                  MGradientUtilsReverse *gutils) const {
-    return SmallVector<Value>();
+    auto storeOp = cast<LLVM::StoreOp>(op);
+    auto addr = storeOp.getAddr();
```
wsmoses, line 192: if the pop is gated by `if (!gutils->isConstantValue(addr))`, the push also needs to have the same condition

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
Pangoraw: good point thanks

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
```diff
@@ -108,7 +136,11 @@ struct LoadOpInterfaceReverse
 
   SmallVector<Value> cacheValues(Operation *op,
                                  MGradientUtilsReverse *gutils) const {
-    return SmallVector<Value>();
+    auto loadOp = cast<LLVM::LoadOp>(op);
+    auto addr = loadOp.getAddr();
+    OpBuilder cacheBuilder(gutils->getNewFromOriginal(op));
+    return {gutils->initAndPushCache(gutils->invertPointerM(addr, cacheBuilder),
```
wsmoses, line 145: same push/pop condition comment here

PR #2518, opened 2025-11-01
ProbProg: MemAlloc effect and remove unused InitTrace op

sbrantq:
~Attaching a `source_fn` so CSE won't merge trace init ops (for context, fresh trace objects are required for probprog calls as they may have different mode/address chains/rng states). ProbProg lowering always clones `toeval` with a new name that automatically differs across `MProbProgMode` and sample op address chains~

~Update: Mark InitTraceOp as `MemoryEffects<[MemAlloc]>` and add remove-unnecessary-enzyme-ops pattern~

Update^2: Mark InitTraceOp as `MemoryEffects<[MemAlloc]>` and add canonicalizer pattern to remove unused ones

### DIFF
--- a/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
+++ b/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
@@ -578,6 +578,7 @@ def InitTraceOp : Enzyme_Op<"initTrace", [MemoryEffects<[MemAlloc]>]> {
   let arguments = (ins );
   let results = (outs Trace:$trace);
   let assemblyFormat = "attr-dict `:` type($trace)";
+  let hasCanonicalizer = 1;
 }
 
 def AddSampleToTraceOp : Enzyme_Op<"addSampleToTrace", [Pure]> {
--- a/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
+++ b/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
@@ -1039,3 +1039,27 @@ LogicalResult MCMCOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
 
   return success();
 }
+
+//===----------------------------------------------------------------------===//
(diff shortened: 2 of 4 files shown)


enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -575,7 +575,7 @@ def InitTraceOp : Enzyme_Op<"initTrace", [Pure]> {
   let description = [{
     Initialize an execution trace for a probabilistic function.
   }];
-  let arguments = (ins );
+  let arguments = (ins OptionalAttr<FlatSymbolRefAttr>:$source_fn);
```
wsmoses: I mean the bigger problem/reason for this is that you marked initTrace as pure?

Why not remove the purity attribute, and only mark it as alloc like (and not read/write/free)

sbrantq: Yes that's absolutely what I should do! Can I manually DCE them in `remove-unnecessary-enzyme-ops`? Does it look ok now?

wsmoses (approved):

enzyme/Enzyme/MLIR/Passes/RemoveUnusedEnzymeOps.cpp
```diff
@@ -319,6 +319,20 @@ struct InitSimplify : public OpRewritePattern<enzyme::InitOp> {
   }
 };
 
+struct InitTraceSimplify : public OpRewritePattern<enzyme::InitTraceOp> {
```
wsmoses: I would actually add this as a canonicalization for InitTraceOp itself, rather than only within removeunusedenzymeops

sbrantq: Completely agree & done

PR #2519, opened 2025-11-02
ProbProg: NUTS

sbrantq:
Looks like changes needed to generalize`enzyme.MCMC` op to have serious stuff (like sampling loops and markov chains) would be more than 1k lines of code; i will open another PR

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/ProbProgMLIRPass.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/ProbProgMLIRPass.cpp
@@ -655,7 +655,8 @@ struct ProbProgPass : public enzyme::impl::ProbProgPassBase<ProbProgPass> {
           TypeRange{rngInput.getType(), rngInput.getType(), rngInput.getType()},
           initSplit.getResult(0));
       auto rngTree = kernelSplit.getResult(0);
-      auto rngMomentumInit = kernelSplit.getResult(2);
+      // auto rngWarmup = kernelSplit.getResult(1); TODO (ascertain)
+      // auto rngMomentumInit = kernelSplit.getResult(2); TODO (diagnostics)
 
       // Extract initial position vector q0
       int64_t positionSize =


enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -244,15 +244,15 @@ def AutoDiffRegionOp : Enzyme_Op<"autodiff_region", [AutomaticAllocationScope]>
...
+    ParentOneOf<["AutoDiffRegionOp", "ForLoopOp"]>]> {
+  let summary = "Yield values at the end of an autodiff_region or loop ops";
   let arguments = (ins Variadic<AnyType>:$operands);
   let assemblyFormat = [{
     attr-dict ($operands^ `:` type($operands))?
   }];
 }
 
-def LoopOp : Enzyme_Op<"loop", [AutomaticAllocationScope]> {
+def ForLoopOp : Enzyme_Op<"for_loop", [AutomaticAllocationScope]> {
```
wsmoses, line 239: is there a reason we need this in contrast to say an scf.for/affine.for ?

wsmoses: if we lower to the loop from a higher level op, then yeah I get the reasoning, since we want to then convert to scf vs stablehlo

wsmoses: if this comes as an input, then we should just reuse the existing for ops

sbrantq: Yes it comes from lowering a high level `enzyme.mcmc` op and is to be lowered to `stablehlo.while`. In the loop we will have to pass in/yield custom types like `!enzyme.Trace` and hypothetically something like `!enzyme.TreeState` for WIP NUTS

wsmoses: Got it!

PR #2523, opened 2025-11-03
[mlir] Handle dynamic inner loop dimension in cache to memref

Pangoraw:


enzyme/test/MLIR/ReverseMode/scf_for_memref_inner_dynamic.mlir
```diff
@@ -0,0 +1,63 @@
+// RUN: %eopt %s --enzyme-wrap="infn=reduce outfn= argTys=enzyme_active,enzyme_const retTys=enzyme_active mode=ReverseModeCombined" --canonicalize --remove-unnecessary-enzyme-ops --canonicalize --enzyme-simplify-math --canonicalize | FileCheck %s
+
+func.func @reduce(%x: f32, %ub: index) -> (f32) {
+  %lb = arith.constant 0 : index
+  %step = arith.constant 1 : index
+
+  // Initial sum set to 0.
+  %sum_0 = arith.constant 1.0 : f32
+  // iter_args binds initial values to the loop's region arguments.
+  %sum2 = scf.for %iv2 = %lb to %ub step %step
```
wsmoses: just to confirm the test is working as intended, can we use different lb/ub for the two loops (and that way we can check the right index is going to the right place)

PR #2524, opened 2025-11-03
Forward-Mode MPI_Recv

vchuravy:
Turns out that we "forgot" to emit the shadow operations in forward mode.

Is there a way to update the lit tests?


wsmoses (approved):
got catch lmao

PR #2527, opened 2025-11-04
LLVM optimizes the alloca for the mpi_wrapper to undef

vchuravy:
Took me a while to find the culprit, but looking at 

```julia
using Enzyme, MPI
MPI.Init()

function f(x, comm)
   @show MPI.Comm_size(comm)
   return x^2
end
Enzyme.Compiler.enzyme_code_llvm(f, Duplicated, Tuple{Duplicated{Float64}, Const{MPI.Comm}}, mode=Enzyme.API.DEM_ForwardMode, dump_module=true)
```

Showed that:
```
; Function Attrs: alwaysinline mustprogress nofree nosync nounwind speculatable willreturn memory(none)
define private i32 @"enzyme_wrapmpi$$ejlptr$MPI_Comm_size$1#"(i32 %0) #7 {
entry:
  %1 = alloca i32, align 4
  %2 = ptrtoint i32* %1 to i64
  %3 = call i32 @"ejlptr$MPI_Comm_size$1"(i32 %0, i64 noundef %2) #17
  ret i32 undef
}
```

Compare this to the Compiler-Explorer
https://godbolt.org/z/MErThjvz5

The culprit seems to be that `memory(argmem: readwrite, inaccessiblemem: readwrite)` and the pointers as integer ABI from Julia conflict,
and allow LLVM to prove that the alloca is never initialized, thus replacing the load with an undef.

I am worried that this assumption is in more places. @wsmoses could you take a look?

This does indeed fix the 1.11 issue on github.com/EnzymeAD/Enzyme.jl/pull/518 where MPI_Comm_size returns 0

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -186,67 +186,98 @@ bool attributeKnownFunctions(llvm::Function &F) {
     F.addFnAttr(Attribute::NoFree);
   }
   if (F.getName() == "MPI_Irecv" || F.getName() == "PMPI_Irecv") {
+    auto FT = F.getFunctionType();
+    bool PointerABI = true;
     changed = true;
-#if LLVM_VERSION_MAJOR >= 16
-    F.setOnlyAccessesInaccessibleMemOrArgMem();
-#else
-    F.addFnAttr(Attribute::InaccessibleMemOrArgMemOnly);
-#endif
     F.addFnAttr(Attribute::NoUnwind);
     F.addFnAttr(Attribute::NoRecurse);
     F.addFnAttr(Attribute::WillReturn);
     F.addFnAttr(Attribute::NoFree);
     F.addFnAttr(Attribute::NoSync);
-    F.addParamAttr(0, Attribute::WriteOnly);
-    if (F.getFunctionType()->getParamType(2)->isPointerTy()) {
+    if (FT->getParamType(0)->isPointerTy()) {
+      F.addParamAttr(0, Attribute::WriteOnly);
+    } else {
+      PointerABI = false;
+    }
+    // OpenMPI vs MPICH
+    if (FT->getParamType(2)->isPointerTy()) {
       addFunctionNoCapture(&F, 2);
       F.addParamAttr(2, Attribute::WriteOnly);
     }
-    if (F.getFunctionType()->getParamType(6)->isPointerTy()) {
+    if (FT->getParamType(6)->isPointerTy()) {
       F.addParamAttr(6, Attribute::WriteOnly);
+    } else {
+      PointerABI = false;
+    }
+    if (PointerABI) {
+#if LLVM_VERSION_MAJOR >= 16
+      F.setOnlyAccessesInaccessibleMemOrArgMem();
+#else
+      F.addFnAttr(Attribute::InaccessibleMemOrArgMemOnly);
+#endif
     }
   }
   auto name = getFuncName(&F);
   if (name == "MPI_Isend" || name == "PMPI_Isend") {
+    auto FT = F.getFunctionType();
+    bool PointerABI = true;
     changed = true;
-#if LLVM_VERSION_MAJOR >= 16
-    F.setOnlyAccessesInaccessibleMemOrArgMem();
-#else
-    F.addFnAttr(Attribute::InaccessibleMemOrArgMemOnly);
-#endif
     F.addFnAttr(Attribute::NoUnwind);
     F.addFnAttr(Attribute::NoRecurse);
     F.addFnAttr(Attribute::WillReturn);
     F.addFnAttr(Attribute::NoFree);
     F.addFnAttr(Attribute::NoSync);
-    F.addParamAttr(0, Attribute::ReadOnly);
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -186,65 +186,97 @@ bool attributeKnownFunctions(llvm::Function &F) {
...
+    }
     F.addFnAttr(Attribute::NoUnwind);
     F.addFnAttr(Attribute::NoRecurse);
     F.addFnAttr(Attribute::WillReturn);
     F.addFnAttr(Attribute::NoFree);
     F.addFnAttr(Attribute::NoSync);
-    F.addParamAttr(0, Attribute::WriteOnly);
-    if (F.getFunctionType()->getParamType(2)->isPointerTy()) {
+    if (PointerABI) {
+      F.addParamAttr(0, Attribute::WriteOnly);
```
wsmoses, line 198: this shouldn't be necessary....

vchuravy:
> frankly we should just force the MPI wrapper to always be a pointer, and do the inttoptr/etc inside the wrapper

It's not just the wrapper though. Currently we optimistically annotate these functions as `setOnlyAccessesInaccessibleMemOrArgMem` which will happily miscompile the `ptrtoint` code anywhere.

enzyme/Enzyme/Enzyme.cpp
vchuravy: How do you mean? Can I apply the WriteOnly argument to  a `int` argument? 

This is https://www.mpich.org/static/docs/v4.1/www3/MPI_Irecv.html

and Julia ccall ABI will call this as `MPI_Irecv(Int64, Int32, ...)`

wsmoses: Ah sorry misread the code

wsmoses:
> > A larger lift, but because this is actually required for MPI perf im potentially inclined to force enzyme side rewriting into pointer abi
> 
> This is made more complicated by the MPI ABI itself (sometimes MPI_Comm is a pointer and sometimes an integer and that is not Julia missing things up)

I mean this only really matters for things we would mark as read/writeonly/nocapture, so really the request/status/input/output

enzyme/Enzyme/Enzyme.cpp
wsmoses: ah sorry I meant to apply this to the change of 
```
      F.setOnlyAccessesInaccessibleMemOrArgMem();
```

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -186,65 +186,97 @@ bool attributeKnownFunctions(llvm::Function &F) {
     F.addFnAttr(Attribute::NoFree);
   }
   if (F.getName() == "MPI_Irecv" || F.getName() == "PMPI_Irecv") {
+    auto FT = F.getFunctionType();
+    bool PointerABI = FT->getParamType(0)->isPointerTy();
+    if (!PointerABI) {
```
wsmoses: we should still do this check per arg, since we could get mixed abis

vchuravy: That's why I have the assert, for the arguments that are coming from the Julia ABI we should enforce consistency.

wsmoses: more that julia exists that I'd be prone to something like
```
bool PointerABI = true;
if (F->getParamType(0)->isPointerType()) {
   ....
} else {
   PointerABI = false;
}
...
if (F->getParamType(i)->isPointerType()) {
   ....
} else {
   PointerABI = false;
}

if (PointerABI) {
#if LLVM_VERSION_MAJOR >= 16
      F.setOnlyAccessesInaccessibleMemOrArgMem();
#else
      F.addFnAttr(Attribute::InaccessibleMemOrArgMemOnly);
#endif
}

```

vchuravy: So to be clear:

```
int MPI_Irecv(void *buf, int count, MPI_Datatype datatype, int source, int tag,
              MPI_Comm comm, MPI_Request *request)
```

Julia passes `void * buf` and `MPIRequest *request` as `int64_t` and MPICH passes `datatype` as `int32_t` and OpenMPI passes it as `ptr`.

As far as I understand we cannot use `setOnlyAccessesInaccessibleMemOrArgMem` if we encounter the Julia ABI, since that opens us up for miscompiles.

> more that julia exists

Citation needed xD

IMO I would not allow mixed ABI either you are Julia and you stupidly used opaque pointers before they were a real thing, or you are actually are passing pointers in the right place.

I have also been debating adjusting MPI.jl to just use `Core.LLVMPtr`, but it feels silly to do that when 1.12 will actually switch us to the right ABI.

wsmoses: I can see the argument, but I can also see some weird ABI dlopen shenanigans getting optimized into ptrtoint, so if it's not a large maintenance burden I would tend to keep support for?

For one thing it also means that we can also potentially emit d/dx mpi_send -> mpi_recv calls with the mixed abi (which is possible is happening already, for blas at least I think may be the case currently).

PR #2533, opened 2025-11-06
Check for both MPI and PMPI versions

vchuravy:
@wsmoses on 1.10 I am seeing:

```
; Function Attrs: nofree norecurse nosync nounwind willreturn
declare i32 @PMPI_Wait(i64, i64) local_unnamed_addr #13

; Function Attrs: nofree norecurse nosync nounwind willreturn
declare i32 @MPI_Comm_rank(i32, i64) local_unnamed_addr #16

; Function Attrs: nofree norecurse nosync nounwind willreturn
declare i32 @MPI_Comm_size(i32, i64) local_unnamed_addr #17

; Function Attrs: nofree norecurse nosync nounwind willreturn
declare i32 @MPI_Irecv(i64, i32, i32, i32, i32, i32, i64) local_unnamed_addr #18

; Function Attrs: nofree norecurse nosync nounwind willreturn
declare i32 @MPI_Isend(i64, i32, i32, i32, i32, i32, i64) local_unnamed_addr #1
```

So when `PMPI_Wait` goes looking for `MPI_Isend` it fails to look at the right one.

If this looks kosher to you, I can also go and fix all the other users for getRenamedPerCallingConv

### DIFF
--- a/enzyme/Enzyme/Utils.h
+++ b/enzyme/Enzyme/Utils.h
@@ -2377,15 +2377,26 @@ static inline std::string getRenamedPerCallingConv(llvm::StringRef caller,
                                                    llvm::StringRef callee) {
   if (startsWith(caller, "ejl")) {
     auto &&[prefix, name, postfix] = tripleSplitDollar(caller);
-    return (prefix + "$" + getRenamedPerCallingConv(name, callee) + "$" +
-            postfix)
-        .str();
-  }
-  if (startsWith(caller, "PMPI_")) {
-    assert(startsWith(callee, "MPI"));
-    return ("P" + callee).str();
+    return (prefix + "$" + callee + "$" + postfix).str();
   }
   return callee.str();
+}
+
+static inline std::string getMPIFunction(llvm::Module &M,
+                                         llvm::StringRef caller,
+                                         llvm::StringRef callee) {
+  // The function could exist in two forms in the module - with PMPI_ or MPI_
+  // prefix. Check for both.
+  assert(startsWith(callee, "MPI"));
+  auto mpi_name = getRenamedPerCallingConv(caller, callee);
+  if (M.getFunction(mpi_name))
+    return mpi_name;
+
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/Utils.h
```diff
@@ -2377,15 +2377,26 @@ static inline std::string getRenamedPerCallingConv(llvm::StringRef caller,
...
   }
   return callee.str();
 }
 
+static inline std::string getMPIFunction(llvm::Module &M,
+                                         llvm::StringRef caller,
+                                         llvm::StringRef callee) {
+  // The function could exist in two forms in the module - with PMPI_ or MPI_
+  // prefix. Check for both.
+  assert(startsWith(callee, "MPI"));
```
wsmoses, line 2390: no this is wrong, we should create the func decl if not exists

vchuravy: This only checks that the user side is passing in `MPI_Gather`. (Used to be in getRenamedPerCallingConv)

wsmoses: sorry comment applied to three lines later

wsmoses:
the issue imo is that this https://github.com/EnzymeAD/Enzyme/blob/6b1848d8582e57dd57c0bb5d0c373c5cb1c1bbfb/enzyme/Enzyme/Utils.cpp#L1803 needs to become a getOrInsertFunction, not getFunction

wsmoses:
we have the args, so therefore we ought be able to form the functiontype

vchuravy:
> since I assume julia never gives us a mix. So we should never generate a mix?

Enzyme.jl is to blame, it tries to invert a pointer from Julia and what name it ends up is a 50/50. 


> since I assume julia never gives us a mix. So we should never generate a mix?

See the module, Julia precisely gives us a mix and that is the issue I am trying to fix.

wsmoses:
ah bleh

PR #2535, opened 2025-11-06
Support persistent MPI

vchuravy:
Adding support for `MPI_Send_init`, `MPI_Recv_init` and `MPI_Start`

### DIFF
--- a/enzyme/Enzyme/CallDerivatives.cpp
+++ b/enzyme/Enzyme/CallDerivatives.cpp
@@ -44,8 +44,8 @@ void AdjointGenerator::handleMPI(llvm::CallInst &call, llvm::Function *called,
   BuilderZ.setFastMathFlags(getFast());
 
   // MPI send / recv can only send float/integers
-  if (funcName == "PMPI_Isend" || funcName == "MPI_Isend" ||
-      funcName == "PMPI_Irecv" || funcName == "MPI_Irecv") {
+  if (funcName == "PMPI_Isend" || funcName == "MPI_Isend" || funcName == "PMPI_Send_init" || funcName == "MPI_Send_init" ||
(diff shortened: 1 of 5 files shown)


enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -2178,6 +2189,23 @@ void AdjointGenerator::handleMPI(llvm::CallInst &call, llvm::Function *called,
     return;
   }
 
+  // Adjoint of MPI_Send is to place a MPI_send at the corresponding
+  // location in the reverse.
+  if (func == "MPI_Send" || func == "PMPI_Send") {
+    if (Mode == DerivativeMode::ReverseModeGradient ||
```
wsmoses, line 2195: did we miss this from earlier??

if so we should add a test

enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -2178,6 +2189,23 @@ void AdjointGenerator::handleMPI(llvm::CallInst &call, llvm::Function *called,
...
+        Mode == DerivativeMode::ReverseModeCombined) {
+      IRBuilder<> Builder2(&call);
+      getReverseBuilder(Builder2);
+      auto callval = call.getCalledOperand();
+      Value *args[] = {
+          lookup(gutils->getNewFromOriginal(call.getOperand(0)), Builder2)};
+      Builder2.CreateCall(call.getFunctionType(), callval, args);
```
wsmoses, line 2202: doesn't cover fwd mode, so we shouldn't return in that case

enzyme/Enzyme/CallDerivatives.cpp
vchuravy: Sorry this was a brain fart on my side this ought to be `MPI_Start` not MPI_Send

PR #2553, opened 2025-11-13
Support correct free type in opaque pointer mode

wsmoses:


vchuravy (approved):
Can confirm that this locally fixes https://github.com/EnzymeAD/Enzyme.jl/issues/2770

PR #2559, opened 2025-11-15
Bugfix: Region Outlining Primal Args

vimarsh6739:
We shouldn't re-add primal inputs as const args

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
@@ -126,12 +126,16 @@ outlineAutoDiffFunc(enzyme::AutoDiffRegionOp op, StringRef funcName,
 
   llvm::SmallVector<Value> primalValuesAbove = op.getPrimalInputs();
   llvm::SmallVector<Value> blockArgs(autodiffRegion.getArguments());
-  for (Value value : freeValues) {
+  for (Value value : llvm::make_early_inc_range(freeValues)) {
     bool isPrimal = false;
     for (auto [pval, bval] : llvm::zip(primalValuesAbove, blockArgs)) {
       if (value == pval) {
         isPrimal = true;
-        value.replaceAllUsesWith(bval);
+        for (OpOperand &use : llvm::make_early_inc_range(value.getUses())) {
+          if (op->isProperAncestor(use.getOwner()))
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
```diff
@@ -124,11 +124,23 @@ outlineAutoDiffFunc(enzyme::AutoDiffRegionOp op, StringRef funcName,
   llvm::SetVector<Value> freeValues;
   getUsedValuesDefinedAbove(autodiffRegion, freeValues);
 
+  llvm::SmallVector<Value> primalValuesAbove = op.getPrimalInputs();
+  llvm::SmallVector<Value> blockArgs(autodiffRegion.getArguments());
   for (Value value : freeValues) {
```
pengmai:
```suggestion
  for (Value value : llvm::make_early_inc_range(freeValues)) {
```

enzyme/Enzyme/MLIR/Passes/InlineEnzymeRegions.cpp
```diff
@@ -124,11 +124,23 @@ outlineAutoDiffFunc(enzyme::AutoDiffRegionOp op, StringRef funcName,
...
-    inputs.push_back(value);
-    argTypes.push_back(value.getType());
-    argLocs.push_back(value.getLoc());
-    argActivities.push_back(enzyme::Activity::enzyme_const);
+    bool isPrimal = false;
+    for (auto [pval, bval] : llvm::zip(primalValuesAbove, blockArgs)) {
+      if (value == pval) {
+        isPrimal = true;
+        value.replaceAllUsesWith(bval);
```
pengmai: This leads to dominance issues if there is a use of the primal input that is outside of the `autodiff_region` op, so we need to make sure that the uses we're replacing are solely inside of the region. Furthermore, by replacing these uses, it's no longer a free value, so it needs to be removed from the list.

```suggestion
        for (OpOperand &use : llvm::make_early_inc_range(value.getUses())) {
          if (op->isProperAncestor(use.getOwner()))
            use.assign(bval);
        }
        freeValues.remove(value);
```

pengmai (commented):
It looks like the new test is failing (both in CI and I've verified it locally). The suggestions should fix them

PR #2562, opened 2025-11-18
primal code motion

vimarsh6739:
LICM-like primal code motion for both reverse and forward regions

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/HoistEnzymeRegions.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/HoistEnzymeRegions.cpp
@@ -41,10 +41,10 @@ namespace enzyme {
 
 namespace {
 
-static bool checkRangeDominance(IRMapping &btop, DominanceInfo &dom,
-                                AutoDiffRegionOp &rootOp,
-                                SetVector<Operation *> &specialOps,
-                                ValueRange values) {
+template <typename SourceOp>
+static bool
+checkRangeDominance(IRMapping &btop, DominanceInfo &dom, SourceOp &rootOp,
+                    SetVector<Operation *> &specialOps, ValueRange values) {
   SmallVector<Value> blockArgs(rootOp.getBody().getArguments());
   for (auto value : values) {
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/MLIR/Passes/HoistEnzymeRegions.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+#define DEBUG_TYPE "enzyme-hoist"
+#define ENZYME_DBGS llvm::dbgs() << "[" << DEBUG_TYPE << "]"
+
+namespace {
+
+static bool checkRangeDominance(enzyme::AutoDiffRegionOp &rootOp,
+                                SetVector<Operation *> &specialOps,
+                                ValueRange values) {
+  DominanceInfo dominance(rootOp);
```
wsmoses: can you move this outside

enzyme/Enzyme/MLIR/Passes/HoistEnzymeRegions.cpp
```diff
@@ -0,0 +1,156 @@
...
+      for (Operation &bodyOp : blk.without_terminator()) {
+        bool canLift = true;
+        llvm::SmallVector<MemoryEffects::EffectInstance> bodyOpEffects;
+        bool couldCollectEffects =
+            enzyme::oputils::collectOpEffects(&bodyOp, bodyOpEffects);
+
+        if (!couldCollectEffects)
+          canLift = false;
+
+        canLift = checkRangeDominance(rootOp, liftOps, bodyOp.getOperands());
```
wsmoses: you also need to check the op is guaranteed to execute (e.g. if multiple blocks in the region, it must always be executed, not within an if statement).

ops at the top level of the entry block have this property for sure.

wsmoses: one potential resolution, the block must postdominate (or be) the entry block to the autodiff region

PR #2564, opened 2025-11-19
mlir: single memref cache allocation for dynamic sizes

Pangoraw:
cc @pengmai if you want to give it a try with your memref linearization pass


pengmai (approved):
This looks like it'll be useful in converting the nested memrefs in the GPU benchmarks

PR #2565, opened 2025-11-19
tblgen: Prevent double broadcasting in subroutine evaluation

Pangoraw:
Fixes https://github.com/EnzymeAD/Enzyme-JAX/issues/1627.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/Common.td
+++ b/enzyme/Enzyme/MLIR/Implementations/Common.td
@@ -136,6 +136,12 @@ class ConstantFP<string val, string dialect_, string op_, string type_=""> : Ope
 }
 
 def ResultTypes : GlobalExpr</*needsprimal*/0, /*needsshadow*/0, "op->getResultTypes()">;
+def ResultShadowTypes : GlobalExpr</*needsprimal*/0, /*needsshadow*/0, [{
+  SmallVector<Type> shadowTypes(op->getResultTypes());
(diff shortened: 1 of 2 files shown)


avik-pal:
does this also happen to resolve https://github.com/EnzymeAD/Enzyme-JAX/issues/1469?

enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
```diff
@@ -240,7 +240,7 @@ bool handle(const Twine &curIndent, const Twine &argPattern, raw_ostream &os,
             const Record *pattern, const Init *resultTree, StringRef builder,
             VariableSetting &nameToOrdinal, bool lookup,
             ArrayRef<unsigned> retidx, StringRef origName, bool newFromOriginal,
-            ActionType intrinsic);
+            ActionType intrinsic, bool broadcastInputs = true);
```
wsmoses: so in princple we should already have a handler for this sort of thing (trying to keep track of each variable if it is already broadcast or not). Can we somehow make this work with that rather than adding an extra variable to pipe?

Pangoraw: Makes sense, I pinpointed the more precise cause (broadcasted inputs not marked as being vector valued)

wsmoses (approved):

PR #2600, opened 2025-12-01
Move benchmarks to separate repo, take N

wsmoses:


wsmoses:
@giordano @avik-pal in a follow up, may ask for your help setting up the nice UI of perf results from the json here

PR #2607, opened 2025-12-05
Bump llvm

wsmoses:
@pengmai I was seeming some errosr on the alias/activity analysis [seeing if this repros now via ci]. can you take a look?


pengmai:
The problem looks like some new memory locations in the `llvm.memory_effects` attributes were added recently (`errnoMem`, `targetMem0`, and `targetMem1`) and the MLIR parser rejects `llvm.memory_effects` attributes that don't have all of these locations. I can push a fix that updates the appropriate tests

wsmoses:
awesome, yeah feel free to takeover this pr

PR #2619, opened 2025-12-07
Migrate to Bzlmod

vimarsh6739:
Based it on llvm's implementation ([https://github.com/llvm/llvm-project/tree/main/utils/bazel](https://github.com/llvm/llvm-project/tree/main/utils/bazel))

### DIFF
--- a/enzyme/.bazelrc
+++ b/enzyme/.bazelrc
@@ -1,4 +1,4 @@
-common --enable_bzlmod --noenable_workspace
+common --enable_bzlmod
 build --announce_rc
 
 build --experimental_repo_remote_exec


enzyme/.bazelversion
```diff
@@ -1 +1 @@
-7.7.0
```
wsmoses, line 1: we should maintain both, and also the old bazel version right now since oss xla [and thus enzymejax/etc] is on workspace builds by default with this bazel version

enzyme/MODULE.bazel.lock
```diff
@@ -0,0 +1,318 @@
+{
+  "lockFileVersion": 13,
```
wsmoses, line 2: this doesn't seem like a file that should be committed?

enzyme/WORKSPACE
```diff
@@ -1,85 +0,0 @@
...
-hedron_compile_commands_setup_transitive()
-load("@hedron_compile_commands//:workspace_setup_transitive_transitive.bzl", "hedron_compile_commands_setup_transitive_transitive")
-hedron_compile_commands_setup_transitive_transitive()
-load("@hedron_compile_commands//:workspace_setup_transitive_transitive_transitive.bzl", "hedron_compile_commands_setup_transitive_transitive_transitive")
-hedron_compile_commands_setup_transitive_transitive_transitive()
-
-SKYLIB_VERSION = "1.3.0"
-
-http_archive(
-    name = "bazel_skylib",
```
wsmoses, line 41: we need to retain the workspace

enzyme/MODULE.bazel.lock
vimarsh6739: Apparently it's "best practice" to commit it(it is tied to bazelversion and MODULE.bazel deps).

enzyme/WORKSPACE
vimarsh6739: Is there a reason why? Enzyme's workspace/MODULE.bazel shouldn't affect downstream repos right?(alternatively Im checking if I can locally mount this branch with Enzyme-JaX)

vimarsh6739: @wsmoses this seems to work with Enzyme-JAX

PR #2640, opened 2025-12-21
Canonicalization: Remove unused primal arguments for region op

vimarsh6739:
It is safe to remove unused primal arguments for AutoDiffRegionOp when the primal has no uses within the function and is enzyme_const

### DIFF
--- a/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
+++ b/enzyme/Enzyme/MLIR/Dialect/Ops.cpp
@@ -958,6 +958,65 @@ public:
   }
 };
 
+class RemoveUnusedArgs final : public OpRewritePattern<AutoDiffRegionOp> {
+
+public:
+  using OpRewritePattern<AutoDiffRegionOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(AutoDiffRegionOp uop,
+                                PatternRewriter &rewriter) const override {
+    SmallVector<Value> newInArgs;
+    SmallVector<size_t> argIdxToErase;
+    SmallVector<ActivityAttr> newInActivityArgs;
+    llvm::SmallVector<Value> blockArg(uop.getBody().getArguments());
+    auto in_idx = 0;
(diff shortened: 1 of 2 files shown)


wsmoses:
> e assume that the only use of primals in the region are the block arguments of the entry basic block(and not the passed arguments to the op). Should we add a verifier for this? Currently, this is a manual check which is repeated for both out

No that's a limitation that should be fixed by the requisite users, as that is legal IR that ought be supported [tho perhaps not as well optimized]

enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -909,14 +909,74 @@ template <> struct ReverseRetOpt<AutoDiffRegionOp>::SourceOpCreator {
     return newOp;
   }
 };
+
+class RemoveUnusedPrimals final : public OpRewritePattern<AutoDiffRegionOp> {
```
wsmoses: Rename to RemoveUnusedArgs

enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -909,14 +909,74 @@ template <> struct ReverseRetOpt<AutoDiffRegionOp>::SourceOpCreator {
...
+
+public:
+  using OpRewritePattern<AutoDiffRegionOp>::OpRewritePattern;
```
wsmoses: I forget if we also had a forwarddiffregionop but we should do the same if so

vimarsh6739: We do, but its under review.

wsmoses:
> ove (`%xx` == {`%x`, `%dx`}) (since it is never used, we would run into issues, since `%x2` is effectively re-added always a

I don't understand this case, x2 is always const here. in fact the entire region is constant? %xx is literally unused in the region so it is always safe to remove, definitionally.

PR #2664, opened 2026-01-17
fix: memcpy -> memset uses destination alignment

### LINKED ISSUE #2663: memcpy -> memset in fwd autodiff uses source alignment rather than destination alignment
I have some LLVM IR that includes a memcpy that has a source alignment of 16, and a destination alignment of 8

```
  tail call void @llvm.memcpy.p0.p0.i64(ptr noundef nonnull align 8 dereferenceable(160) %invariant.gep1996, ptr noundef nonnull align 16 dereferenceable(160) getelementptr inbounds nuw (i8, ptr @enzyme_const_constants, i64 3520), i64 160, i1 false), !dbg !147
 ```

For forward mode autodiff, enzyme maps this to the following memset with destination aligment 16.

```
  tail call void @llvm.memset.p0.i64(ptr nonnull align 16 %"invariant.gep2103'ipg", i8 0, i64 160, i1 true), !dbg !4558
```

This is causing a segfault in the code as the particular destination vector only has an aligment of 8.

I think the source of this is this particular line in Enzyme, which seems to be using the source alignment rather than the destination. I'm not familar with this code, or alignment issues. Hence this issue to check the validity of my reasoning before I go any further!

https://github.com/EnzymeAD/Enzyme/blob/20a407618fdbc28e6720cbe5c9ccad31040577fb/enzyme/Enzyme/AdjointGenerator.h#L3677

### PR DESCRIPTION
**Pull request by martinjrobins**
memcpy -> memset fwd rule uses destination alignment

fixes #2663

### DIFF
--- a/enzyme/test/Enzyme/ForwardMode/memcpy-align.ll
+++ b/enzyme/test/Enzyme/ForwardMode/memcpy-align.ll
@@ -19,7 +19,7 @@ declare void @llvm.memcpy.p0i8.p0i8.i64(i8* nocapture writeonly, i8* nocapture r
 ; Function Attrs: nounwind uwtable
 define dso_local void @dmemcpy_align(double* %dst, double* %dstp1, double* %src, i64 %n) local_unnamed_addr #0 {
 entry:
-  tail call void (...) @__enzyme_fwddiff.f64(void (double*, double*, i64)* nonnull @memcpy_align, metadata !"enzyme_width", i64 1, metadata !"enzyme_dup", double* %dst, double* %dstp1, metadata !"enzyme_const", double* %src, i64 %n)
+  tail call void (...) @__enzyme_fwddiff.f64(void (double*, double*, i64)* nonnull @memcpy_align, metadata !"enzyme_dup", double* %dst, double* %dstp1, metadata !"enzyme_const", double* %src, i64 %n)
   ret void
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: enzyme/test/Enzyme/ForwardModeVector/memcpy-align.ll [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+  ret void
+}
+
+; Function Attrs: argmemonly nounwind
+declare void @llvm.memcpy.p0i8.p0i8.i64(i8* nocapture writeonly, i8* nocapture readonly, i64, i1) #1
+
+; Function Attrs: nounwind uwtable
+define dso_local void @dmemcpy_align(double* %dst, double* %dstp1, double* %src, i64 %n) local_unnamed_addr #0 {
+entry:
+  tail call void (...) @__enzyme_fwddiff.f64(void (double*, double*, i64)* nonnull @memcpy_align, metadata !"enzyme_width", i64 1, metadata !"enzyme_dup", double* %dst, double* %dstp1, metadata !"enzyme_const", double* %src, i64 %n)
```
wsmoses: if this isn't a vector mode test, it should be but in just forwardmode/

also you can remove the enzyme_width arg [and corersponding i64 1 thereafter]

PR #2676, opened 2026-01-27
[mlir] Support mincut with nested regions

pengmai:
* Add option to print GraphViz mincut graph

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
@@ -23,6 +23,12 @@ using namespace mlir::enzyme;
 
 #define DEBUG_TYPE "enzyme-mincut"
 
+static llvm::cl::opt<bool>
+    DebugGraphviz("mincut-print-graphviz", llvm::cl::init(false),
+                  llvm::cl::Hidden,
+                  llvm::cl::desc("Use with DEBUG_TYPE 'enzyme-mincut' to print "
+                                 "the mincut graphs in GraphViz"));
+
 void mlir::enzyme::localizeGradients(OpBuilder &builder,
                                      MGradientUtilsReverse *gutils,
                                      Block *fwd) {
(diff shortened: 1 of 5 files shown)


enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -23,6 +23,10 @@ using namespace mlir::enzyme;
 
 #define DEBUG_TYPE "enzyme-mincut"
 
+// Set to true to dump a graphviz representation of the mincut graph. Nice for
+// debugging.
+#define DEBUG_GRAPHVIZ false
```
wsmoses: is this worth turning into an llvm::cl option so it can be enabled on the cli?

wsmoses (approved):
lgtm, but let @Pangoraw take a look at

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -391,6 +467,23 @@ void mlir::enzyme::minCutCache(Block *forward, Block *reverse,
     return;
   }
 
+  // Maintain a mapping of forward to reverse blocks. We later use this to place
+  // the new cache pops and cloned ops to the correct blocks.
+  DenseMap<Block *, OpBuilder::InsertPoint> insertionPointMap;
+  for (const auto &info : caches) {
+    Block *fwdBlock = info.pushOp->getBlock();
```
Pangoraw, line 476: I am not sure I fully understand this here. Since the removal works from inner most to outer-most. At this point all pushes should be in `fwdBlock` not in nested blocks. Am I missing something?

pengmai: Right, so the motivation here is to actually override the behaviour of going inner-to-outer. The inner-to-outer traversal can actually make suboptimal decisions when the inner block executes at most once (e.g. `scf.if`) because it works greedily.

In the new test case, `affine_parallel_mincut.mlir`, running mincut block-by-block results in 3 values being cached in the outer loop. If we instead keep the push/pops in the inner block and run this modified mincut on the outer `parallel` op, only 2 values are cached. In LBM, this translates to about 5 values being cached vs 60 values.

wsmoses: so specifically for things that aren't potentially run multiple times, its better to do the mincut on all variables in that scope. We can start conservatively to begin with assuming all regions are unknown execution count, and for ifs/other opt ins we have the mincut run across all thereof

Pangoraw: Oh makes sense, I had missed the preserve_cache mechanism. Can we also run mincut in the RemoveEnzymeOpsInterface implementation for scf.if so it can benefit even when it's not within a loop?

Pangoraw (approved):
>  In LBM, this translates to about 5 values being cached vs 60 values.

awesome!

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
pengmai: I agree that running mincut on scf.if alone would be useful, though I'd like to make it a follow-up PR so I can prioritize getting the C++ benchmarks working.

PR #2692, opened 2026-02-04
ProbProg: Pure StableHLO Migration

sbrantq:


enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -1026,6 +814,70 @@ def SelectOp : Enzyme_Op<"select", [Pure]> {
   }];
 }
 
+def ReshapeOp : Enzyme_Op<"reshape", [Pure]> {
```
wsmoses, line 817: This [and the usage of below] is extremely specific to stablehlo. Longer term we're going to need to lower this for generic mlir dialects [like arith/mutable memory/etc] for use in c++ and beyond.

Given the tight push right now this is okay as is, but it's important to keep the design extensible

sbrantq: Yeah will refactor in a future PR

PR #2695, opened 2026-02-08
ProbProg: Raw MCMC

sbrantq:


enzyme/Enzyme/MLIR/Dialect/Ops.cpp
```diff
@@ -1062,13 +1062,21 @@ LogicalResult MHOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
+             << getFn().value() << "' does not reference a valid global funcOp";
+  }
+
+  if (auto logpdfAttr = getLogpdfFnAttr()) {
+    auto global =
+        symbolTable.lookupNearestSymbolFrom<func::FuncOp>(*this, logpdfAttr);
+    if (!global)
+      return emitOpError("'") << logpdfAttr.getValue()
+                              << "' does not reference a valid global funcOp";
+  }
```
wsmoses, line 1079: is it useful to check that one of these two is set ? [e.g. if getfnattr else getlogpdffnattr else failure ?

wsmoses (approved):

enzyme/Enzyme/MLIR/Dialect/Ops.cpp
sbrantq: Yes indeed and done

PR #2697, opened 2026-02-10
[mlir] bugfix when some scf/affine.for results are inactive

pengmai:
The derivatives with respect to primal loop results are passed in as iteration arguments to the reversed loop. Therefore, the loop differentiation previously assumed that the number of primal results is the same as the number of reverse iteration args. However, this is not the case if some of the primal loop results are inactive, since these results won't have derivatives.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
@@ -412,14 +412,15 @@ public:
 
         auto term = oBB.getTerminator();
 
-        for (auto &&[active, arg, operand] :
-             llvm::zip_equal(operandsActive, revBB.getArguments().slice(1),
-                             term->getOperands())) {
+        unsigned argIdx = 1; // Skip over the reversed IV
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -451,11 +451,12 @@ struct ForOpInterfaceReverse
...
 
-    for (auto &&[active, res, arg] : llvm::zip_equal(
-             operandsActive, repFor->getResults(), forOp.getInitArgs())) {
+    unsigned resIdx = 0;
+    for (auto &&[active, arg] :
+         llvm::zip_equal(operandsActive, forOp.getInitArgs())) {
       if (active) {
         if (!gutils->isConstantValue(arg))
-          gutils->addToDiffe(arg, res, builder);
+          gutils->addToDiffe(arg, repFor.getResult(resIdx++), builder);
```
wsmoses: The postfix operator here is always a bit unclear to readers, can you put the increment on a different line?

Pangoraw (approved):

PR #2699, opened 2026-02-11
fix the crash in eraseFictiousPHIs

minansys:
# Fix crash in `eraseFictiousPHIs()` by rematerializing `!enzyme_fromstack` malloc/calloc placeholders into entry-block allocas
## Bug
https://github.com/EnzymeAD/Enzyme/issues/2700

## Summary
This PR fixes a hard abort in Enzyme reverse-mode differentiation where `GradientUtils::eraseFictiousPHIs()` encounters a “fictitious” placeholder PHI node that still has live uses and triggers an assertion (`pp->getNumUses() == 0`). The failure was observed when the placeholder corresponded to `malloc`/`calloc` calls annotated with `!enzyme_fromstack`.

Instead of asserting (or replacing with `undef`), Enzyme now rematerializes `!enzyme_fromstack` allocations as a real stack allocation (`alloca`) in the entry block, rewires uses to the new alloca, and erases the placeholder PHI. This matches the intended semantics of `!enzyme_fromstack` (stack-like/rematerializable scratch allocation) and prevents compilation from aborting.

---

## Background / Motivation
Enzyme’s reverse-mode AD works by transforming LLVM IR: cloning and restructuring control flow, inserting tape operations, and erasing or rewriting instructions that are not needed for gradient computation. During these transformations, Enzyme may call `eraseWithPlaceholder()` which replaces an instruction with a temporary “fictitious” PHI node to keep IR well-formed while the transformation is in progress.

At the end, `eraseFictiousPHIs()` is expected to remove these placeholders, assuming they have become dead (0 uses). However, in some pipelines a placeholder PHI can remain *live*—particularly when it represents a `malloc`/`calloc` call tagged with `!enzyme_fromstack`, and downstream GEP/load/select chains still reference it. Previously, this resulted in:
- a diagnostic like `Illegal replace ficticious phi...`
- followed by a hard assert and compilation abort.

This is especially disruptive in CUDA/NVPTX device compilation and other complex CFG cases where these patterns are more common.

---

### 1) Rematerialize `!enzyme_fromstack` malloc/calloc placeholders
In `GradientUtils::eraseFictiousPHIs()`:

- If a fictitious placeholder `PHINode* pp` still has uses
- and `pair.second` (the mapped original value) is a `CallInst` to `malloc` or `calloc`
- and the call has metadata `!enzyme_fromstack`

Then Enzyme:
1. Creates an entry-block `alloca i8, byteCount` that dominates all uses
2. Casts it as needed to match the PHI’s pointer type (bitcast or addrspacecast)
3. For `calloc`, inserts a `memset` to zero-initialize the buffer
4. Replaces all uses of the placeholder PHI with the alloca
5. Erases the placeholder PHI

This prevents the illegal-live-placeholder condition from aborting compilation and preserves the intended “stack-like scratch allocation” semantics.

### 2) Address-space correctness
To support targets such as NVPTX where pointer address spaces may differ, the replacement handles:
- `addrspacecast` when the PHI pointer address space differs from the `alloca`’s address space
- `bitcast` otherwise

### 3) Fallback behavior (unchanged intent)
If the placeholder PHI remains live but cannot be safely rematerialized under the `!enzyme_fromstack` rules, Enzyme retains diagnostics and falls back to IR-valid behavior (replacing uses with `undef` and erasing) rather than crashing. The PR avoids relying on an assert for correctness in the metadata-guided rematerializable case.

---

## Why this is correct
- `!enzyme_fromstack` explicitly communicates that the allocation can be treated as stack-like/rematerializable scratch memory.
- Rematerializing as an entry-block `alloca` ensures:
  - dominance over all uses
  - well-formed SSA without introducing new PHI cycles
- `calloc` zeroing is preserved via `memset`, matching expected semantics for scratch buffers that depend on initialization.

---

## Reproduction (before)
A reverse-mode Enzyme pipeline aborts during compilation with errors similar to:
- `Illegal replace ficticious phi for: %X = phi ptr of call @malloc(...), !enzyme_fromstack`
- followed by an assertion in `GradientUtils::eraseFictiousPHIs()`.

Disabling `-enzyme-phi-restructure` does not resolve the issue, indicating the core failure is not only PHI restructuring but the survival of placeholder PHIs for `!enzyme_fromstack` allocations.

---

## Testing
- ✅ Verified the previously failing CUDA/NVPTX build completes successfully after this patch.
- ✅ Confirmed Enzyme no longer aborts on `!enzyme_fromstack` malloc/calloc placeholder PHIs in the reported workload.

**Recommended follow-up:**
- Add a regression test that differentiates a function containing `malloc`/`calloc` annotated with `!enzyme_fromstack`, and verify Enzyme:
  - does not assert
  - produces valid LLVM IR

---

## Risk / Compatibility
- The change is scoped to `malloc`/`calloc` calls that carry `!enzyme_fromstack` metadata.
- Other placeholder PHI behavior is unchanged except avoiding a hard abort in the rematerializable case.
- Potential tradeoff: additional entry-block allocas (expected and consistent with metadata semantics).

---

## Follow-ups
- Add a dedicated regression test for `!enzyme_fromstack` placeholder rematerialization.
- If additional “live placeholder” patterns appear (non-fromstack), handle them in similarly targeted, semantics-preserving rematerialization rules.

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -8633,26 +8633,172 @@ void GradientUtils::eraseFictiousPHIs() {
     phis.emplace_back(pair.first, pair.second);
   fictiousPHIs.clear();
 
+  auto *i8Ty = Type::getInt8Ty(newFunc->getContext());
+
+  // Helper: create an i8 alloca in the entry block that dominates all uses.
+  auto createEntryAllocaBytes = [&](PHINode *pp, Value *byteCount) -> Value * {
+    IRBuilder<> B(&*pp->getFunction()->getEntryBlock().getFirstInsertionPt());
+
+    Value *sz = byteCount;
+    if (sz->getType()->isIntegerTy() && sz->getType()->getIntegerBitWidth() != 64)
+      sz = B.CreateZExtOrTrunc(sz, B.getInt64Ty(), "enzyme.sz64");
+    else if (!sz->getType()->isIntegerTy())
+      sz = B.CreatePtrToInt(sz, B.getInt64Ty(), "enzyme.sz64");
+
+    AllocaInst *AI = B.CreateAlloca(i8Ty, sz, "enzyme_fromstack_alloca");
+    // Alignment is not critical here; 16 is usually safe and helps codegen.
+    AI->setAlignment(Align(16));
+
+    Value *rep = AI;
+
+    // If pointer address spaces differ, cast appropriately.
+    if (pp->getType() != AI->getType() && pp->getType()->isPointerTy() &&
+        AI->getType()->isPointerTy()) {
+      unsigned ppAS = cast<PointerType>(pp->getType())->getAddressSpace();
+      unsigned aiAS = cast<PointerType>(AI->getType())->getAddressSpace();
+      if (ppAS != aiAS) {
+        rep = B.CreateAddrSpaceCast(AI, pp->getType(), "enzyme.asc");
+      } else {
+        rep = B.CreateBitCast(AI, pp->getType(), "enzyme.bc");
+      }
+    }
+
+    return rep;
+  };
+
+  // Helper: handle malloc/calloc tagged !enzyme_fromstack by turning into alloca
+  auto tryRematFromstackMalloc = [&](PHINode *pp, Value *orig) -> bool {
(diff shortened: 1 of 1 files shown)


minansys:
This PR is not needed

PR #2705, opened 2026-02-17
bazel: export PreserveNVVM

vimarsh6739:
This is needed for proper resolution of dependencies in the Reactant C++ frontend. 

[https://github.com/EnzymeAD/Reactant/pull/28](https://github.com/EnzymeAD/Reactant/pull/28)

### DIFF
--- a/enzyme/BUILD
+++ b/enzyme/BUILD
@@ -132,6 +132,29 @@ gentbl_cc_library(
 )
 
 cc_library(
+    name = "EnzymePreserveNVVM",
+    srcs = [
+        "Enzyme/Utils.cpp",
+        "Enzyme/PreserveNVVM.cpp"
+    ],
+    hdrs = glob([
+        "Enzyme/*.h",
+        "Enzyme/TypeAnalysis/*.h",
+    ]),
+    copts = [
+        "-Wno-unused-variable",
+        "-Wno-return-type",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":blas-attributor",
+        "@llvm-project//llvm:Analysis",
+        "@llvm-project//llvm:CodeGen",
+        "@llvm-project//llvm:Core"
+    ],
+)
+
+cc_library(
     name = "EnzymeStatic",
     srcs = glob(
         [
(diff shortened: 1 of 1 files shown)


enzyme/BUILD
```diff
@@ -131,51 +131,38 @@ gentbl_cc_library(
...
+
 cc_library(
-    name = "EnzymeStatic",
+    name = "EnzymeFrontendSupport",
     srcs = glob(
         [
             "Enzyme/*.cpp",
             "Enzyme/TypeAnalysis/*.cpp",
+        ],
+        exclude = [
```
wsmoses: can enzymefrontendsupport be written to depend on enzymestatic ?

vimarsh6739: I think it's the opposite here, EnzymeStatic depends on EnzymeFrontendSupport. We just need to reuse the passes (like PreserveNVVM + JLInstSimplify) in Reactant's version of EnzymeStatic. 

Some of the symbols are basically unresolved(isTargetNVPTX and an alias check)

wsmoses: The reactant version of those things doesn't exist at all and should be deleted except preserve nvvm which also now does some different things (and should be renamed to a new pass)

enzyme/Enzyme/Utils.cpp
```diff
@@ -4216,15 +4216,6 @@ arePointersGuaranteedNoAlias(TargetLibraryInfo &TLI, llvm::AAResults &AA,
   return {};
 }
 
-bool isTargetNVPTX(llvm::Module &M) {
```
wsmoses, line 4219: also rm this from utils.h then

enzyme/BUILD
```diff
@@ -195,6 +195,38 @@ cc_library(
     alwayslink = 1,
 )
 
+cc_library(
+    name = "EnzymeFrontendSupport",
```
wsmoses: just call this EnzymePreserveNVVM [and make it a dependency of the other target]

enzyme/BUILD
```diff
@@ -195,6 +195,38 @@ cc_library(
...
+        "@llvm-project//llvm:InstCombine",
+        "@llvm-project//llvm:Passes",
+        "@llvm-project//llvm:Scalar",
+        "@llvm-project//llvm:Support",
+        "@llvm-project//llvm:Target",
+        "@llvm-project//llvm:TargetParser",
+        "@llvm-project//llvm:TransformUtils",
+        "@llvm-project//llvm:config",
+    ],
+    alwayslink = 1,
```
wsmoses: don't alwayslink

enzyme/BUILD
```diff
@@ -195,6 +195,29 @@ cc_library(
     alwayslink = 1,
 )
 
+cc_library(
+    name = "EnzymePreserveNVVM",
```
wsmoses: Can you make enzymestatic depend on this ? [and and remove utils and preservenvvm.cpp as requirements on enzymestatic]

vimarsh6739: Just excluded both from the glob

PR #2710, opened 2026-02-21
Export PreserveNVVM from CMake and Bazel

vimarsh6739:
We are adding a new target to Enzyme to enable Reactant C++ builds, since we deleted a bunch of source files from there (accompanied by further fixes in [pr](https://github.com/EnzymeAD/Reactant/pull/28))

### DIFF
--- a/enzyme/Enzyme/CMakeLists.txt
+++ b/enzyme/Enzyme/CMakeLists.txt
@@ -165,40 +165,16 @@ endif()
 if (${ENZYME_ENABLE_REACTANT})
   set(ENZYME_REACTANT_SUPPORT_SRC
     PreserveNVVM.cpp
-    Utils.cpp
-    TypeAnalysis/TypeAnalysis.cpp
-    TypeAnalysis/TypeTree.cpp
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/CMakeLists.txt
```diff
@@ -161,6 +161,68 @@ if (${ENZYME_EXTERNAL_SHARED_LIB})
         COMPONENT dev)
 endif()
 
+# Reactant-focused Enzyme subset as one static lib.
+if (${ENZYME_ENABLE_REACTANT})
+  set(ENZYME_REACTANT_SUPPORT_SRC
+    PreserveNVVM.cpp
+    TypeAnalysis/TypeAnalysis.cpp
```
wsmoses: I don't think any of this is required, just preserve nvvm and its deps are

vimarsh6739: For now I just copied what's in bazel, I'm adding the CMake bits in the other repo before coming back to prune this

Issue #2711, opened 2026-02-23
Enzyme crash for partial accumulation


minansys:
This is in CPU and is likely due to one function's return is a struct of 4 float number struct DiffusionVariables { float A; float B; float C; float D}. It works fine in O0 but not in O1 and O3. Here is a reproducer: https://tinyurl.com/22jnzhgk

```
 VT: <2 x float> idxs:{i32 0,} start=0 size=4 storeSize=8 val=  %13 = tail call { <2 x float>, <2 x float> } @_ZN8Cromwell11EnergyModel7Helpers32AnisotropicConductivityOperationILNS_13GeometryTools4Mesh17AdjacencyOrderingE2EE23GetAnisotropicDiffusionEjPKjbRKNS_10Containers21ColumnMajorMatrixViewIKfEESE_SE_(ptr noundef nonnull align 8 dereferenceable(336) %8, i32 noundef %1, ptr noundef %2, i1 noundef zeroext true, ptr noundef nonnull align 8 dereferenceable(12) %9, ptr noundef nonnull align 8 dereferenceable(12) %0, ptr noundef nonnull align 8 dereferenceable(12) %10) #45

[31mclang++: /home/user/Enzyme/enzyme/Enzyme/DiffeGradientUtils.cpp:352: llvm::SmallVector<llvm::SelectInst*, 4> DiffeGradientUtils::addToDiffe(llvm::Value*, llvm::Value*, llvm::IRBuilder<>&, llvm::Type*, unsigned int, unsigned int, llvm::ArrayRef<llvm::Value*>, llvm::Value*, size_t): Assertion `0 && "unhandled accumulate with partial sizes"' failed.


1.  <eof> parser at end of file
2.  Optimizer
3.  Running pass "EnzymeNewPM" on module "/home/mixu/dev/vNNN/Cromwell/Cromwell/CFD/EnergyModel/src/Operators/EnergyResidual.cxx"
 #0 0x00001507b3a8f86b llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/mixu/dev/vNNN/Cromwell/Dependencies/clang/linx64/bin/../lib/libLLVM.so.19.1+0x7f886b)
 #1 0x00001507b3a8d7e4 llvm::sys::CleanupOnSignal(unsigned long) (/home/mixu/dev/vNNN/Cromwell/Dependencies/clang/linx64/bin/../lib/libLLVM.so.19.1+0x7f67e4)
[...]

PR #2712, opened 2026-02-23
Fix crash on partial-window accumulation into fixed-width vectors

minansys:
Note this is generated by chatgpt

### Problem

fix the issue
https://github.com/EnzymeAD/Enzyme/issues/2711

Enzyme could crash (assert) during reverse-mode differentiation when LLVM lowers small aggregates into vector chunks and only a *byte slice* of such a vector is considered active.

A common pattern is an aggregate return lowered to:

- `{ <2 x float>, <2 x float> }`

and then only the first `float` lane is consumed (e.g., by loading only 4 bytes from the stored `<2 x float>`). In this case, Enzyme’s activity/type analysis can request a *partial window* accumulation into a vector-typed differential:

- `VT: <2 x float> ... start=0 size=4 storeSize=8`

Enzyme previously handled partial windows for structs/arrays, but not for vectors. When `VT` was a fixed vector type and `start/size` described a subset of the vector storage, Enzyme hit:

## Root cause

`DiffeGradientUtils::addToDiffe(..., start, size, ...)` supports accumulating gradients for **partial byte ranges** of a value. This is needed when only a subset of an aggregate is active.

Before this change, Enzyme’s partial-window logic had explicit handling for:

- `StructType`
- `ArrayType`

When LLVM represents aggregates as **fixed vectors** (e.g. `<2 x float>`), Enzyme can still compute partial activity windows (e.g. one float lane → `start=0 size=4`). However, the vector case fell through to the debug print and then:

- `assert(0 && "unhandled accumulate with partial sizes")`

leading to a crash.

## Fix

Add support for partial-window accumulation into **fixed-width vectors** (`VectorType` with non-scalable element counts) when the requested byte window is **element-aligned**.

### Implementation details

1. Detect `VT` as a `VectorType` and require:
   - non-scalable vectors
   - `start` and `size` are multiples of the element size (in bytes)

2. Convert the partial-window differential into a full vector differential by **masking lanes**:
   - Build `masked = <0, 0, ...>`
   - For lanes covered by `[left_idx, right_idx)`, copy lane values from the sub-differential (`dsub`) into `masked`
   - Leave all other lanes as zero

3. Delegate to the existing full-vector accumulation path:
   - call `addToDiffe(val, masked, ...)`

4. Handle both:
   - `getWidth() == 1` (single differential slice)
   - `getWidth() > 1` (multiple slices) by extracting the correct sub-differential via `extractMeta` and applying the same lane masking per slice

This preserves Enzyme semantics: only the active lanes receive gradient contributions; inactive lanes accumulate zero.

## Why this approach

- **General:** Works for any fixed vector width and element type (`float`, `double`, etc.) as long as the window aligns to element boundaries.
- **Safe:** Does not attempt to handle partial byte ranges *within* an element (which would require more complex packing/bitcasting rules).
- **Minimal impact:** Reuses Enzyme’s existing, well-tested vector accumulation logic by converting partial windows into masked full-vector updates.

## Testing

Added a regression test (LLVM IR) that triggers the previous failure mode: partial use of a `<2 x float>` extracted from a `{ <2 x float>, <2 x float> }` return.

- **Before:** Enzyme asserts with `"unhandled accumulate with partial sizes"`.
- **After:** Enzyme generates the derivative without crashing.


minansys:
@wsmoses I just noticed that this is not merged yet. Whether could you please review and approve it? Thanks!

enzyme/Enzyme/DiffeGradientUtils.cpp
```diff
@@ -344,6 +344,65 @@ SmallVector<SelectInst *, 4> DiffeGradientUtils::addToDiffe(
...
+
+        unsigned numElts = VecT->getElementCount().getFixedValue();
+        if (left_idx > numElts)
+          left_idx = numElts;
+        if (right_idx > numElts)
+          right_idx = numElts;
+
+        auto maskVec = [&](Value *dsub) -> Value * {
+          Value *masked = Constant::getNullValue(VT);
+          for (unsigned i = left_idx; i < right_idx; i++) {
```
wsmoses, line 387: minor optimization if start == 0 && elemBytes == size, you can just shortcut and return dsub

enzyme/test/Enzyme/ReverseMode/partial_vec_window.ll
```diff
@@ -0,0 +1,39 @@
...
+}
+
+define float @test_derivative(float %x) {
+entry:
+  %d = call float (float (float)*, ...) @__enzyme_autodiff(float (float)* @tester, float %x)
+  ret float %d
+}
+
+declare float @__enzyme_autodiff(float (float)*, ...)
+; CHECK: @diffetester
```
wsmoses, line 39: also add all the check-next lines [see other tests for examples]

wsmoses: also can you add a batched reverse mode test [to confirm that code path works],

as well as for the masked version

wsmoses (approved):

enzyme/Enzyme/DiffeGradientUtils.cpp
minansys: @copilot resolve this issue

enzyme/test/Enzyme/ReverseMode/partial_vec_window.ll
minansys: @copilot resolve this comments

PR #2722, opened 2026-03-03
MLIR:  Add forward mode support for scf.parallel, scf.reduce, and scf.reduce.return

etphipp:
These changes add forward mode differentiation support for the scf.parallel, scf.reduce, and scf.redcue.return MLIR operations.  Two tests are included, one with a reduce and one without.  Most of this is fairly straightforward, but because of peculiarities of these operations, in particular that the assumption that shadows follow primals is violated with scf.reduce, there was some intrusion into CoreDialectsAutoDiffImplementations.cpp.  I would be happy to do this differently if there is a better way.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
@@ -1018,9 +1018,13 @@ struct ParallelOpADDataFlow
     const size_t num_upper = parOp.getUpperBound().size();
     const size_t num_step = parOp.getStep().size();
     const size_t init_vals_offset = num_lower + num_upper + num_step;
-    return {
-        parOp->getOperand(res.getResultNumber() + init_vals_offset),
-        parOp.getBody()->getTerminator()->getOperand(res.getResultNumber())};
+    return {parOp->getOperand(res.getResultNumber() + init_vals_offset),
+            parOp.getBody()
+                ->getTerminator()
+                ->getRegion(res.getResultNumber())
+                .front()
+                .getTerminator()
+                ->getOperand(0)};
   }
   SmallVector<Value> getPotentialIncomingValuesArg(Operation *op,
                                                    BlockArgument arg) const {
@@ -1030,14 +1034,11 @@ struct ParallelOpADDataFlow
   }
   SmallVector<Value> getPotentialTerminatorUsers(Operation *op, Operation *term,
                                                  Value val) const {
-    auto parOp = cast<scf::ParallelOp>(op);
     SmallVector<Value> sv;
 
-    for (auto &&[res, arg, barg] : llvm::zip_equal(
-             parOp->getResults(), term->getOperands(), parOp.getInitVals())) {
+    for (auto [idx, arg] : llvm::enumerate(term->getOperands())) {
       if (arg == val) {
-        sv.push_back(res);
-        sv.push_back(barg);
+        sv.push_back(term->getRegion(idx).front().getArgument(0));
       }
     }
 


enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
```diff
@@ -265,12 +266,10 @@ void mlir::enzyme::detail::returnReverseHandler(Operation *op,
...
     Operation *origTerminator, OpBuilder &builder, MGradientUtils *gutils) {
   auto parentOp = origTerminator->getParentOp();
-
   llvm::SmallDenseSet<unsigned> operandsToShadow;
   auto termIface = dyn_cast<RegionBranchTerminatorOpInterface>(origTerminator);
-  auto regionBranchOp =
-      dyn_cast<RegionBranchOpInterface>(origTerminator->getParentOp());
-  if (termIface && regionBranchOp) {
+  auto regionBranchOp = dyn_cast<RegionBranchOpInterface>(parentOp);
+  if (termIface && regionBranchOp && !isa<scf::ParallelOp>(parentOp)) {
```
wsmoses: we shouldn't special case scf-parallel here, instead can we either make a new interface or use a different interface for defining this specifically for scf::parallel ?

wsmoses: [that way we can handle other things like affine::ParallelOp / etc in the future without hardcoding]

wsmoses: at this point I might just attach the forward mode interface manually for scf::parallelop instead of using the tablegen handler (which just forces a call into here)

etphipp: Yeah, I didn't like doing that either.  I think it needs an entirely new interface because scf.parallel and scf.reduce behave very differently from what is expected from RegionBranchOpInterface/RegionBranchTermiantorOpInterface.  Can you provide a little more detail for what you mean by "attach the forward mode interface manually"?

etphipp: I moved the special handling of scf.reduce and scf.reduce.return into specializations of AutoDiffOpInterface in SCFAutoDiffOpInterfaceImpl.cpp using some similar specializations for linalg as an example.  Is that what you're looking for?  There is still the question of what to do with the small special handling of scf.parallel in controlFlowHandler().  I can do something similar, although it will duplicate almost all of that function implementation.  It is needed because `scf::ParallelOp` doesn't implement `RegionBranchOpInterace::getEntrySuccessorOperands()` with anything useful (returns an empty range).

wsmoses: yeah so I think we can add a new interface and first check if our special interface exists, otherwise fall back to `RegionBranchOpInterace`. that way we can reuse code. worst case we can copy it in a similar manual extension, but re-use is nice if possible

etphipp: Do you mean an interface to specialize all of controlFlowForwardHandler(), or just the extraction of the successor operands/initial values?  If the former, it isn't clear to me how that reuses code, but I'm happy to do whatever you want.

wsmoses: I was thinking just `RegionBranchOpInterace::getEntrySuccessorOperands` (which correct me if I'm wrong is the only part we need at the moment as different from the regular controlFlowForwardHandler)

wsmoses: also cc @ftynse for thoughts here

etphipp: How about piggy backing off of ControlFlowAutoDiffOpInterface which we already specialize?  I added a new interface method to it to extract the successor operands and specialized it for scf.parallel.

wsmoses: yeah works for me!

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -1008,6 +1008,256 @@ struct ForOpADDataFlow
...
+                                                   OpResult res) const {
+    auto parOp = cast<scf::ParallelOp>(op);
+    const size_t num_lower = parOp.getLowerBound().size();
+    const size_t num_upper = parOp.getUpperBound().size();
+    const size_t num_step = parOp.getStep().size();
+    const size_t init_vals_offset = num_lower + num_upper + num_step;
+    return {
+        parOp->getOperand(res.getResultNumber() + init_vals_offset),
+        parOp.getBody()->getTerminator()->getOperand(res.getResultNumber())};
+  }
```
etphipp, line 1028: I'm now wondering if this is the best logic here.  In particular, line 1023 gets the operand of the scf.reduce terminator corresponding to the scf.parallel result, which is bypassing the reduce block.  I wonder if it would be better to instead do:
```
return {
        parOp->getOperand(res.getResultNumber() + init_vals_offset),
        parOp.getBody()->getTerminator()->getRegion(res.getResultNumber()).front().getTerminator()->getOperand(0)
}
```
which gets the operand of the scf.reduce.return op for the reducer region that corresponds to the result, which would allow the data flow analysis to flow through the reduce block.

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -1008,6 +1008,256 @@ struct ForOpADDataFlow
...
+    SmallVector<Value> sv;
+
+    for (auto &&[res, arg, barg] : llvm::zip_equal(
+             parOp->getResults(), term->getOperands(), parOp.getInitVals())) {
+      if (arg == val) {
+        sv.push_back(res);
+        sv.push_back(barg);
+      }
+    }
+
```
etphipp, line 1044: Similar question here as I wonder if it would be better to do:
```
    for (auto [idx,arg] : llvm::enumerate(term->getOperands())) {
      if (arg == val) {
        sv.push_back(term->getRegion(idx).front().getArgument(0));
      }
    }
```
which gets the operand of the scf.reduce.return op instead of getting the result and the operand of the scf.reduce op.

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
wsmoses: I think the reduce return here makes the most sense

etphipp: OK, I pushed that change along with the corresponding one below.

PR #2734, opened 2026-03-11
mlir: mark loads of read-only pointers as movable during mincut

pengmai:
* implement MemoryEffectOpInterface for enzyme.push/pop
* fix warnings with int signedness

### DIFF
--- a/enzyme/Enzyme/MLIR/Analysis/DataFlowAliasAnalysis.cpp
+++ b/enzyme/Enzyme/MLIR/Analysis/DataFlowAliasAnalysis.cpp
@@ -30,6 +30,7 @@
 #include "mlir/Analysis/AliasAnalysis.h"
 #include "mlir/Analysis/DataFlow/DenseAnalysis.h"
 #include "mlir/Analysis/DataFlow/SparseAnalysis.h"
+#include "mlir/Analysis/DataFlow/Utils.h"
 #include "mlir/Analysis/DataFlowFramework.h"
 #include "mlir/Dialect/LLVMIR/LLVMAttrs.h"
 #include "mlir/Interfaces/FunctionInterfaces.h"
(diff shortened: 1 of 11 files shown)


enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -279,23 +279,23 @@ def PushOp : Enzyme_Op<"push", [
...
+  let arguments = (ins Arg<AnyType, "the cache to push to", [MemWrite]>:$cache, AnyType:$value);
 }
 
 def PopOp : Enzyme_Op<"pop", [
     TypesMatchWith<"type of 'output' matches element type of 'cache'",
                    "cache", "output",
                    "::llvm::cast<enzyme::CacheType>($_self).getType()">]> {
   let summary = "Retrieve information for the reverse mode pass.";
-  let arguments = (ins AnyType : $cache);
+  let arguments = (ins Arg<AnyType, "the cache to pop from", [MemRead]>:$cache);
```
wsmoses: so this isn't purely memread, because it also removes an element. E.g. we want to avoid accidentally rm'ing a pop just because the result is unused since then indexing will be wrong for other push/pops.

pengmai: Would it be sufficient to also add a `MemWrite`? `[MemRead, MemWrite]` is how `memref.atomic_rmw` is defined.

wsmoses: I think so [cc @ftynse]. technically it more writes to the "length" instead of the actual position, but that is encompoased within the write to a fake -1 index of the stack pointer itself or something

wsmoses (approved):

enzyme/Enzyme/MLIR/Analysis/DataFlowAliasAnalysis.cpp
```diff
@@ -1194,3 +1195,47 @@ void enzyme::AliasAnalysis::visitExternalCall(
...
+  if (failed(solver.initializeAndRun(funcOp))) {
+    assert(false && "dataflow analysis failed");
+    return;
+  }
+
+  // Determine all alias classes that were modified
+  AliasClassLattice modified(nullptr);
+  funcOp.walk([&](MemoryEffectOpInterface memory) {
+    SmallVector<MemoryEffects::EffectInstance> effects;
+    memory.getEffects(effects);
```
wsmoses, line 1226: Is there an issue here for an op that doesn't define memoryopinterfqce but could touch memory?

wsmoses: Long term this isn't sufficient internally as in split mode AD the local function may be read only but the caller function could modify memory between said calls.

For now we only do combined mode, but when @Pangoraw's work on this lands we will have an issue

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -344,11 +344,18 @@ static inline void bfs(const Graph &G, const llvm::SetVector<Value> &Sources,
   }
 }
 
+static inline bool isLoadMovable(Operation *op) {
+  if (!hasEffect<MemoryEffects::Read>(op)) {
```
wsmoses: I think this is wrong, I think this needs to check that it does not have a memory effect write and then return false [but if it only has a read it is fine]

enzyme/Enzyme/MLIR/Analysis/DataFlowAliasAnalysis.cpp
pengmai: The alias analysis will set the lattices to unknown states in the event of ops that don't define MemoryEffectOpInterface, but I can add an additional check here to be safe.

Regarding split mode, IIRC you mentioned having an additional array that indicates if a function argument is modified outside of the body between invocations. I'm not sure how to implement this feature without that explicit information. I can add a TODO in the meantime

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
pengmai: I'll make this `hasSingleEffect<MemoryEffects::Read>` so we only consider ops that solely load (excluding allocations and frees, in addition to things like `atomic_rmw`)

wsmoses: Sure

PR #2750, opened 2026-03-18
scf.for fix when iter-args and result args have shadow mismatch in fwd

spikerheado1234:
In forward mode AD, if an iter-arg is deemed a constant, yet its corresponding result has a tangent, then there will be a mismatch between the number of iter-args and result-args. This PR fixes that by unioning if either an iter-arg or its corresponding result is active, inserting a corresponding shadow arg if at least one is active.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
@@ -349,13 +349,31 @@ LogicalResult mlir::enzyme::detail::controlFlowForwardHandler(
                              : regionBranchOp.getSuccessorInputs(successor);
 
     // Need to know which of the arguments are being forwarded to from
-    // operands.
+    // operands. An operand needs a shadow — and the ForOp needs a matching
+    // shadow result — whenever EITHER its iter arg OR its corresponding op
+    // result is active. Using only iter arg activity misses the
+    // constant-accumulator case (constant init arg that produces an active
+    // result because the loop body accumulates active values into it).
+    // Using only result activity misses the case where an iter arg is active
+    // but its result is not (e.g. pointer-typed iter args used for address
+    // arithmetic whose final values are unused downstream).
+    // forceAugmentedReturns uses only iter arg activity, so for positions
+    // where the result is active but the iter arg is constant, the second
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
```diff
@@ -349,13 +349,31 @@ LogicalResult mlir::enzyme::detail::controlFlowForwardHandler(
...
+    // but its result is not (e.g. pointer-typed iter args used for address
+    // arithmetic whose final values are unused downstream).
+    // forceAugmentedReturns uses only iter arg activity, so for positions
+    // where the result is active but the iter arg is constant, the second
+    // overload inserts the missing shadow block arg after takeBody.
     for (auto &&[i, regionValue, operand] :
          llvm::enumerate(targetValues, operandRange)) {
-      if (gutils->isConstantValue(regionValue))
+      bool iterArgActive = !gutils->isConstantValue(regionValue);
+      bool resultActive = i < op->getNumResults() &&
```
wsmoses, line 366: this isn't generally correct since an input operand doesn't necessarily correspond to the outputs as here.

Instead really you need to construct this correctly [and separately] for scf for, or with an interface -- and pass said array to the general controlFlowForwardHandler which accepts arrays

enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
```diff
@@ -423,6 +441,47 @@ LogicalResult mlir::enzyme::detail::controlFlowForwardHandler(
...
+  // values across iterations), the first overload still adds that position to
+  // both operandPositionsToShadow and resultPositionsToShadow (union
+  // criterion), so replacement has the right number of results. However, the
+  // body block is missing the shadow block arg that the replacement's
+  // iter_arg slot expects. Insert it here, after takeBody has placed the
+  // cloned body into replacement.
+  //
+  // We also register the mapping in invertedPointers so that invertPointerM,
+  // which checks invertedPointers before isConstantValue, returns the shadow
+  // block arg instead of zero when body ops reference this iter arg.
```
wsmoses, line 456: this should be handled in createWithShadows, not here

vimarsh6739:
#2780 re-attempt

PR #2768, opened 2026-03-26
ProbProg: propagate strong zero to AD + position vec convention

sbrantq:
Custom logpdf path now takes a 1d position vector (per NumPyro convention) to avoid front end reshaping. This is the only path with user facing logpdf function atm. Internally MLIR probprog passes still use 2d tensor<1xNxf64> everywhere for position vec as 2d trace slices -- potentially to be refactored in future to use 1d position internally to eliminate more reshaping within an HMC/NUTS kernel

### DIFF
--- a/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
+++ b/enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
@@ -730,7 +730,7 @@ def MCMCOp : Enzyme_Op<"mcmc", [DeclareOpInterfaceMethods<SymbolUserOpInterface>
     Optional<AnyRankedTensor>:$initial_gradient,
     Optional<AnyRankedTensor>:$initial_potential_energy,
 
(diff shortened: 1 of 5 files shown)


enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -730,6 +730,7 @@ def MCMCOp : Enzyme_Op<"mcmc", [DeclareOpInterfaceMethods<SymbolUserOpInterface>
     Optional<AnyRankedTensor>:$initial_gradient,
     Optional<AnyRankedTensor>:$initial_potential_energy,
 
+    DefaultValuedAttr<BoolAttr, "false">:$strong_zero,
```
wsmoses: do we want to instead just have a DictionaryAttr:$autodiff_attrs that we then forward for use in the autodiff op creation?

sbrantq: Yeah makes a lot of sense

PR #2780, opened 2026-04-08
Add const shadows for forward mode AD on `RegionBranchOpInterface`

vimarsh6739:
This primarily affects scf.for and affine.for, but should also handle a the same case in scf.parallel and affine.parallel ops (iter-arg is a constant/dead inside loop but terminators still have activity). 

Aside: I think one interesting thing we can do here is to reuse the successor information to effectively eliminate the creation of some dead regions. This already kind of happens in scf.for which prunes the entry successors depending on if the loop has no iterations, or non-zero trip count(e.g. if the only entry successor is the parent, then just skip diffing the regionBranchOp body).

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
@@ -358,23 +358,42 @@ LogicalResult mlir::enzyme::detail::controlFlowForwardHandler(
     for (auto &&[i, regionValue, operand] :
          llvm::enumerate(targetValues, operandRange)) {
 
-      // check if all the predecessorValues are const too
-      SmallVector<Value> possibleActivePreds;
-      regionBranchOp.getPredecessorValues(successor, i, possibleActivePreds);
+      // check if we need to create a shadow for an inactive region value
+      if (gutils->isConstantValue(regionValue)) {
 
-      bool skipOpShadow = true;
-      for (auto pv : possibleActivePreds) {
+        // if all the possible predecessors for this value are also const, then
+        // we can skip creating a shadow. Else we need to create a shadow for
+        // syntactic correctness
+
+        SmallVector<Value> possibleActivePreds;
+        SmallVector<RegionBranchPoint> predecessors;
+        regionBranchOp.getPredecessors(successor, predecessors);
+        for (RegionBranchPoint predecessor : predecessors) {
+          if (predecessor.isParent()) {
+            // if the predecessor is the parent itself, then it's just
+            // operand!
+            possibleActivePreds.push_back(operand);
+            continue;
+          }
+          auto terminator = predecessor.getTerminatorPredecessorOrNull();
+          auto predecessorOperands = terminator.getSuccessorOperands(successor);
+          if (i < predecessorOperands.size())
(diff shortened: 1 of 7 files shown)


enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
```diff
@@ -352,8 +357,44 @@ LogicalResult mlir::enzyme::detail::controlFlowForwardHandler(
...
          llvm::enumerate(targetValues, operandRange)) {
-      if (gutils->isConstantValue(regionValue))
-        continue;
+
+      // check if we need to create a shadow for an inactive region value
+      if (gutils->isConstantValue(regionValue)) {
+
+        // if all the possible predecessors for this value are also const, then
+        // we can skip creating a shadow. Else we need to create a shadow for
+        // syntactic correctness
```
wsmoses: activity correctness

enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
```diff
@@ -397,24 +434,73 @@ LogicalResult mlir::enzyme::detail::controlFlowForwardHandler(
 
   SmallVector<Value> newOperands;
   newOperands.reserve(op->getNumOperands() + operandPositionsToShadow.size());
+
+  auto iface = dyn_cast<ControlFlowAutoDiffOpInterface>(op);
+  if (!iface) {
+    op->emitError() << " ControlFlowAutoDiffOpInterface not implemented for "
```
wsmoses: you can `return op->emitError...`

enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
```diff
@@ -397,24 +434,73 @@ LogicalResult mlir::enzyme::detail::controlFlowForwardHandler(
...
+            cast<AutoDiffTypeInterface>(Ty).getShadowType(gutils->width);
+        shadowValue = cast<AutoDiffTypeInterface>(shadowType)
+                          .createNullValue(builder, operand.get().getLoc());
+        hasConstOperandShadow = true;
+
+        // modify block arguments for entry successors to newOp, since
+        // forceAugmentedReturns will not shadow const operands. No need to add
+        // to the invertPointers map since `operand` is const, the shadow will
+        // be unused.
+        for (const RegionSuccessor &successor : entrySuccessors) {
```
wsmoses, line 457: I don't think this should be here, but within the impl of createWithShadows, right?

vimarsh6739: No, it has to be here. We are shadowing only the const operands for any augmented newOp.

vimarsh6739: createWithShadows correctly adds const shadows. But newOp doesn't have them(since it was created with forceAugmentedReturns), so the takeBody() discards the args in the replacement op

We have 2 choices here - either add the const shadows to newOp or the replacement (after takeBody). The end result is the same.

wsmoses (approved):

PR #2795, opened 2026-04-23
fix umax issue

### LINKED ISSUE #2794: crash at GradientUtils::invertPointerM
The issue is listed at https://tinyurl.com/26gyjubw

minansys:
# Enzyme Crash: `invertPointerM` assertion failure on `@llvm.umax.i32` with `-enzyme-loose-types`

## Summary
This is to fix #2794 

Enzyme crashes with `assert(0 && "cannot find deal with ptr that isnt arg")` in `GradientUtils::invertPointerM()` when differentiating code that uses the `@llvm.umax.i32` intrinsic in reverse mode with `-enzyme-loose-types=1` enabled.

## Root Cause

When a struct passed as `enzyme_dup` contains both floating-point and integer pointer fields, and the function stores the result of `@llvm.umax.i32` to an integer pointer loaded from that struct:

1. `visitCommonStore` handles the store of the `i32` umax result to the dup pointer
2. With loose types, Enzyme treats the stored `i32` value as potentially active
3. `visitCommonStore` calls `invertPointerM()` on the `i32` umax result to obtain its shadow
4. `invertPointerM()` cannot find or create a shadow for the `@llvm.umax.i32` call result
5. The function falls through to the assertion at `GradientUtils.cpp:6601`

The `@llvm.umax.i32` intrinsic arises from `SerialOperations::AtomicMax(dest, flag)`, which the LLVM optimizer lowers to a `load → @llvm.umax.i32 → store` pattern.

### LLVM IR reproducer (standalone)

`test/Enzyme/ReverseMode/loosetypes_umax.ll`

```
opt -load-pass-plugin=LLVMEnzyme-21.so \
    -passes="enzyme" -enzyme-preopt=false -enzyme-loose-types=1 \
    -S loosetypes_umax.ll
```

The IR models a struct with `float*` output data, `i32*` flags, and a `float` scalar. The function:
1. Loads and computes active float values
2. Derives an `i32` flag from a float comparison
3. Loads existing `i32` from the flags pointer, calls `@llvm.umax.i32`, stores result back

### Original failing code

Cromwell's `ComputeAlpha.cxx` boundary operation using Oak's `DualFixedNeighborMapIteration` reverse-mode template with `SerialOperations`.

The explicit template instantiation `flux_map_kernel_cuda_reverse<OperationBoundaryComputeAlpha, ...>` triggers the bug during x86 host compilation when the `__enzyme_autodiff` call inside the reverse kernel is processed with `-enzyme-loose-types=1`.

## Stack trace

```
GradientUtils::invertPointerM(...)
  GradientUtils.cpp:6601: Assertion `0 && "cannot find deal with ptr that isnt arg"' failed.
AdjointGenerator::visitCommonStore(...)
AdjointGenerator::visitStoreInst(...)
EnzymeLogic::CreatePrimalAndGradient(...)
```

## Proposed Fix

In `GradientUtils::invertPointerM()`, before the `CustomErrorHandler` fallback and the assertion, add a guard for integer/integer-vector values under loose type analysis:

```cpp
if (looseTypeAnalysis && oval->getType()->isIntOrIntVectorTy()) {
    auto *shadow = Constant::getNullValue(getShadowType(oval->getType()));
    invertedPointers.insert(
        std::make_pair((const Value *)oval, InvertedPointerVH(this, shadow)));
    return shadow;
}
```

This returns a zero shadow for integer values that cannot have meaningful floating-point derivatives, which is correct because:
- Under loose types, integer stores to dup pointers are processed by `visitCommonStore`
- The integer flag values (derived from float comparisons) have zero derivative
- A null constant shadow is the correct adjoint for non-differentiable integer operations

## Affected versions

- Enzyme v0.0.256 with LLVM 21
- Likely affects all Enzyme versions with `-enzyme-loose-types` support

## Test

The regression test `test/Enzyme/ReverseMode/loosetypes_umax.ll` guards against this issue.

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -6573,6 +6573,13 @@ end:;
     return Constant::getNullValue(getShadowType(oval->getType()));
   }
 
+  if (looseTypeAnalysis && oval->getType()->isIntOrIntVectorTy()) {
+    auto *shadow = Constant::getNullValue(getShadowType(oval->getType()));
+    invertedPointers.insert(
+        std::make_pair((const Value *)oval, InvertedPointerVH(this, shadow)));
+    return shadow;
+  }
+
   if (CustomErrorHandler) {
     std::string str;
     raw_string_ostream ss(str);
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -6334,9 +6334,8 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
 
       auto rule = [&](Value *val0, Value *val1) {
         Value *args[] = {val0, val1};
-        auto shadow = cast<CallInst>(
-            bb.CreateCall(II->getCalledFunction(), args,
-                          II->getName() + "'ipbi"));
+        auto shadow = cast<CallInst>(bb.CreateCall(
+            II->getCalledFunction(), args, II->getName() + "'ipbi"));
         shadow->setAttributes(II->getAttributes());
         shadow->setCallingConv(II->getCallingConv());
         shadow->setTailCallKind(II->getTailCallKind());
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -6323,6 +6323,34 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
     switch (II->getIntrinsicID()) {
     default:
       goto end;
+#if LLVM_VERSION_MAJOR >= 12
+    case Intrinsic::smax:
+    case Intrinsic::smin:
+    case Intrinsic::umax:
+    case Intrinsic::umin: {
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -6573,6 +6573,13 @@ end:;
     return Constant::getNullValue(getShadowType(oval->getType()));
   }
 
+  if (looseTypeAnalysis && oval->getType()->isIntOrIntVectorTy()) {
```
wsmoses: This is the wrong place for the fix imo, you should instead make a special binop case for umax (like there is for (inteegr) Add, etc)

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -6323,6 +6323,33 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
     switch (II->getIntrinsicID()) {
     default:
       goto end;
+#if LLVM_VERSION_MAJOR >= 12
```
wsmoses, line 6326: don't modify invertPointerM

instead do a special case handling for these intrinsics in visitIntrinsicInst, similar to how the binary operator handling is for some interger instructions like on AdjointGenerator.h:2614

PR #2798, opened 2026-04-29
Fix julia calling conv segfault

wsmoses:


wsmoses:
@copilot fix the error on llvm 21

PR #2801, opened 2026-04-30
Register reverse mode AD for memref.alloca_scope

### LINKED ISSUE #2791: memref.alloca_scope and its teminator memref.alloca_scope.return does not register reverse mode AD
The minimal reproducible example is 
```
func.func @scope_val(%x: f64) -> f64 {
  %r = memref.alloca_scope -> (f64) {
    %sq = arith.mulf %x, %x : f64
    memref.alloca_scope.return %sq : f64
  }
  return %r : f64
}

func.func @dscope_val(%x: f64, %dr: f64) -> f64 {
  %res = enzyme.autodiff @scope_val(%x, %dr)
    {
      activity=[#enzyme<activity enzyme_active>],
      ret_activity=[#enzyme<activity enzyme_activenoneed>]
    } : (f64, f64) -> f64
  return %res : f64
}
```

xys-syx:
Try to fix #2791

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/MemRefAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/MemRefAutoDiffOpInterfaceImpl.cpp
@@ -234,8 +234,8 @@ struct AllocaScopeOpInterfaceReverse
     Block &oldBody = scopeOp.getBodyRegion().front();
     bool valid = true;
 
-    // Values defined in the scoped region cannot be used outside it. Reset their
-    // adjoints before propagating gradients through the scoped body.
+    // Values defined in the scoped region cannot be used outside it. Reset
+    // their adjoints before propagating gradients through the scoped body.
     for (Operation &innerOp : oldBody.getOperations()) {
       for (Value result : innerOp.getResults()) {
         if (!gutils->isConstantValue(result)) {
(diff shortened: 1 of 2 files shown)


vimarsh6739:
I dont think this will work if the `alloca_scope` has a `memref.alloca` inside it (caching it will be unsafe across the 2 alloca_scope ops that are emitted)

Something like this will fail

```mlir
func.func @foo(%x : f64) -> f64{
    %out = memref.alloca_scope -> (f64) {
      %buf = memref.alloca() : memref<f64>
      memref.store %x, %buf[] : memref<f64>
      %y = memref.load %buf[] : memref<f64>
      memref.alloca_scope.return %y : f64
    }
    return %out
}

func.func @dfoo(%x, %dout) {
   %dx = enzyme.autodiff @foo(%x,%dout) {act=[active], ret_act=[activenoneed] } 
   return %dx
}
```
The primal is basically
```
out = alloca_scope{
buf[] = x
y = buf[]
return y
}
```
Reverse Mode should give us
```
dy += dout
d(buf) += dy // load
dx += d(buf) // store
d(buf) = 0 //alloca
```

We want this to lower to something like this (need to discuss what the caching should look like, since I dont think we can just push the memref to the cache here)

```
//forward
memref.alloca_scope {
    %buf = memref.alloca() 
    %dbuf = memref.alloca() 
    zero %dbuf
    
    enzyme.push %cache_store %dbuf
    memref.store %x, %buf
    
    enzyme.push %cache_load %dbuf
    %y = memref.load %buf
    
     alloca_scope.return %y
}

//reverse
memref.alloca_scope{
  //terminator
   dy += dout
   //load
   dbuf1 = enzyme.pop %cache_load
   old = memref.load dbuf1
   memref.store  (old + dy), dbuf1 
   //store
   dbuf2 = enzyme.pop %cache_store
   %old2 = memref.load dbuf2
   dx += old2
   
   memref.store 0, dbuf2
}
```

The issue is that the cache push/pop will not be valid across the 2 alloca_scopes here

vimarsh6739:
#2814

PR #2802, opened 2026-05-04
LLVM integrate

wsmoses:

PR #2803, opened 2026-05-05
update rust sha in CI

sgasho:
related: https://github.com/EnzymeAD/Enzyme/pull/2802#issuecomment-4374551075

new sha is from 2026/05/04


ZuseZ4:
(@sgasho, please feel free to ping wsmoses directly in the future)
I still don't have my permissions back to run ci, so ping @wsmoses

PR #2807, opened 2026-05-07
Create Fortran bindings module

### LINKED ISSUE #2805: Create Fortran module for bindings
@isaacaka and I now have some Fortran code that we can contribute back - in particular, bindings for the activity descriptors. We'll:

- [x] Add a Fortran module into the repo containing these bindings
- [x] Hook it up in the build system

joewallwork:
Closes #2805

Thanks to @isaacaka for figuring out how to make use of activity descriptors in Fortran: bind to those that Enzyme picks up in C. To avoid users having to know how to do Fortran-C binding, we thought it would be beneficial to create a Fortran module containing these for convenience. There will likely be other things that come up that we can add to the module later in the project.

To use these descriptors in Fortran, import them with
```fortran
use enzyme_mod, only: enzyme_const, enzyme_dup
```
or just
```fortran
use enzyme_mod
```

## Checklist

- [x] Add a Fortran module into the repo containing these bindings
- [x] Hook it up in the build system

### DIFF
--- a/enzyme/CMakeLists.txt
+++ b/enzyme/CMakeLists.txt
@@ -34,9 +34,11 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 option(ENZYME_ENABLE_PLUGINS "Enable Clang/LLD/Opt plugins" ON)
 option(ENZYME_ENABLE_BENCHMARKS "Enable benchmarks" OFF)
 option(ENZYME_BC_LOADER "Enable bitcode loader" ON)
+option(ENZYME_FORTRAN "Build Enzyme Fortran bindings" OFF)
 option(ENZYME_CLANG "Build enzyme clang plugin" ON)
 option(ENZYME_FLANG "Build enzyme flang symlink" OFF)
 option(ENZYME_MLIR "Build enzyme mlir plugin" OFF)
+option(ENZYME_IFX "Enable enzyme support for the Intel Fortran compiler IFX" OFF)
 option(ENZYME_IFX "Enable enzyme support for the Intel Fortran compiler IFX" OFF)
 option(ENZYME_EXTERNAL_SHARED_LIB "Build external shared library" OFF)
 option(ENZYME_APPLE_DYNAMIC_LOOKUP
@@ -299,7 +301,9 @@ endif()
 if (ENZYME_ENABLE_PLUGINS)
     add_subdirectory(test)
 endif()
-add_subdirectory(Fortran)
+if (ENZYME_FORTRAN)
+    add_subdirectory(Fortran)
+endif()
 
 # The benchmarks data are not in git-exported source archives to minimize size.
 # Only add the benchmarks if the directory exists.
--- a/enzyme/CMakeLists.txt
+++ b/enzyme/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.13)
-project(Enzyme LANGUAGES C CXX Fortran)
+project(Enzyme)
 
 include(CMakePackageConfigHelpers)
 include(CheckIncludeFile)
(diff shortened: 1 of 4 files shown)


vchuravy:
Is `use enzyme_mod` common in Fortran? Would it make more sense to just have `use enzyme`?

enzyme/Fortran/enzyme.f90
```diff
@@ -0,0 +1,36 @@
...
+  private
+
+  ! Bindings for activity descriptors
+  integer(c_int), public, bind(C, name="enzyme_const")     :: enzyme_const
+  integer(c_int), public, bind(C, name="enzyme_dup")       :: enzyme_dup
+  integer(c_int), public, bind(C, name="enzyme_dupnoneed") :: enzyme_dupnoneed
+  integer(c_int), public, bind(C, name="enzyme_out")       :: enzyme_out
+  integer(c_int), public, bind(C, name="enzyme_scalar")    :: enzyme_scalar
+  integer(c_int), public, bind(C, name="enzyme_width")     :: enzyme_width
+  integer(c_int), public, bind(C, name="enzyme_vector")    :: enzyme_vector
```
vchuravy, line 35: Should we also add `__enzyme_autodiff` and `__enzyme_fwddiff`

joewallwork:
> Is `use enzyme_mod` common in Fortran? Would it make more sense to just have `use enzyme`?

Both are common although I suppose `_mod` tends to be appended in older codes. I'm happy to `use enzyme` instead.

enzyme/Fortran/enzyme.f90
joewallwork: ~~Oh good point, we could provide interfaces using `iso_c_binding`. I'll look into it. Note that you can also declare externals as in #2809. We can go with whichever ends up working best.~~

Edited: Never mind, I originally tried setting up explicit interfaces but it was a non-starter because of all the possible signatures. The implicit approach using `external` is better so let's defer this for #2809.

vchuravy: I think it would make a lot of sense to have all of these in one module.

joewallwork: I suppose we could just define them as externals here. I'll do that.

joewallwork:
> > Is `use enzyme_mod` common in Fortran? Would it make more sense to just have `use enzyme`?
> 
> Both are common although I suppose `_mod` tends to be appended in older codes. I'm happy to `use enzyme` instead.

Done in 27b7edc9658da36fcfc8881a97b73aa68d4b986a.

enzyme/Fortran/enzyme.f90
joewallwork: Changed #2809 to target the branch associated with this PR to keep the tasks separate.

vchuravy (approved):

enzyme/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.13)
-project(Enzyme)
+project(Enzyme LANGUAGES C CXX Fortran)
```
wsmoses: what does this do?

joewallwork: Without adding Fortran to the languages in the project, I get the CMake error
```
FortranCInterface requires the Fortran language to be enabled
```
when building Enzyme. If I remove that from `Fortran/CMakeLists.txt` then I get
```
CMake can not determine linker language for target: EnzymeFortran
```

joewallwork: Fixed in 0479c37962c5d24c379213ea820be0f79f38efab by enabling Fortran in the subdir rather than globally.

joewallwork: Also added a CMake option `ENZYME_FORTRAN` in 827e439a404abcca9508feb2fd80e8ad9a89e864 that's defaulted to off like the other Fortran-related options. Turning it on builds the bindings.

wsmoses (approved):

PR #2809, opened 2026-05-07
Add implicit interfaces for function hooks in Fortran module

joewallwork:
Closes #2808.
~~Merges into #2807.~~

As mentioned in the issue, it turns out that the interfacing to the special Enzyme function hooks can be done using implicit interfaces in Fortran. That is, you can get away without providing an explicit `interface` block like is done in the examples so long as you declare
```fortran
external :: __enzyme_autodiff
```
for example.

## Checklist

- [x] Add to Fortran `enzyme` module
- [x] Use in `square` example

Hooking up in the `allocatableArraySimple` example will require fixing #2820.

PR #2821, opened 2026-05-11
Account for Fortran assumed shape

### LINKED ISSUE #2820: Error applying Enzyme to `alloctableArraySimple` Fortran example
I tried compiling https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/test/Fortran/ReverseMode/allocatableArraySimple.f90 locally with Flang. I got Flang to emit LLVM but I got an error when applying Enzyme. It appears that Enzyme doesn't handle the assumed shape array properly. It surprised me that this didn't work given that the test is included in the Enzyme repo.

joewallwork:
Closes #2820
Merges into #2837

As documented in the issue, I wasn't able to build the `allocatableArraySimple` example locally. It seemed that Enzyme didn't fully account for assumed shape arrays in Fortran. The small patch in this PR addresses this and I can confirm I can now forward and reverse mode differentiate through things like
```fortran
  subroutine normalise(x, y)
    real, dimension(:), intent(in) :: x
    real, dimension(:), intent(out) :: y
    y(:) = x / sum(x)
  end subroutine normalise
```
which is really nice.

If I understand correctly, the error message I was hitting was slightly misleading so I propose an alternative phrasing.

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -2642,7 +2642,7 @@ public:
              << " val:" << gutils->isConstantValue(&I)
              << " type: " << TR.query(&I).str() << "\n";
         }
-      ss << "cannot handle unknown binary operator: " << BO << "\n";
+      ss << "derivative of active binary operator not known: " << BO << "\n";
       EmitNoDerivativeError(ss.str(), BO, gutils, Builder2);
     }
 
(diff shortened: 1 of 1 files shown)


joewallwork:
This all works well on it's own. However, I've just discovered that it doesn't work in conjunction with #2809. I tested this branch merged with that one on cases with assumed shape arrays and they compile but produce segfaults. This branch also produces a warning like
```
warning: binary operator is integer and assumed constant:   %19 = mul nsw i64 %11, %15
```
for each assumed shape array encountered, which isn't ideal. For a full fix, we should address #2822.

joewallwork:
Closing because we should address properly according to #2822.

PR #2824, opened 2026-05-12
Add external context accessor to type analysis

vchuravy:
To remove the last uses of `enzyme_extract_world` in https://github.com/EnzymeAD/Enzyme.jl/pull/2643

### DIFF
--- a/enzyme/Enzyme/CApi.cpp
+++ b/enzyme/Enzyme/CApi.cpp
@@ -947,6 +947,11 @@ const char *EnzymeTypeAnalyzerToString(void *src) {
   return cstr;
 }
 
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/CApi.cpp
```diff
@@ -947,6 +947,11 @@ const char *EnzymeTypeAnalyzerToString(void *src) {
   return cstr;
 }
 
+void *EnzymeTypeAnalyzerGetExternalContext(void *src) {
```
wsmoses: what if instead we did a TypeAnalyzerGetLogic ?

PR #2825, opened 2026-05-14
Add BLAS triangular solve derivatives

jlperla:
## Summary

Adds native BLAS/LAPACK derivative support for real triangular solve routines:

- `trsv`
- `trsm`
- `potrs`

This covers forward and reverse mode for real `s/d` routines, with tests for Fortran and selected CBLAS entry points.

## Implementation Notes

The `trsv` rule follows the existing `trtrs` shape with a single RHS/vector solve. Reverse mode needs to solve the `x` cotangent before forming the `A` cotangent, so the generated reverse rewrite order is extended for `trsv` like `trtrs`.

The `trsm` rule adds active `A`/`B` support with inactive `alpha`. It also adds `trsm` to BLAS extraction. Supporting both left/right side solves and CBLAS layout required a few small tablegen extensions rather than hand-written derivative code.

The `potrs` rule handles differentiation through the Cholesky solve using triangular BLAS operations. This does not add pivot support and does not address LU `getrf`/`getrs`.

## Generator/Infrastructure Changes

Most changes are local extensions to the existing BLAS tablegen machinery:

- Generalize hidden/by-ref character handling from only `trans` to all BLAS char arguments such as `uplo`, `diag`, and `side`.
- Add `mat_ld` support for temporary leading dimensions that depend on CBLAS row-major vs column-major layout.
- Add a `side_square` temporary shape for `trsm`, where the triangular factor dimension depends on `side`.
- Add a `Dep` DAG helper so generated operands can express data/cache dependencies while emitting the intended BLAS argument.
- Add a layout-aware matrix accumulation helper, analogous to the existing differential matrix memcpy helper, for row-major/column-major shadow accumulation.

## Tests

Adds focused lit tests for:

- `dtrsv_64_`
- `dtrsm_64_`
- `cblas_dtrsm64_`
- `dpotrs_64_`

Extends BLAS integration tracing and integration tests for forward and reverse triangular solve coverage.

I also validated the Julia integration locally by building Enzyme.jl against this branch with `deps/build_local.jl`, temporarily disabling fallback for `trsv`, `trsm`, and `potrs`, and running:

- `julia --project=test test/runtests.jl --verbose blas`
- `julia --project=test test/runtests.jl --verbose rules/internal_rules/linear_algebra_rules`
- downstream triangular BLAS MWE

The Julia checks passed without fallback warnings for `BLAS.trsv!` and `BLAS.trsm!`. LU `getrf`/`getrs` and high-level `A \ b` remain separate follow-up work.

### DIFF
--- a/enzyme/Enzyme/BlasDerivatives.td
+++ b/enzyme/Enzyme/BlasDerivatives.td
@@ -56,6 +56,7 @@ def Rows : MagicInst; // given a transpose, normal rows, normal cols get the tru
 def Concat : MagicInst;
 
 def ShadowNoInc : MagicInst;
+def Dep : MagicInst;
 
 class Binop<string _s, list<string> _tys> {
   string s = _s;
@@ -89,8 +90,13 @@ class IntMatchers<string _inty, string _outty, list<string> _before, list<string
 
 def side_to_trans : IntMatchers<
     "charType", "charType",
-    ["'l'", "'L'", "'R'", "'r'"],
-    ["'n'", "'N'", "'T'", "'t'"]
+    ["'l'", "'L'", "141", "0", "'R'", "'r'", "142", "1"],
+    ["'n'", "'N'", "'N'", "'N'", "'T'", "'t'", "'T'", "'T'"]
+    >;
+def side_to_rtrans : IntMatchers<
+    "charType", "charType",
+    ["'l'", "'L'", "141", "0", "'R'", "'r'", "142", "1"],
+    ["'t'", "'T'", "'T'", "'T'", "'N'", "'n'", "'N'", "'N'"]
     >;
 
 def is_diag_int : IntMatchers<
@@ -107,6 +113,7 @@ def is_nonunit : MagicInst;
 def First : MagicInst;
 def Lookup : MagicInst;
 def LoadLookup : MagicInst;
+def mat_ld : MagicInst;
 
 class FAdd<string _tmp=""> {
   string unused = _tmp;
@@ -160,6 +167,10 @@ class CopyLowerToUpper<string _tmp> {
 }
 
 class MemcpyMatAdd<bit _shouldZero> {
+  bit shouldZero = _shouldZero;
+}
+
+class MatAdd<bit _shouldZero> {
   bit shouldZero = _shouldZero;
 }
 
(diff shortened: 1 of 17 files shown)


jlperla:
This is intended as the first step in solving https://github.com/EnzymeAD/Enzyme.jl/issues/3039

Full disclosure: This is heavily LLM generated because I do not understand the LLVM/etc. side.  Beyond my capability, but given the comment from @vchuravy  it seemed that this approach was the only feasible one to get support.

I tried to test the usage with downstream Enzyme.jl by flipping a fallback flag, etc. but I might have made a mistake.

enzyme/Enzyme/Utils.cpp
```diff
@@ -2068,6 +2068,129 @@ Function *getOrInsertDifferentialFloatMemcpyMat(
   return F;
 }
 
+Function *getOrInsertDifferentialFloatMemcpyMatLayout(
```
wsmoses: what is this, why is it needed?

enzyme/Enzyme/BlasDerivatives.td
```diff
@@ -718,33 +768,78 @@ def trsm: CallBlasPattern<(Op $layout, $side, $uplo, $transa, $diag, $m, $n, $al
...
+                    (MatAdd<false> $layout, (ISelect (is_nonunit $diag), ConstantInt<0>, $m), $n, (Shadow $B), (Concat (Dep (Shadow $A), $B), $ldb)),
+                    (BlasCall<"trsm"> $layout, $side, $uplo, $transa, $diag, $m, $n, Constant<"1.0">, $A, (ld $A, (side_to_trans $side), $lda, $n, $m), (Shadow $B))
+                  )
                   >;
 
 def uplo_to_normal : IntMatchers<
     "charType", "charType",
-    ["'l'", "'L'", "'U'", "'u'"],
-    ["'n'", "'N'", "'T'", "'t'"]
+    ["'l'", "'L'", "122", "0", "'U'", "'u'", "121", "1"],
```
wsmoses: this shouldn't be needed

jlperla:
Thanks @wsmoses   I was about to ping you.  This is a test to see if I can get an LLM to generate the C++ rules (impenetrable for someone like me) to get the triangular and (and eventually LU stuff) up to date.  I want to push out some communications on Enzyme to the economics community but right now it is failing on some basic examples.

If what the LLM generated is complete gibberish then I should just close this entirely.  If it isn't then keep putting comments and I will tell the AI agent to implement anything you suggest and push again.

enzyme/Enzyme/Utils.cpp
jlperla: Removed. This was a separate layout-aware helper, but it was unnecessary. I folded the layout handling into the existing `getOrInsertDifferentialFloatMemcpyMat` helper instead, so triangular and full-matrix accumulation both go through the same path.

enzyme/Enzyme/BlasDerivatives.td
jlperla: Agreed, removed. The numeric CBLAS enum cases should not be in these tablegen matchers. The generated code now handles the CBLAS layout separately, and these matchers only cover the actual character `U`/`L` cases.

jlperla:
Updated this to reduce the helper surface and address the review comments.

Main changes:
- Removed the separate layout-aware differential matrix memcpy helper.
- Folded row-major/column-major addressing into the existing `getOrInsertDifferentialFloatMemcpyMat`.
- Removed the numeric CBLAS enum cases from the `uplo_to_*` tablegen matchers.
- Adjusted the generated BLAS rule code so CBLAS layout is handled explicitly before calling the matrix accumulation helper.
- Regenerated/updated the affected lit checks.

Validation:
- `ninja -C build-llvm16 BlasDerivativesIncGen BlasDeclarationsIncGen BlasTAIncGen BlasDiffUseIncGen`
- `ninja -C build-llvm16 LLVMEnzyme-16`
- `lit -sv build-llvm16/test/Enzyme/ForwardMode/blas`
- `lit -sv build-llvm16/test/Enzyme/ReverseMode/blas`
- `lit -sv build-llvm16/test/Integration/ForwardMode/blas.cpp`
- `lit -sv build-llvm16/test/Integration/ReverseMode/blas.cpp`
- `ninja -C build-llvm16 check-enzyme`
- local Enzyme.jl build against this branch with fallback temporarily disabled for `trsv`, `trsm`, and `potrs`
- `julia --project=test test/runtests.jl --verbose blas`
- `julia --project=test test/runtests.jl --verbose rules/internal_rules/linear_algebra_rules`
- downstream triangular BLAS MWE

jlperla:
@wsmoses   OK, take two with the AI.  It changed the code to simplify given your feedback.

I feel like this is worth at most one more iteration.  If this is nonsense and a waste of your time just let me know and I will kill it.  If you think the AI is getting pretty close to the right design then we can continue the experiment.

wsmoses (commented):
I don't understand why we would need any changes to enzyme/Enzyme or tools/enzyme-tblgen other than adding `def trsv`, or adding the forward mode rules to things like potrf/etc.

Concurrently, it would be a lot easier to separate these into individual prs for each rule added.

jlperla:
@wsmoses Splitting this up per your suggestion. First of the split is #2828, which is just `def trsv` (forward + reverse), with the minimum-viable tablegen surface to support it — no Utils.cpp/Utils.h changes, no other BLAS rule changes. Subsequent PRs will follow for `trsm`, `potrs`, and forward-mode rules for `potrf`/etc.

PR #2837, opened 2026-05-21
Set up Flang CI

### LINKED ISSUE #2832: Set up Flang in the CI
Currently, the Fortran CI only considers ifx. We should at least consider flang, too.

joewallwork:
Closes #2832
~~Merges into #2809~~ 

Adds Flang to the Fortran CI workflow and separates out versions of the tests that work under Flang vs IFX.


joewallwork:
Okay this is now at the stage where it needs #2821.

.github/workflows/bcload.yml.txt
vchuravy, line 1: Lots of unrelated changes?

joewallwork: I temporarily turned off other CI workflows - see the "DO NOT MERGE" commits.

joewallwork:
Okay, I've got something working!

This PR keeps the ifx setup that uses llvm version 15 and adds another setup with a more recent flang version.

With the following caveats, this extends the test suite to provide some coverage for the `enzyme_autodiff` driver from the Fortran bindings introduced in #2809.

## Caveats

* The `square` test is skipped under ifx because that test makes use of the `enzyme_autodiff` driver from our Fortran bindings, which we haven't managed to get it working with ifx yet. (See #2809.)
* The `allocatedArraySimple` tests fail under flang. However, #2821 (that merges into this PR) fixes those.

enzyme/test/Fortran/ForwardMode/allocatableArraySimple_with_bindings.f90
```diff
@@ -0,0 +1,58 @@
+! RUN: if [[ %llvmver -ge 13 && %fc != ifx ]]; then %fc -flto -O0 -c %loadFortran %s -o /dev/stdout | %opt %loadEnzyme %enzyme -o /dev/stdout | %opt -O0 -S -o %t.ll && %fc -flto -O0 %t.ll -o %t1 && %t1 | FileCheck %s; fi
```
vchuravy: Would be nice to simplify this logic a bit, by putting it partially into lit.cfg

joewallwork: Could you give a hint of which bits you'd like simplifying?

joewallwork: Avoid being `%fc` specific - e.g., `REQUIRES` flag. Or put IFX/Flang configs into lit.

joewallwork: Done in cbeb30b0fdcdb55278de9225c7bd806d317a7a20 and 9cda6a5c65f168ef749426aefc34d270d421cd08. Unfortunately, it doesn't seem to be possible to have a per-run `REQUIRES` statement so we'll have to keep the optimisation level check in for a few tests for now.

vchuravy (approved):

Issue #2834, opened 2026-05-19
Enzyme  generates undef-fed tape PHIs for packed-state nested guard reproducer


minansys:
https://tinyurl.com/2c2v5q9u

## Summary

On current Enzyme `main` (`d0749cbb`, `Attempt to fix rematerialization ficticious phi issues (#2827)`), the attached LLVM IR reproducer still generates `undef`-fed tape state in `augmented_eval`.

This looks related to the old nested-guard family from #2629, but not like the exact same control-flow bug. In this reproducer, reverse still remains gated by the saved outer state, but Enzyme still packs `undef` into saved `i1`, `ptr`, and `double` values along inactive paths.

So this appears to be a remaining tape / PHI / rematerialization issue rather than a duplicate of the specific nested-predicate control-flow bug fixed for #2629.

## Attached reproducer

The reproducer is the attached LLVM IR file:

- `packed_state_outer_guard_undef_tape.ll`

It is self-contained and does not depend on any C++ source or Cromwell code.

The reduced shape is:

- packed state with booleans and an optional pointer
- outer guard around a helper call
- nested helper branches inside that helper
- a looped dot-product path that Enzyme tapes as `{ ptr, double }`
- a later unrelated optional-pointer join

## Reproduction

The file already contains `RUN:` lines for the Enzyme test suite.

Equivalent local reproduction command:

```bash
opt < packed_state_outer_guard_undef_tape.ll \
	-load-pass-plugin=/path/to/LLVMEnzyme.so \
	-passes="enzyme,function(sroa,mem2reg,early-cse,simplifycfg,instsimplify,correlated-propagation,simplifycfg,adce)" \
	-S -enzyme-preopt=0 -enzyme-detect-readthrow=0 | \
	FileCheck packed_state_outer_guard_undef_tape.ll
```

## Problematic transformed shape

In the transformed IR, Enzyme emits:

```llvm
define internal { { ptr, ptr, { double, double, double }, ptr, i1, i1, i1, double, double, ptr, i1, double }, double } @augmented_eval(...)
```

and the scalarized output still contains `undef`-fed tape fields such as:

```llvm
%.sroa.14.0 = phi i1 [ %has_rotation, %with.frame ], [ undef, %entry ]
%.sroa.16.0 = phi i1 [ %has_translation, %with.frame ], [ undef, %entry ]
%.sroa.10.0 = phi ptr [ %subcache, %with.dot ], [ undef, %after.frame ]
[...]

PR #2839, opened 2026-05-22
Guard reverse-mode Store adjoint against NULL shadow pointers; preser…

minansys:
#2834 

Two related crashes / silent-miscompilations surfaced when with packed-state structs, an outer "is-this-frame-moving?" guard, and inactive frame-velocity fields whose shadow pointers can legitimately be NULL on some control-flow paths.

1) Enzyme/AdjointGenerator.h -- visitCommonStore reverse path
   Reverse-mode Store adjoint was unconditionally doing
       dif1 = load(shadow_ptr);
       setPtrDiffe(shadow_ptr, 0);
       addToDiffe(orig_val, dif1, ...);
   on the shadow pointer, with no NULL check. When the upstream code
   produced an undef/null shadow for an inactive packed-state field that the
   outer guard had previously protected, the reverse pass dereferenced it
   and SIGSEGV'd inside the load-zero-fadd emitted for the gradient
   accumulation. The fix wraps that triple in a runtime `if (shadow_ptr !=
   null)` guard, splitting the reverse BB into <bb>_nnactive / <bb>_nnmerge
   and letting the existing merge-block plumbing close the diamond.

2) Enzyme/GradientUtils.cpp -- SubTransferHelper srcConstant branch
   When the memcpy/memset source is a constant (mode != ForwardModeSplit),
   upstream lowered the shadow-side transfer to a `memset(shadow_dst, 0,
   len)`. That is wrong for shadow allocas that visitCommonStore has
   already populated with valid scratch-pointer entries: the memset
   silently zeros those pointers and downstream loads read NULL, leading
   to either gradient drops or the crash above.

   The fix replaces the srcConstant memset with an unconditional memcpy
   from the *primal* source into the shadow destination, with the proper
   inttoptr casts, optional GEP-by-offset, and MaybeAlign on both
   operands. This preserves any scratch-pointer payload that the reverse
   pass needs.

   The dead helper `ShadowDstHasPointerLoadInRange` and a leftover
   `if (true || ...)` gate from an earlier defense-in-depth attempt are
   also removed in this commit; both are subsumed by the unconditional
   memcpy.

Regression test
   test/Enzyme/ReverseMode/packed_state_outer_guard_undef_tape.ll
   reproduces the original crash class: packed-state struct, outer frame-
   motion guard, optional-pointer join leaving tape fields path-dependent.
   Before the fix the IR FileCheck fails (or opt crashes); after the fix
   it produces well-formed reverse-mode IR with the NULL-shadow diamond.

Validation
   - Stand-alone .ll repro: opt -enzyme exits cleanly, FileCheck passes.
   - End-to-end CFD case (Cromwell wall-function adjoint, runcrmadcpu -dp): previously SIGSEGV'd inside the visitStoreInst reverse path; now exits 0 and emits the full reverse-mode hash output. HashDump comparison confirms the first 146271 lines are byte-identical to a prior-known-good prefix and 67262 additional rule blocks are produced past the v6 truncation point -- i.e. no silent gradient regression.


minansys:
[#2872
](https://github.com/EnzymeAD/Enzyme/pull/2872) is a better fix

PR #2841, opened 2026-05-28
Add reverse AD for llvm.intr.memcpy

xys-syx:
llvm.intr.memcpy has no registry in reverse AD, which cause "could not compute the adjoint for this operation "llvm.intr.memcpy"". 

Reverse rule used here:
```
  d_src[i] += d_dst[i]
  d_dst[i] = 0
```


enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
```diff
@@ -297,6 +299,109 @@ struct InsertValueOpInterfaceReverse
...
+    Value n =
+        arith::IndexCastOp::create(builder, loc, builder.getIndexType(), nInt);
+
+    Value c0 = arith::ConstantIndexOp::create(builder, loc, 0);
+    Value c1 = arith::ConstantIndexOp::create(builder, loc, 1);
+    Value zeroElem = adt.createNullValue(builder, loc);
+    Type ptrTy = cp.getDst().getType();
+
+    auto forOp = scf::ForOp::create(builder, loc, c0, n, c1);
+    OpBuilder body(forOp.getBody()->getTerminator());
```
Pangoraw: better to use the provided `builder` with setInsertionPoint as it can have hooks

Pangoraw (commented):
the rule looks good, we will need to include scftocf and cftollvm lowering passes in pipelines.

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
xys-syx: @Pangoraw thanks, I have modified

> we will need to include scftocf and cftollvm lowering passes in pipelines

we have covert-polygeist-to-llvm in the downstream, I guess it may could work for it
https://github.com/EnzymeAD/Enzyme-JAX/blob/5db2b680308c38613ed6d5d01ed7d9d9a1353a65/src/enzyme_ad/jax/Passes/ConvertPolygeistToLLVM.cpp#L4573

Pangoraw: ah yes perfect in this case

enzyme/Enzyme/MLIR/Implementations/LLVMAutoDiffOpInterfaceImpl.cpp
```diff
@@ -297,6 +299,111 @@ struct InsertValueOpInterfaceReverse
...
+          if (isa<AutoDiffTypeInterface>(st.getValue().getType()))
+            return st.getValue().getType();
+      }
+      return nullptr;
+    };
+    if (Type t = walk(cp.getDst()))
+      return t;
+    if (Type t = walk(cp.getSrc()))
+      return t;
+    return Float64Type::get(cp.getContext());
```
wsmoses: this is wrong, we should throw an error if things can't be deduced. This is also somewhat hacky and we should use type analysis

xys-syx: ok, I will fix it

xys-syx: @wsmoses  I have fixed it, but only for the case where the inferred element type is a scalar int or float and using upstream MLIR for inferring types. The memcpy ops in RSBench that copy arrays of `struct<(f64, f64)>` and `array<10 x f64>` need to handle it recursively, I plan to open a new PR for it.

vimarsh6739:
@codex review

PR #2846, opened 2026-06-01
mlir: run mincut for scf.if ops

pengmai:


wsmoses (approved):
lgtm, but can we also do this for affineif as well?

Pangoraw (approved):
lgtm! we could potentially run mincut for every pairs of forward / reverse blocks as present in gutils->mapReverseModeBlocks (though nested ops don't populate the map rn)

wsmoses:
@pengmai we have a conflict here rn if you can take a quick look [then we can merge, and address affine.if in follow ups]

pengmai:
> @pengmai we have a conflict here rn if you can take a quick look [then we can merge, and address affine.if in follow ups]

Done. FYI it looks like we don't currently support differentiation of affine.if at all, but these changes are templated in an `IfLike` remover that should work with affine.if without modification when support for affine.if is added.

> we could potentially run mincut for every pairs of forward / reverse blocks as present in gutils->mapReverseModeBlocks (though nested ops don't populate the map rn)

This would be interesting. We'd need to make sure it respects the `preserve_cache` attributes.

PR #2848, opened 2026-06-01
mlir: support active pointers in scf.if

pengmai:


wsmoses (approved):

wsmoses:
@pengmai we should also consider affine if here in a follow up too

PR #2852, opened 2026-06-05
Initialize inactive reverse tape values

### LINKED ISSUE #2834: Enzyme  generates undef-fed tape PHIs for packed-state nested guard reproducer

minansys:
Fixes #2834 and #2629.\n\nRoot cause:\n- #2834: augmented reverse tapes were assembled by sparse stores into the returned tape aggregate. When a guarding branch was not taken, SROA exposed the unstored fields as undef-fed PHIs, then those values were packed into the tape. The tape backing slot now gets a defined zero/null baseline before path-local tape stores.\n- #2629: manual LCSSA values synthesized for blocks where the primal instruction is inactive used undef. Enzyme postopt could propagate that through noundef math and infer an invalid guard assumption. Those inactive LCSSA entries now use the type's safe zero/null value.\n\nTests:\n- llvm-lit -v /home/minxu/code/enzyme/build/enzyme2-llvmorg-19.1.7/test/Enzyme/ReverseMode/inactive-tape-values.ll\n- llvm-lit -v /home/minxu/code/enzyme/build/enzyme2-llvmorg-19.1.7/test/Integration/ReverseMode/nested_guard_inactive_lcssa.c\n- manual #2834 opt pipeline with LLVM 19.1.7: no remaining i1/ptr/double PHIs with undef in the generated output\n- manual #2629 direct Clang plugin repros at -O0/-O1 for minimized C and original C++ repro: gradients are {1} and {1,1,1}\n

### DIFF
--- a/enzyme/Enzyme/EnzymeLogic.cpp
+++ b/enzyme/Enzyme/EnzymeLogic.cpp
@@ -2959,10 +2959,8 @@ const AugmentedReturn &EnzymeLogic::CreateAugmentedPrimal(
         tapeMemory = ib.CreateGEP(RetType, ret, Idxs, "");
         cast<GetElementPtrInst>(tapeMemory)->setIsInBounds(true);
       }
-      if (EnzymeZeroCache) {
-        ZeroMemory(ib, tapeType, tapeMemory,
-                   /*isTape*/ true);
-      }
+      ZeroMemory(ib, tapeType, tapeMemory,
+                 /*isTape*/ true);
     }
 
     unsigned i = 0;
--- a/enzyme/Enzyme/EnzymeLogic.cpp
+++ b/enzyme/Enzyme/EnzymeLogic.cpp
@@ -2912,7 +2912,8 @@ const AugmentedReturn &EnzymeLogic::CreateAugmentedPrimal(
   AllocaInst *ret = noReturn ? nullptr : ib.CreateAlloca(RetType);
 
   if (!noTape) {
-    Value *tapeMemory;
+    Value *tapeMemory = nullptr;
+    Instruction *directTapeInitInsert = nullptr;
     if (recursive && !omp) {
       auto i64 = Type::getInt64Ty(NewF->getContext());
       auto size =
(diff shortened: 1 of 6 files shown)


minansys:
This did not resolve the issue [#2872](https://github.com/EnzymeAD/Enzyme/pull/2872) fix it better

PR #2855, opened 2026-06-06
fix: add x==0 y>0 guard to pow derivatives

martinjrobins:
Enzyme uses this forward-mode directional derivative for `pow(x, y)`

`x' * y * pow(x, y - 1) + y' * pow(x, y) * log(x)`

at `x == 0`, the `log(x)=inf`, therefore the second term would be 0 * inf = NaN

I believe the result should be:

```
pow(0, 1)  = 0.0
d/dx       = 1.0
d/dy       = 0.0
```

in general this is true for `x == 0` and `y > 0`

So I think there should be a guard to test for `x == 0` and `y > 0` and set the second term to 0 in the fwd and rev derivatives.

This PR implements the guard and updates the foward and reverse tests for pow. I've added a couple of integration tests as well to check the numerical outputs.

I checked this against JAX, so I'll put this script below for reference, this gives the output copied above

```python
import jax
import jax.numpy as jnp

def f(x, y):
    return jnp.power(x, y)

grad_f = jax.grad(f, argnums=(0, 1))

x, y = 0.0, 1.0
dx, dy = grad_f(jnp.float32(x), jnp.float32(y))

assert float(dx) == 1.0, f"d/dx pow(0,1) should be 1.0, got {float(dx)}"
assert float(dy) == 0.0, f"d/dy pow(0,1) should be 0.0, got {float(dy)}"

print(f"pow(0, 1)  = {f(x, y):.1f}")
print(f"d/dx       = {dx}")
print(f"d/dy       = {dy}")
```


enzyme/Enzyme/InstructionDerivatives.td
```diff
@@ -1051,8 +1051,14 @@ def : IntrPattern<(Op $x, $y),
                   [["pow"]],
                   [
                     (CheckedMul (DiffeRet), (FMul $y, (Call<(SameFunc)> $x, (FSub $y, (ConstantFP<"1"> $y))))),
-                    (CheckedMul (DiffeRet), (FMul (Call<(SameFunc)> $x, $y), (Intrinsic<"log"> $x)
-                    ))
+                    (CheckedMul (DiffeRet),
+                      (Select
+                        (And (FCmpOEQ $x, (ConstantFP<"0"> $x)),
+                             (FCmpOLT (ConstantFP<"0"> $y), $y)),
```
wsmoses: why do we need the 0 < y check, does the x == 0 check suffice here?

martinjrobins: no unfortunately, because for y < 0 it should still be nan, for pow(0, -1) JAX gives:

```
pow(0, -1)  = inf
d/dx       = -inf
d/dy       = nan
```

wsmoses: but I don't think the y derivative is meaningful there -- or do you require it be nan?

martinjrobins: yea, that is a good point. My usecase is y>0 so I'm ambivalent for y < 0, happy to remove the y > 0 check

PR #2856, opened 2026-06-06
Fix tablegen path in cross builds

sgasho:
I hit the following error in r-l/r when I tried to enable Enzyme on x86_64 mac

The dist-x86_64-apple job runs on an arm64 macOS runner and cross-compiles Enzyme for x86_64. In that setup, the generated tablegen rules try to run `enzyme-tblgen` from `PATH`,
which fails with `enzyme-tblgen: command not found`.

Use `$<TARGET_FILE:enzyme-tblgen>` explicitly.

```shell
-- Build files have been written to: /Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/enzyme/build
running: cd "/Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/enzyme/build" && DESTDIR="" LC_ALL="C" LLVM_CONFIG_REAL="/Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/aarch64-apple-darwin/llvm/bin/llvm-config" "cmake" "--build" "/Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/enzyme/build" "--target" "install" "--config" "Release" "--" "-j" "14"
[0/2] Re-checking globbed directories...
[6/40] Building BinopDerivatives.inc...
FAILED: [code=127] Enzyme/BinopDerivatives.inc /Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/enzyme/build/Enzyme/BinopDerivatives.inc 
cd /Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/enzyme/build/Enzyme && enzyme-tblgen -gen-binop-derivatives -I/Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/src/tools/enzyme/enzyme/Enzyme -I/Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/llvm/include -I/Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/enzyme/build/include -no-warn-on-unused-template-args /Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/src/tools/enzyme/enzyme/Enzyme/InstructionDerivatives.td --write-if-changed -o /Volumes/WD_BLACK_SN850X_HS_1TB/rust-lang/x86-mac-enzyme/build/x86_64-apple-darwin/enzyme/build/Enzyme/BinopDerivatives.inc
/bin/sh: enzyme-tblgen: command not found
[7/40] Building BlasDerivatives.inc...
....
```

FYI: https://github.com/rust-lang/rust/pull/157385

### DIFF
--- a/enzyme/Enzyme/CMakeLists.txt
+++ b/enzyme/Enzyme/CMakeLists.txt
@@ -3,7 +3,7 @@
 #endif()
 
 if ("${Enzyme_TABLEGEN_EXE}" STREQUAL "")
-  set(Enzyme_TABLEGEN_EXE enzyme-tblgen)
+  set(Enzyme_TABLEGEN_EXE $<TARGET_FILE:enzyme-tblgen>)
 endif()
 
 get_target_property(TBL_LINKED_LIBS LLVMSupport INTERFACE_LINK_LIBRARIES)
--- a/enzyme/tools/CMakeLists.txt
+++ b/enzyme/tools/CMakeLists.txt
@@ -1,2 +1,1 @@
 add_subdirectory(enzyme-tblgen)
-set(Enzyme_TABLEGEN_EXE "${Enzyme_TABLEGEN_EXE}" PARENT_SCOPE)
--- a/enzyme/tools/CMakeLists.txt
+++ b/enzyme/tools/CMakeLists.txt
@@ -1,1 +1,2 @@
 add_subdirectory(enzyme-tblgen)
+set(Enzyme_TABLEGEN_EXE "${Enzyme_TABLEGEN_EXE}" PARENT_SCOPE)
--- a/enzyme/tools/enzyme-tblgen/CMakeLists.txt
+++ b/enzyme/tools/enzyme-tblgen/CMakeLists.txt
@@ -19,3 +19,4 @@ set_target_properties(enzyme-tblgen PROPERTIES FOLDER "Tablegenning")
 target_link_libraries(enzyme-tblgen
   PRIVATE)
 set(Enzyme_CONFIG_TABLEGEN_EXE enzyme-tblgen)
+set(Enzyme_TABLEGEN_EXE "$<TARGET_FILE:enzyme-tblgen>" PARENT_SCOPE)
--- a/enzyme/tools/enzyme-tblgen/CMakeLists.txt
+++ b/enzyme/tools/enzyme-tblgen/CMakeLists.txt
@@ -19,4 +19,3 @@ set_target_properties(enzyme-tblgen PROPERTIES FOLDER "Tablegenning")
 target_link_libraries(enzyme-tblgen
   PRIVATE)
 set(Enzyme_CONFIG_TABLEGEN_EXE enzyme-tblgen)
-set(Enzyme_TABLEGEN_EXE "$<TARGET_FILE:enzyme-tblgen>" PARENT_SCOPE)


sgasho:
@wsmoses 
Could you take a look? Is it reasonable to land this if CI is green?

wsmoses:
to confirm are you doing a cross compile build?

sgasho:
Yes, rust CI has an arm64 -> x86_64 macOS cross-compilation build.

sgasho:
The Julia CI failures do not look related to this PR.

wsmoses:
I don't understand why this is necessary -- can you explain what's happening? Folks are already cross compiling enzyme (see .packaging for an example)

sgasho:
In r-l/r

* The x86_64 macOS artifact is built on an arm64 macOS runner (= cross-compiling).
* The build fails with "enzyme-tblgen: command not found."
* Enzyme_TABLEGEN_EXE is set to enzyme-tblgen, which is then used as the COMMAND of an add_custom_command. https://github.com/EnzymeAD/Enzyme/blob/6f0b4c9945ada9461d054d7afd6455f78dacaa9f/enzyme/Enzyme/CMakeLists.txt#L5-L7
* cmake rewrites a target name there to its built path only when not cross-compiling — [COMMAND section in this doc](https://cmake.org/cmake/help/latest/command/add_custom_command.html)

.packaging avoids this by passing Enzyme_TABLEGEN_EXE explicitly in CMAKE_FLAGS:
https://github.com/EnzymeAD/Enzyme/blob/6f0b4c9945ada9461d054d7afd6455f78dacaa9f/.packaging/build_tarballs.jl#L94

So — is this change reasonable, or should we resolve it on the consumer side (r-l/r in this case), like Julia does?

sgasho:
While I was writing the above comment, I realized that we can just modify the following code to explicitly set the full path

https://github.com/EnzymeAD/Enzyme/blob/6f0b4c9945ada9461d054d7afd6455f78dacaa9f/enzyme/Enzyme/CMakeLists.txt#L5-L7

[fix Enzyme_TABLEGEN_EXE](https://github.com/EnzymeAD/Enzyme/pull/2856/commits/35e95988a74942168e102b04adba3dcee25e472f)

I'll check this change in rust side

sgasho:
> I'll check this change in rust side

I confirmed it works without issues

sgasho:
@wsmoses 
ping
Is the current diff reasonable now?

My conclusion is that the following fallback is not cross-compile-safe, and Julia's cross-compiling path avoids the issue because .packaging passes Enzyme_TABLEGEN_EXE explicitly in its build config.

```txt
 if ("${Enzyme_TABLEGEN_EXE}" STREQUAL "") 
   set(Enzyme_TABLEGEN_EXE enzyme-tblgen) 
 endif()
```

wsmoses (approved):

Issue #2858, opened 2026-06-06
[MLIR] MLIREnzymeAnalysis fails to build: getSuccessorInputs removed in current MLIR


agarret7:
## `MLIREnzymeAnalysis` fails to build against current MLIR (`getSuccessorInputs` removed)

`enzyme/Enzyme/MLIR/Analysis/ActivityAnalysis.cpp` calls
`RegionBranchOpInterface::getSuccessorInputs` and
`RegionBranchTerminatorOpInterface` in a way that no longer compiles
against MLIR main (and LLVM/MLIR 22). Both APIs were changed as part of
the region branch interface refactor upstream.

### Error

```
ActivityAnalysis.cpp:1169:28: error: 'class mlir::RegionBranchOpInterface'
    has no member named 'getSuccessorInputs'; did you mean 'getSuccessorRegions'?

ActivityAnalysis.cpp:1284:37: error: no match for call to
    '(...lambda...)(mlir::Operation*&, mlir::RegionBranchTerminatorOpInterface,
    std::deque<mlir::Block*>&)'
    note: no known conversion for argument 2 from
    'mlir::RegionBranchTerminatorOpInterface' to 'mlir::RegionBranchPoint'
```

### Impact

`MLIREnzymeAnalysis` is a transitive dependency of `MLIREnzymeTransforms`
and `enzymemlir-opt`, so the entire MLIR test suite (`test/MLIR/`) is
currently broken on main.

### Relevant upstream change

`getSuccessorInputs` was removed and `RegionBranchTerminatorOpInterface`
was replaced by `RegionBranchPoint` in the MLIR region branch interface
refactor. The fix likely involves updating the two call sites in
`ActivityAnalysis.cpp` to use the new API.

wsmoses:
currently we only support a ~near llvm main commit for mlir because the api changes so much. This is in contrast to llvm where we support llvm major versions 15+, and main. Is there a particular reason you want a build against mlir 22?

agarret7:
Yeah, I was thinking of exposing an Enzyme-enabled autodiff feature using Melior ([prototype](https://github.com/agarret7/melior-autodiff/tree/main)). Their bindings were upgraded to 22 a couple of months ago.

Perhaps I can add mlir 22 compatibility as a follow-up to #2859, or include it as part of the current PR?

wsmoses:
[...]

PR #2859, opened 2026-06-06
[MLIR] Add C API for Enzyme dialect (MLIRCAPIEnzyme)

agarret7:
## Add C API for Enzyme MLIR dialect (`MLIRCAPIEnzyme`)

The Enzyme MLIR dialect has no C API, making it inaccessible to
foreign-language bindings (Rust, Python, etc.) without depending on C++
headers. This PR adds the missing layer, following the same pattern used
by upstream MLIR's own dialect C APIs (`mlir-c/Dialect/Arith.h`, etc.).

### What's added

- **Dialect registration** — `MLIR_DECLARE/DEFINE_CAPI_DIALECT_REGISTRATION`
  so the dialect can be loaded into an `MlirContext` from C
- **`enzymeActivityAttrGet`** — constructs an `enzyme.activity` attribute
  from a C-side `EnzymeActivity` enum
- **`enzymeAutoDiffOpCreate` / `enzymeForwardDiffOpCreate`** — construct
  unattached `enzyme.autodiff` / `enzyme.fwddiff` ops from C, taking the
  same arguments as the TableGen-generated C++ builders
- **Build system** — `Integrations/c/CMakeLists.txt` with an
  `add_mlir_public_c_api_library(MLIRCAPIEnzyme ...)` target, wired into
  `Integrations/CMakeLists.txt` and `MLIR/CMakeLists.txt`

### Not included (deferred)

`enzymeRegisterPasses` is not included because `MLIREnzymeTransforms`
depends on `MLIREnzymeAnalysis`, which currently fails to build due to a
pre-existing MLIR API change (`getSuccessorInputs` →
`getSuccessorRegions`), tracked in #2858.

### Testing

~~`MLIRCAPIEnzyme` builds cleanly against LLVM/MLIR 22. The round-trip test will pass once the `MLIREnzymeAnalysis` build failures are resolved.~~

Edit: The test/MLIR/CAPI/roundtrip.mlir I included is a round-trip test for the ops themselves, not the C API. The existing ForwardMode/ReverseMode tests already cover that. A proper C API test requires a compiled C binary and lit infrastructure that doesn't exist in Enzyme's test suite yet. I've removed the placeholder. Happy to add proper C test infrastructure as a follow-up if that's the preferred direction.


enzyme/Enzyme/MLIR/Integrations/c/EnzymeMLIR.cpp
```diff
@@ -1,10 +1,100 @@
...
 #include "mlir/CAPI/IR.h"
+#include "mlir/CAPI/Registration.h"
 
 #include "Dialect/Dialect.h"
 #include "Dialect/Impulse/Impulse.h"
 #include "Dialect/Ops.h"
+MLIR_DEFINE_CAPI_DIALECT_REGISTRATION(Enzyme, enzyme,
+                                      mlir::enzyme::EnzymeDialect)
+
+MlirAttribute enzymeActivityAttrGet(MlirContext ctx, EnzymeActivity activity) {
```
wsmoses: make sure to mark these as MLIR_CAPI_EXPORTED

enzyme/Enzyme/MLIR/Integrations/c/EnzymeMLIR.cpp
```diff
@@ -1,10 +1,100 @@
...
+enzymeAutoDiffOpCreate(MlirContext ctx, MlirStringRef fn, MlirType *resultTypes,
+                       intptr_t nResults, MlirValue *inputs, intptr_t nInputs,
+                       MlirAttribute *activity, intptr_t nActivity,
+                       MlirAttribute *retActivity, intptr_t nRetActivity,
+                       int64_t width, bool strongZero, MlirLocation loc) {
+  auto *mlirCtx = unwrap(ctx);
+  llvm::SmallVector<mlir::Type> results;
+  collectTypes(resultTypes, nResults, results);
+  llvm::SmallVector<mlir::Value> operands;
+  collectValues(inputs, nInputs, operands);
```
mofeing: I believe you can use a `llvm::ArrayRef` here instead of a `llvm::SmallVector`, so you avoid the copying but not critical, and pass them directly to op creation.

mofeing (approved):
lgtm

wsmoses:
@mofeing note once this lands, we should adapt reactantextra to remove its copy of similar apis

wsmoses:
@agarret7 minor, but clang-format fails

agarret7:
Updated this PR with the following additions:

- `MLIR_CAPI_EXPORTED` on all function definitions
- Pass creation functions: `enzymeCreateBatchPass`, `enzymeCreateBatchDiffPass`, `enzymeCreateConvertEnzymeToMemRefPass`, `enzymeCreateRemoveUnusedEnzymeOpsPass`, 
`enzymeBatchOpCreate` constructor
- Rebased onto current main

Motivation: I'm planning a PR to [Melior](https://github.com/mlir-rs/melior) to expose Enzyme-backed autodiff. Melior upgraded to MLIR 22 a couple months ago, so I validated this C API end-to-end against MLIR 22 via a small Rust test harness using bindgen.

enzyme/Enzyme/MLIR/Integrations/c/EnzymeMLIR.cpp
```diff
@@ -1,13 +1,190 @@
...
+
+MLIR_CAPI_EXPORTED MlirPass enzymeCreateBatchDiffPass(void) {
+  return wrap(mlir::enzyme::createBatchDiffPass().release());
+}
+
+MLIR_CAPI_EXPORTED MlirPass enzymeCreateRemoveUnusedEnzymeOpsPass(void) {
+  return wrap(mlir::enzyme::createRemoveUnusedEnzymeOpsPass().release());
+}
+
+MLIR_CAPI_EXPORTED MlirAttribute enzymeActivityAttrGet(MlirContext ctx,
```
wsmoses, line 71: @agarret7 this can be simplified, see https://github.com/EnzymeAD/Reactant.jl/blob/a1092b113c9568b14eafad4eceaa89837eeae64a/deps/ReactantExtra/API.cpp#L390

wsmoses (commented):
sorry one more nit for the activity one, then lgtm

enzyme/Enzyme/MLIR/Integrations/c/EnzymeMLIR.cpp
agarret7: Thank you for the feedback! Will simplify `enzymeActivityAttrGet`. Also wondering if we shouldn't remove `enzymeAutoDiffOpCreate`, `enzymeForwardDiffOpCreate`, `enzymeJacobianOpCreate`, `enzymeBatchOpCreate`, and `activityArrayAttr` from this PR (the rest could be built downstream). What do you think, worth keeping?

agarret7: ditto for `enzymeRegisterPasses`

mofeing: for Op creation, at least in Reactant.jl, we always use `mlirOperationCreate`, so they are clearly not required.

`enzymeRegisterPasses` might be required though because AFAIK there's no other way to create a pass without calling the C++ API right?

wsmoses: even if not required I'm okay with keeping it, if it makes your or other downstream folks life easier

agarret7: `enzymeRegisterPasses` also isn't strictly needed today. Pass constructors return the pass directly, so you can add to a local `PassManager` without the global registry (unsafety can be wrapped):

```rust
pm.add_pass(unsafe {
    melior::pass::Pass::from_raw_fn(enzymeCreateDifferentiatePass)
});
```

The registry does matter for MLIR string parsing (`mlir-opt --pass-pipeline=...`), which melior already exposes through `parse_pass_pipeline`.

Given that, I suggest:

- Keep exposing the individual `enzymeCreate*Pass` constructors as it's needed to append to `PassManager` directly
- Keep `enzymeRegisterPasses` for the `mlir-opt` pipeline.
- Drop the AD `*Op`-construction helpers, since both should be expressible with the help of melior's existing builders (`OperationBuilder` + `enzymeActivityAttrGet`).

Curious what you think.

wsmoses: I'd go for keeping them, if it may make folks lives easier longer term [not just the work we're doing in julia or rust right now]

wsmoses (commented):
tangent [and @mofeing feel free to offer suggestions]

but can you also add this to the bazel build system too?

agarret7:
> tangent [and @mofeing feel free to offer suggestions]
> 
> but can you also add this to the bazel build system too?

I'm not super familiar with bazel, what would that entail?

agarret7:
> > tangent [and @mofeing feel free to offer suggestions]
> > but can you also add this to the bazel build system too?
> 
> I'm not super familiar with bazel, what would that entail?

Looking at enzyme/BUILD, the CInterface target ([BUILD:1052-1053](https://github.com/EnzymeAD/Enzyme/blob/c96508349d9f121957292308c646b4a1fd53c507/enzyme/BUILD#L1051-L1053)) already builds `Integrations/c/EnzymeMLIR.{cpp,h}`. Maybe no changes are needed for to add to bazel? I don't have it set up locally to confirm with a build, happy to if someone can point me at what is needed to test it.

PR #2861, opened 2026-06-08
Optionally check regionBranchOp

vimarsh6739:


wsmoses:
@vimarsh6739 can you revert this and instead consider ControlFlowAutoDiffOpInterface

wsmoses:
actually that doesn't cover it here, nvm

wsmoses:
found it, it was ADDataFlowOpInterface !

wsmoses:
see https://github.com/EnzymeAD/Enzyme-JAX/blob/689f721c7d09f6a3414dcd32c3d6ac3e58c4703f/src/enzyme_ad/jax/Implementations/StableHLOAutoDiffOpInterfaceImpl.cpp#L1983

PR #2863, opened 2026-06-08
Add Fortran test using activity descriptors

### LINKED ISSUE #2862: Add Fortran test using activity descriptors
In #2807 we introduced Fortran bindings to the activity descriptors. However, they aren't tested yet. I propose to start by including an implementation of the norm example from the Enzyme tutorial https://github.com/EnzymeAD/Enzyme-Tutorial/tree/main/2_norm.

joewallwork:
Closes #2862 
Merges into #2837

Adds the norm example from the Enzyme tutorial https://github.com/EnzymeAD/Enzyme-Tutorial/tree/main/2_norm with variants under ifx and flang.


joewallwork:
Well this is upsetting
https://community.intel.com/t5/Intel-Fortran-Compiler/symbol-E-missing-in-output-of-real-numbers-with-E0-or-ES0-format/m-p/1310862
Under the scientific notation format specifier `es0.0`, IFX writes `1.-06`, whereas every other Fortran compiler I've used would write `1.E-06`. For now we can just have different checks because the two variants of the `norm` example are currently run with two different compilers (see https://github.com/EnzymeAD/Enzyme/pull/2863/commits/ad68e034deea1bc0166f73c34a0d22f38feffe57). However, eventually we'll need to align this. A few options:
1. Add custom lit `CHECK` commands for different compilers (prefer not)
2. Use `#ifdef __INTEL_COMPILER` etc (prefer not)
3. Change what's being printed to avoid the issue, e.g., scale by 1e6 (preferable?)

vchuravy:
Yeah let's go with 3

joewallwork:
> Yeah let's go with 3

Done in f084c97fb3ea1ae823d0f9a38c9959ec00c8613e.

joewallwork:
To finish this example, I need to teach Enzyme how to differentiate the Fortran intrinsic `sum` (i.e. `_FortranASumReal4`) and also the `_FortranAAssign` that it sees in the LLVM.

vchuravy:
Suggestion from meeting: Choose a different test for this that doesn't require an intrinsic.

joewallwork:
Closing as superseded by #2917.

PR #2870, opened 2026-06-10
Add ifx 2023.2.0 to CI

vchuravy:


joewallwork (approved):
I added a fix to pick up the different pass syntaxes but otherwise looks good!

.github/workflows/fortran.yml
joewallwork: Oh actually, how about adding this to differentiate between the CI jobs?
```suggestion
    name: Fortran ${{matrix.build}} ${{matrix.os}} ifx ${{matrix.ifx}}
```

PR #2880, opened 2026-06-12
Support Julia 1.13's julia.atomicmodify pseudo-intrinsic

vchuravy:
Julia 1.13 (JuliaLang/julia#57010) emits atomic modify operations on integer-typed pointer-free fields as a variadic pseudo-intrinsic

```llvm
{old, new} = julia.atomicmodify.iN.pAS(ptr, op, ordering, syncscope, args...)
```

which atomically performs `old = *ptr; new = op(old, args...); *ptr = new`, where `op`'s parameter `i` (`i >= 1`) is forwarded from call operand `i + 3`. It is only expanded to `atomicrmw`/cmpxchg loops by Julia's `ExpandAtomicModify` pass after GC lowering, so Enzyme must keep it intact in generated code. Previously these calls hit the generic call differentiation path, which rejects variadic calls with `Number of arg operands != function parameters` (EnzymeAD/Enzyme.jl#3086).

### TypeAnalysis

Unify the pointee of `ptr` with both result elements and analyze `op` interprocedurally, relating the forwarded arguments with the modified memory. In particular this lets activity analysis prove atomic counter/lock manipulations inactive (Julia annotates real-world IR with `enzyme_type` already, but the interprocedural step recovers types for the modified value and forwarded operands).

### AdjointGenerator

Handle the call akin to `atomicrmw`:

- fully inactive calls are replayed via the constant fallback,
- non-differentiable (integer/pointer) modifications within duplicated memory are replicated on the shadow location with the primal arguments, keeping counters and lock states of shadow objects consistent,
- in forward mode, ops recognized as linear in `(old, value)` (fadd/fsub/xchg through bitcasts) compute the tangent by applying the same op to the shadow location and shadow value,
- in reverse mode (with inactive result), fadd/fsub accumulate the adjoint of the value operand from a load of the shadow location with downgraded ordering (negated for fsub),
- everything else reports `Active atomic modify not yet handled`.

Note that on Julia 1.13 codegen only emits the intrinsic for integer element types (float fields still lower to inline cmpxchg loops), so the float paths are exercised by the included lit tests; they implement the intrinsic's full semantics for when emission widens.

Companion Enzyme.jl change (running `ExpandAtomicModify` in its post-AD legalization pipeline): EnzymeAD/Enzyme.jl branch `vc/atomicmodify`. Tested end-to-end on Julia 1.13.0-rc1: the previously crashing `locks` and `threads` test suites pass, including reverse mode over atomic counters in duplicated structs.

### DIFF
--- a/enzyme/Enzyme/CallDerivatives.cpp
+++ b/enzyme/Enzyme/CallDerivatives.cpp
@@ -2209,6 +2209,66 @@ void AdjointGenerator::handleMPI(llvm::CallInst &call, llvm::Function *called,
   llvm_unreachable("Unhandled MPI FUNCTION");
 }
 
+// Classify the op callee of a julia.atomicmodify pseudo-intrinsic call
+// (op has signature elty (elty oldval, args...), with elty an integer type
+// holding the bits of the modified value) as an atomicrmw-style operation
+// whose derivative Enzyme knows. On success sets opArgNo to the index of the
+// op parameter providing the update value and FT to the floating point type
+// the update is performed in (nullptr for Xchg).
+static llvm::AtomicRMWInst::BinOp
+classifyAtomicModifyOp(llvm::Function *op, unsigned &opArgNo, llvm::Type *&FT) {
+  using namespace llvm;
+  FT = nullptr;
+  if (!op || op->empty() || op->isVarArg() ||
+      op->getFunctionType()->getNumParams() == 0)
+    return AtomicRMWInst::BAD_BINOP;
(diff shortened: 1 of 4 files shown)


LuuOW (commented):
Technical audit: code patterns and logic verified.

vchuravy:
I think this is needed to fix the assertion failure on 1.13 that is turning Julia CI red

wsmoses (approved):
I only did a quick read, @vchuravy if you gave it a deep review and are happy feel free to merge to unblock things

PR #2886, opened 2026-06-17
Allow setting strict-alias defaults at config time

ZuseZ4:


wsmoses:
as mentioned on the issue, I'm not convinced this is the right solution

Issue #2885, opened 2026-06-17
enzyme-strict-aliasing default should be configurable at build time.


ZuseZ4:
Whether or not some IR is valid or not under strict-aliasing is determined by the language. Rust does not have the concept, so we want to set the default to false in https://github.com/EnzymeAD/Enzyme/blob/41b6c734483d62764381194aeb15b8690f8c99ee/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L93

It doesn't matter for rustc driven builds, but every debug build has to specify `-enzyme-strict-aliasing=0` by hand, which doesn't make sense.
@copilot add a cmake option (default off) to change the default of strict-aliasing from true to false, so that rust people loading enzyme via opt stop running into it.

wsmoses:
what do you mean "It doesn't matter for rustc driven builds". Why is there a difference between a debug and non debug build?

ZuseZ4:
oh yeah, that was unclear. I didn't mean O0 vs O3, I meant the default workflow to minimize Enzyme bugs is to print Rust as LLVM-IR, and reproduce it from there on. We must always add that flag currently when we load enzyme into opt, which doesn't make sense. Languages that don't have strict-aliasing should be able to build Enzyme with a more resonable default.

wsmoses:
I mean there is only one opt? imo it would be more confusing to have the default change depending on setup [since then your reproducer may not repro in the enzyme explorer for example]

wsmoses:
Languages have different defaults and that's fine. Julia needs to always set the julia addr flag as well. I don't think that means we should change the general default here

ZuseZ4:
Fixed in https://github.com/EnzymeAD/Enzyme/pull/2886
C++ can keep it's default, in fact it won't affect any language not setting it. But it will safe significant time for people debugging Rust IR, since each time you minimize IR and forget to set it, you get a wrong error as a red hering.

wsmoses:
I think this will be more confusing if how you install enzyme comes with different defaults, since then there will be an issue when trying to use enzyme itself [eg in the explorer]. Can you just write this in the docs, or have your emit mwe script add the flag?

ZuseZ4:
I can otherwise overwrite it in our Rust fork, but I'd find it not great if we're forced to do that. We just got rid of the last patches.

[...]

PR #2887, opened 2026-06-17
Allow setting strict-alias defaults at config time

ZuseZ4:
Whether or not some IR is valid or not under strict-aliasing is determined by the language. Rust does not have the concept, so we want to set the default to false in https://github.com/EnzymeAD/Enzyme/blob/41b6c734483d62764381194aeb15b8690f8c99ee/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L93

It doesn't matter for rustc driven builds, but every debug build has to specify `-enzyme-strict-aliasing=0` by hand, which doesn't make sense.
@copilot add a cmake option (default off) to change the default of strict-aliasing from true to false, so that rust people loading enzyme via opt stop running into it.

Enzyme was unwilling to accept it in https://github.com/EnzymeAD/Enzyme/issues/2885, but I care more about the concrete benefit to Rust devs willing to spend time debugging Enzyme, than about a potential future person downloading Enzyme via rustup, and then being confused that it used cmake to tune defaults in favour of Rust.


ZuseZ4:
Thanks gh.

PR #2899, opened 2026-06-24
[MLIR] fix: complex.create reverse mode flips imaginary-operand gradient sign

### LINKED ISSUE #2900: Reverse mode: complex.create flips the imaginary-operand gradient sign
In MLIR reverse mode, a real value routed through `complex.create` then `complex.im` comes back with the wrong gradient sign. It is real-in / real-out (complex only as an intermediate), so there is no Wirtinger convention freedom: at an `f32 -> f32` boundary `conj` is identity, so the adjoint is just the ordinary real derivative.

```mlir
func.func @main(%t: f32) -> f32 {
  %zero = arith.constant 0.0 : f32
  %z = complex.create %zero, %t : complex<f32>
  %o = complex.im %z : complex<f32>
  return %o : f32
}
```

`f(t) = im(create(0, t)) = t`, so `df/dt` must be `+1`, but reverse mode gives `-1`. The regression test `complex_create_im.mlir` in #2899 is in the same IR-level form as the existing `test/MLIR/ReverseMode/complex.mlir`; running

```
eopt repro.mlir --enzyme-wrap="infn=main outfn= argTys=enzyme_active retTys=enzyme_active mode=ReverseModeCombined" --canonicalize --remove-unnecessary-enzyme-ops
```

the emitted adjoint computes the cotangent of `%t` as `-seed`:

```
%1 = arith.negf %0                     // -seed
%2 = complex.create %cst_0, %1         // (0, -seed)
%3 = complex.conj %2                   // (0, +seed)
%4 = complex.add %cst, %3              // (0, +seed)
%5 = complex.conj %4                   // (0, -seed)
%6 = complex.im %5                     // -seed   <- returned cotangent of t
```

The reverse generator conjugates the cotangent it fetches for each complex-result op (`createConjOp` on the diffe, `enzyme-tblgen.cpp`), and the other complex rules are written to account for that. `CreateOp` is not: its rule reads `ComplexIm(DiffeRet) = Im(conj(z_bar)) = -Im(z_bar)`, so the gradient flowing to the imaginary operand is negated. Its two contributions are real `f32`, where `createConjOp` is identity, so nothing downstream compensates. `complex.re` of a `create` is fine, since conj leaves the real part alone.

The `re`/`im` tests in `complex.mlir` miss this: their outputs are real (the fetched-diffe conj is irrelevant on real results) and they never put `create` in primal position.

Proposed fix with the regression test in #2899: negate the imaginary contribution in the `CreateOp` rule, matching the `NegF` the `ImOp` rule already uses, so `CreateOp` is consistent with the cotangent convention the other complex rules follow. Found via differential analysis of the derivative rule tables, agent-assisted, confirmed against the MLIR lit run.

kimjune01:
Fixes #2900.

`complex.create` in reverse mode negates the gradient on its imaginary operand. The reverse generator conjugates the cotangent it fetches for each complex-result op (`createConjOp` on the diffe), and the other complex rules account for that; `CreateOp` does not, so its rule reads `ComplexIm(DiffeRet) = Im(conj(z_bar)) = -Im(z_bar)`. Its contributions are real `f32` (where `createConjOp` is identity), so nothing downstream compensates. Full reasoning and the emitted-IR receipt are in #2900.

## Change

- Negate the imaginary-operand contribution in the `CreateOp` rule, matching the `NegF` the `ImOp` rule already uses, so `CreateOp` is consistent with the cotangent convention the other complex rules follow. `complex.re` is unaffected. Only `complex.create` in primal position changes; `complex.mlir` uses `create` only in the adjoint, so nothing there moves.
- Adds `test/MLIR/ReverseMode/complex_create_im.mlir`: differentiates `f(t) = im(create(0, t)) = t` and checks the returned cotangent is `+seed`. The corrected adjoint chain is an identity, so it canonicalizes to `return %arg1`; the buggy `-seed` chain cannot.

## Verification

Developed against your CI (no local MLIR build here). The test-only commit's MLIR run went red on the sign, which is the bug. The post-fix run currently needs maintainer approval to start (`action_required`), so a green-after-fix is pending your approval rather than confirmed by me. Happy to adjust a CHECK line against the exact canonicalized IR once it runs.


mofeing (approved):
thanks! this is correct. we do the same thing in Enzyme-JAX for the `stablehlo::ComplexOp` op.

https://github.com/EnzymeAD/Enzyme-JAX/blob/d4ddff0e2a69caea42c8c75998c75b51a69a662e/src/enzyme_ad/jax/Implementations/HLODerivatives.td#L1087-L1090

PR #2906, opened 2026-06-26
Handle gc-transition bundle from Julia v1.12

vchuravy:
Handle the new gc-transition bundle from Julia 1.12.

It signifies that a function call is potentially blocking and does not interact with the Julia runtime,
and that it is thus safe to run GC during it. 

Fixes:

- https://github.com/EnzymeAD/Enzyme.jl/issues/2139#issuecomment-4763197828
- https://github.com/EnzymeAD/Enzyme.jl/issues/3235
- https://github.com/EnzymeAD/Enzyme.jl/issues/3195

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -260,10 +260,8 @@ bool GradientUtils::usedInRooting(const llvm::CallBase *orig,
   orig->getOperandBundlesAsDefs(OrigDefs);
   SmallVector<OperandBundleDef, 2> Defs;
   for (auto bund : OrigDefs) {
-    if (bund.getTag() == "gc-transition")
-      continue;
-    // Only handle jl_roots tag (for now).
-    if (bund.getTag() != "jl_roots") {
+    // Only handle jl_roots & gc-transition tags (for now).
+    if (bund.getTag() != "jl_roots" && bund.getTag() != "gc-transition") {
       errs() << "unsupported tag " << bund.getTag() << " for " << *orig << "\n";
       llvm_unreachable("unsupported tag");
     }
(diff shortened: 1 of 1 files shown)


vchuravy:
The original version of the PR treated `gc-transition` as inactive since the only value it has is the current pointer to the ptls.

But there is some commentary https://github.com/JuliaLang/julia/blob/e0a6d0e019d9f2ccfe1977ee7da32e8e1cbb1f0c/src/ccall.cpp#L2161 on making it gc_safe + gc-transition in one bundle.

Treating it similarly shouldn't cause any issues currently since we will simple see an inactive value.

enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -303,27 +303,29 @@ GradientUtils::getInvertedBundles(CallInst *orig, ArrayRef<ValueType> types,
...
-        if (ty == ValueType::Shadow || ty == ValueType::Both)
-          anyShadow = true;
-      }
+    bool anyPrimal = false;
+    bool anyShadow = false;
+    for (auto ty : types) {
+      if (ty == ValueType::Primal || ty == ValueType::Both)
+        anyPrimal = true;
+      if (ty == ValueType::Shadow || ty == ValueType::Both)
+        anyShadow = true;
```
vchuravy: @wsmoses Claude review spotted this. `anyShadow` depends only on `types` and as such is invariant to `bund.inputs`.

The alternative fix would be to look at the specific type belonging to `inp`

wsmoses: oh sure. Longer term it would be great if we could figure out how to associate the input with the gc root [and thus only force shadow rooting for the shadowed args], but not available yet

wsmoses (commented):
can we add a test for the gc-transition

PR #2911, opened 2026-06-30
GridKit integration test

nkoukpaizan:
This adds an integration test for Enzyme in GridKit. 

All other GridKit optional dependencies are off by default, so this only tests the ability to  compute model Jacobians. 

Minor modifications to the inputs passed to the package-specific `run.sh`.

cc @wsmoses @pelesh


.github/workflows/integration.yml
```diff
@@ -55,5 +56,5 @@ jobs:
         run: make -j `nproc`
       - name: "Run tests"
         run: |
-          bash integration/${{ matrix.package }}/run.sh ${{ matrix.llvm }} enzyme/build/ClangEnzyme-${{ matrix.llvm }}.so `nproc`
```
wsmoses, line 58: is there a reason for removing the path to the .so? I would've expected this to be useful [and can be ignored if needed]

integration/gridkit/run.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Running GridKit integration tests"
+
+CLANGV=$1
+ENZYME_DIR=$2
+
+echo "CLANGV: $CLANGV" 
+echo "ENZYME_DIR: $ENZYME_DIR"
+
+git clone git@github.com:ORNL/GridKit.git
```
wsmoses: can we pick a fixed commit of GridKit for now [and we can setup an action or something to update over time]

just for the sake of making sure that the only reason this fails is that code in this repo changes, instead of trying to figure out if here or upstream

.github/workflows/integration.yml
wsmoses: ah I understand, okay if that's the case we might as well also change this into the full path, adding `pwd`/enzyme/build

.github/workflows/integration.yml
```diff
@@ -55,5 +56,5 @@ jobs:
...
+          bash integration/${{ matrix.package }}/run.sh ${{ matrix.llvm }} enzyme/build `nproc`
```
wsmoses: are you using the gpu right now, if not can we explicitly exclude gridkit + gpu?

wsmoses: @nkoukpaizan bumping this q

nkoukpaizan: We're not using the GPU. Will push a fix in a sec.

pelesh: Just to clarify, GPU support is in our roadmap. It is unlikely that GPU work will start in the next few months.

wsmoses: yeah that's fine, its mostly just so we can avoid running gpu ci for something not using gpus [dont have as much gpu ci atm as cpu]

wsmoses:
hopefully you can see the ci results, but if not:
```
Run bash integration/gridkit/run.sh 18 $(PWD)/enzyme/build `nproc`
Run '/home/runner/k8s/index.js'
using script executor
/__w/_temp/8a2c9c5f-5f59-4690-99a9-37c6268c31a0.sh: line 1: PWD: command not found
Running GridKit integration tests
CLANGV: 18
ENZYME_DIR: /enzyme/build
Cloning into 'GridKit'...
Host key verification failed.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
mkdir: cannot create directory 'GridKit/build': No such file or directory
integration/gridkit/run.sh: line 13: cd: GridKit/build: No such file or directory
fatal: detected dubious ownership in repository at '/__w/Enzyme/Enzyme'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/Enzyme/Enzyme
CMake Warning:
  Ignoring extra path from command line:

   ".."


CMake Error: The source directory "/__w/Enzyme" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
make: *** No targets specified and no makefile found.  Stop.
make: *** No rule to make target 'test'.  Stop.
Error: Error: ScriptExecutorError when trying to execute: "Error: Job failed with exit code 2."
Error: Process completed with exit code 1.
```

(and also let me know if you cant and/or want me to add you to the repo for ease)

nkoukpaizan:
> hopefully you can see the ci results, but if not:
> (and also let me know if you cant and/or want me to add you to the repo for ease)

I can see it. I was just doing something stupid there (trying to clone via ssh instead of https).
Being added to the repo will be helpful!

nkoukpaizan:
@wsmoses, Thanks for bearing with me here. Tests are finally passing.

wsmoses:
nice! as we get more of the infra set up, I may poke you for possibly adding info out of this for benchmark times too [and obviously if anything goes wrong with a random enzyme commit which escapes my understanding, we may ask your help]

PR #2912, opened 2026-07-01
tblgen: add ResultIndex primitive

Pangoraw:
to refer to the result value of an operation in the tablegen rules.

cc @mofeing, as we had discussed about this.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/Common.td
+++ b/enzyme/Enzyme/MLIR/Implementations/Common.td
@@ -81,6 +81,10 @@ class Operation<bit usesPrimal_, bit usesShadow_, bit usesCustom_=0> {
   bit usesCustom = usesCustom_;
 }
 
+class ResultIndex<int index_> {
+  int index = index_;
+}
+def Result : ResultIndex</*index=*/0>;
 class DiffeRetIndex<list<int> indices_> {
   list<int> indices = indices_;
 }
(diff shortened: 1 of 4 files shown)


Pangoraw:
ideally mincut can also identify equivalence to merge nodes in the graph but this can still be interesting to write more memory optimized rules

wsmoses:
I'm slightly worried about this [though less so for EnzymeMLIR than Enzyme proper] because the analysis to detect if a value is needed for the reverse looks at users of the value, and doesn't consider "itself" as a potential user. In principle is there a reason why re-running the op wouldn't suffice (and CSE-style optimizations won't make it equivalent perf?]

Pangoraw:
> CSE-style optimizations won't make it equivalent perf

the push / pop mincut does not currently treat the recomputation as equivalent and as such will always choose to cache the operand of the original op instead of its result.

```mlir
scf.for {
  enzyme.push %v
  %0 = math.exp %v
  ...
}

scf.for {
  %v = enzyme.pop
  %dx = math.exp %v
  ...
}
```

could be
```mlir
scf.for {
  %0 = math.exp %v
  enzyme.push %0
  ...
}

scf.for {
  %dx = enzyme.pop
  ...
}
```

because the graph looks something like this

```
               %v
              /  \
math.exp (fwd)    math.exp (bwd)
             |    |

```

so it is less costly to store %v than its result

wsmoses:
That makes sense, though I feel like we should min cut here (and that would regardless be useful for other things too)

PR #2918, opened 2026-07-03
Autodiff op activity optimization

vimarsh6739:


enzyme/Enzyme/MLIR/Passes/EnzymeAcitivityOpt.cpp
```diff
@@ -72,23 +72,23 @@ static FailureOr<ActivityOptimization> computeClassicalActivityOptimization(
     Operation *diffOp, FunctionOpInterface callee,
     ArrayRef<Activity> argumentActivity, ArrayRef<Activity> resultActivity,
```
wsmoses, line 73: minor but can you fix the name of the file [activity is misspelled]

vimarsh6739: oh just noticed this - will do

PR #2930, opened 2026-07-08
Recognize Numba NRT_decref as a deallocation function

ludgerpaehler:
To cleanly differentiate via the Numba runtime (NRT), NRT_MemInfo_alloc* is registered as an allocation (shadow handler), but its paired release, NRT_decref, was not recognized as a deallocation.

Without this addition Enzyme duplicates the (inactive-marked) refcount op across the augmented-forward and reverse passes and separately frees the allocation, multi-freeing it and corrupting the heap.

Recognizing NRT_decref in isDeallocationFunction lets Enzyme cache the allocation and emit a single, correctly-placed free, matching how it handles malloc / jl_alloc_array.

### DIFF
--- a/enzyme/Enzyme/LibraryFuncs.h
+++ b/enzyme/Enzyme/LibraryFuncs.h
@@ -135,6 +135,13 @@ static inline bool isDeallocationFunction(const llvm::StringRef name,
       return true;
     if (name == "swift_release")
       return true;
+    // Numba runtime (NRT) reference-count release / deallocation.  Recognizing
+    // NRT_decref as a deallocation lets Enzyme cache the paired
+    // NRT_MemInfo_alloc* allocation and emit a single, correctly-placed free
+    // instead of duplicating the refcount op across the augmented/reverse
+    // passes (which multi-frees and corrupts the heap).
+    if (name == "NRT_decref")
+      return true;
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


wsmoses (commented):
add a test?

ludgerpaehler:
Test added. CI still needs to pass.

ludgerpaehler:
Tests look good afaict, let me know if you need more for the PR

wsmoses (changes requested):
can you put an actual numba ir case here?

ludgerpaehler:
> can you put an actual numba ir case here?

Put in, CI looks on track to me. Let me know if you want further changes.

wsmoses (changes requested):
sorry this isn't the right way of doing thing and looks kind of like claude just spewing a bit of ai slop [e.g. this should be symmetrically registered in isAllocationFunction like the free function was added]

ludgerpaehler:
> sorry this isn't the right way of doing thing and looks kind of like claude just spewing a bit of ai slop [e.g. this should be symmetrically registered in isAllocationFunction like the free function was added]

Point taken, I am sorry that you had to read the incoherent IR. I'll sit down in calm on the weekend and hand-write a proper example without running between multiple things s.t. it is semantically coherent and is in line with the style of e.g. the rust deallocator example

PR #2944, opened 2026-07-14
Fix indexing bug in EnzymeBatchDiffPass

vimarsh6739:
The op's index was being used to index into primal inputs


wsmoses:
@vimarsh6739 this fails format, please check ci before merging

PR #2945, opened 2026-07-15
Utils: use target IntPtrType for CreateAllocation's malloc size arg

zacharywhitley:
`CreateAllocation` fed `Count->getType()` as the size type into `CreateMalloc`. On 64-bit hosts this coincidentally matches `size_t`; on 32-bit targets it doesn't. On wasm32-wasi an i64 `Count` (typical when reverse-mode tape math widens the size) produces `call i8* @malloc(i64 %n)` against wasi-libc's `void *malloc(i32)` declaration — the wasm module loads and then traps at first invocation with `signature_mismatch:malloc`.

Fix: compute `IntPtrTy = DataLayout::getIntPtrType(ctx, 0)` and cast `Count` via `CreateZExtOrTrunc` on the default path before feeding `CreateMalloc`. 64-bit hosts unchanged (IntPtrTy == i64 == Count's type, no cast emitted). wasm32 gets `malloc(i32)` as wasi-libc expects. Also updates the size-wrap-flag hint to match against the cast operand.

The `CustomAllocator` branch is untouched — user-supplied callback contract preserved.

Reproducer (wasi-sdk-33 / LLVM 22 / Enzyme main, wasm32-wasip1 target): reverse-mode over `sum((A*x - y)^2)` MSE loss traps at load without this patch, compiles+runs cleanly with it. Verified in a downstream wasi-sdk overlay project (tegmentum/enzyme-wasm) under wasmtime 46.

### DIFF
--- a/enzyme/Enzyme/Utils.cpp
+++ b/enzyme/Enzyme/Utils.cpp
@@ -622,7 +622,30 @@ Value *CreateAllocation(IRBuilder<> &Builder, llvm::Type *T, Value *Count,
   Value *res;
   auto &M = *Builder.GetInsertBlock()->getParent()->getParent();
   auto AlignI = M.getDataLayout().getTypeAllocSizeInBits(T) / 8;
-  auto Align = ConstantInt::get(Count->getType(), AlignI);
+
+  // On 32-bit targets (wasm32-wasi, 32-bit ARM/x86, etc.) the C ABI's
+  // malloc takes size_t = i32; if upstream tape-cache math widened
+  // Count to i64, feeding it straight into CreateMalloc emits
+  // `call i8* @malloc(i64 ...)` against wasi-libc's `malloc(i32)` and
+  // traps at load with signature_mismatch:malloc. Truncate Count to the
+  // target IntPtrType only when Count is WIDER than IntPtrTy -- narrower
+  // Counts are already handled correctly by LLVM's usual i64-widening
+  // before the malloc call on 64-bit hosts, and altering that path would
+  // rearrange the golden IR that existing FileCheck tests record.
+  //
+  // CustomAllocator is a user-supplied callback whose type contract we
+  // preserve unchanged.
+  Type *IntPtrTy = M.getDataLayout().getIntPtrType(M.getContext(), 0);
+  Value *AllocCount = Count;
+  Type *AllocSizeTy = Count->getType();
+  if (!CustomAllocator && AllocSizeTy->isIntegerTy() &&
+      IntPtrTy->isIntegerTy() &&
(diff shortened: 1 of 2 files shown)


wsmoses:
lot of relevant ci failures

zacharywhitley:
Thanks. I'll take a look at what's failing and get you an updated PR. Awesome project BTW

enzyme/Enzyme/Utils.cpp
```diff
@@ -622,7 +622,30 @@ Value *CreateAllocation(IRBuilder<> &Builder, llvm::Type *T, Value *Count,
...
+  //
+  // CustomAllocator is a user-supplied callback whose type contract we
+  // preserve unchanged.
+  Type *IntPtrTy = M.getDataLayout().getIntPtrType(M.getContext(), 0);
+  Value *AllocCount = Count;
+  Type *AllocSizeTy = Count->getType();
+  if (!CustomAllocator && AllocSizeTy->isIntegerTy() &&
+      IntPtrTy->isIntegerTy() &&
+      AllocSizeTy->getIntegerBitWidth() > IntPtrTy->getIntegerBitWidth()) {
+    AllocCount = Builder.CreateZExtOrTrunc(Count, IntPtrTy, Name + ".size");
```
wsmoses: minor but you don't need zext or trunc here, right?

wsmoses (commented):
add a relevant test so this doesn't decay in the future?

zacharywhitley:
Pushed `ecfd980b` addressing the review points:

- **CreateZExtOrTrunc → CreateTrunc.** You're right — the guard above only enters the branch when Count is strictly wider than IntPtrTy, so the zext leg was dead. `CreateTrunc` conveys intent and drops the impossible-branch cost.
- **Regression test** at `enzyme/test/Enzyme/ReverseMode/wasm32_malloc_size.ll`. Reverse-mode over a two-level accumulation (reduced from the MSE reproducer in the PR description), `wasm32-unknown-wasi` triple. The inner accumulator's LCSSA lift forces per-outer-iter tape caching, which is the CreateAllocation path this fix gates. Asserts:
  - a `trunc i64 ... to i32` is emitted (the cast this patch adds),
  - every Enzyme-emitted `@malloc(...)` cache call takes an i32 size,
  - no `@malloc(i64 ...)` slips through (the runtime `signature_mismatch:malloc` mode),
  - the malloc declaration is `@malloc(i32)`.

  Verified locally with LLVM 22 + the LLVMEnzyme-22 plugin built against this branch. On 64-bit hosts the CHECK-NOT wouldn't catch anything (existing tests still emit i64 mallocs), so the wasm32 triple is load-bearing.

The Format failure in the previous run appeared to be against unrelated code that has since been fixed upstream — the new push should be clean, and I'll follow up if it isn't.

PR #2950, opened 2026-07-15
mlir: add enzyme.atomic_rmw op with ordering attribute

pengmai:


enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -444,6 +444,37 @@ def DumpOp : Enzyme_Op<"dump"> {
...
+  let description = [{
+  }];
+
+  let results = (outs AnyType:$result);
+
+  let arguments = (ins
+      AtomicRMWKindAttr:$kind,
+      AtomicOrdering:$ordering,
+      AnyType:$value,
+      Arg<AnyMemRef, "the reference to rmw", [MemRead, MemWrite]>:$memref,
```
wsmoses, line 490: Can we also preserve the alignment here (though I suppose we could just have alignment as a generic setAttr call)

enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
```diff
@@ -480,9 +480,10 @@ struct AffineLoadOpInterfaceReverse
             computeAffineIndices(builder, loadOp.getLoc(),
                                  loadOp.getAffineMap(), retrievedArguments,
                                  indices);
-            memref::AtomicRMWOp::create(builder, loadOp.getLoc(),
-                                        arith::AtomicRMWKind::addf, gradient,
-                                        memrefGradient, indices);
+            enzyme::AtomicRMWOp::create(
+                builder, loadOp.getLoc(), gradient.getType(),
+                arith::AtomicRMWKind::addf, Ordering::monotonic, gradient,
+                memrefGradient, indices);
```
wsmoses: we also need to do the same for the `enzyme::AffineAtomicRMWOp` too

wsmoses (approved):

PR #2951, opened 2026-07-16
Fix compatibility with LLVM 23

nikic:
OptimizationLevel is an enum now.

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -3135,7 +3135,7 @@ AnalysisKey EnzymeNewPM::Key;
 
 static InlineParams getInlineParamsFromOptLevel(OptimizationLevel Level) {
 #if LLVM_VERSION_MAJOR >= 23
-  return getInlineParams(Level.getSpeedupLevel());
+  return getInlineParamsFromOptLevel(static_cast<unsigned>(Level));
 #else
   return getInlineParams(Level.getSpeedupLevel(), Level.getSizeLevel());
 #endif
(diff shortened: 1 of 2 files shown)


enzyme/Enzyme/Enzyme.cpp
```diff
@@ -3135,7 +3135,7 @@ AnalysisKey EnzymeNewPM::Key;
 
 static InlineParams getInlineParamsFromOptLevel(OptimizationLevel Level) {
 #if LLVM_VERSION_MAJOR >= 23
-  return getInlineParams(Level.getSpeedupLevel());
+  return getInlineParamsFromOptLevel(static_cast<unsigned>(Level));
```
nikic, line 3138: This was previously passing the optimization level to an API accepting an inlining threshold ... so this was effectively setting a very low inlining threshold.

enzyme/Enzyme/Enzyme.cpp
```diff
@@ -3271,7 +3271,11 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
     // system libraries and other oracles.
     MPM.addPass(InferFunctionAttrsPass());
 
+#if LLVM_VERSION_MAJOR >= 23
+    if (Level > OptimizationLevel::O1) {
```
wsmoses: Can we put the { outside the ifdef (so basic code completion tools don't get confused)

nikic: Done!

nikic:
I updated the commit has in the MLIR test similar to what was done in https://github.com/EnzymeAD/Enzyme/commit/7be5247eb207215f04a1d8e3150d6a7394b7a85d.

The commit has is the last one before LLVM 23 was branched.

PR #2955, opened 2026-07-18
Fix vectorized MLIR derivative constants

jacobleft:
Fixes zero-operand `ConstantFP` expressions emitted by the MLIR derivative TableGen backend.

In width>1 forward mode these constants previously used the primal result type, while surrounding derivative values use the shadow type with a leading width axis. This caused invalid HLO operations such as `tensor<batchxf32> - tensor<widthxbatchxf32>` when an `enzyme.batch` broadcast was differentiated under vector forward mode.

A companion Enzyme-JAX regression exercises `enzyme.batch` composed with width-2 forward AD. Local focused validation is still building.

### DIFF
--- a/enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
+++ b/enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
@@ -618,7 +618,12 @@ bool handle(const Twine &curIndent, const Twine &argPattern, raw_ostream &os,
         std::string ord;
         bool shadowType = false;
         if (resultRoot->getNumArgs() == 0) {
-          ord = "op->getResult(0)";
+          ord = "gutils->getShadowType(op->getResult(0)";
+          // This constant participates in a tangent expression.  In vector
+          // forward mode, the tangent has a leading width dimension even when
+          // the primal result does not.  Use the shadow type so a constant
+          // such as `1 - tanh(x)^2` composes with width-vectorized operands.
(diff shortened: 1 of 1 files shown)


enzyme/tools/enzyme-tblgen/enzyme-tblgen.cpp
```diff
@@ -619,6 +619,11 @@ bool handle(const Twine &curIndent, const Twine &argPattern, raw_ostream &os,
         bool shadowType = false;
         if (resultRoot->getNumArgs() == 0) {
           ord = "op->getResult(0)";
+          // This constant participates in a tangent expression.  In vector
```
wsmoses, line 622: I'm not sure this is right, can you add a test which fails without this?

jacobleft: Added the StableHLO regression in Enzyme-JAX: https://github.com/jacobleft/Enzyme-JAX/commit/4f2f89ff97eafce107fb2e3e5dba5c127a74718c\n\nIt runs `enzyme-batch` before width-2 forward AD. With the parent `7be5247e`, the test fails with `stablehlo.subtract(tensor<4xf32>, tensor<2x4xf32>)`; with this PR head `5ec3a28c`, the same target passes. The test belongs in Enzyme-JAX because it exercises its StableHLO derivative rules.

PR #2958, opened 2026-07-19
Return log-densities + divergent transition info from `impulse.infer`

penelopeysm:
Currently `infer` only returns a single diagnostic, a boolean which indicates whether or not the transition was accepted. Naturally, it's often useful to have more info :)
Right now (some of) this info is already calculated but is thrown away. This PR is a pretty mechanical way of surfacing that. The `diagnostics` tensor is expanded to include a second column which indicates whether the transition is divergent, plus a separate new tensor containing the logdensities is returned.
This is a bit ugly and not incredibly generalisable, but since it's quite closely tied to the sampling algorithm and `infer` only admits two algorithms, I think it's fine for now. My feeling is that properly generalising diagnostics only really makes sense in the context of broader programmable inference APIs.
This ofc needs changes in Reactant as well which I'll push separately. I have tested this locally end to end and it all works as intendeds


sbrantq (approved):
Thanks this looks great! ... and by returning log densities it helps with  https://github.com/EnzymeAD/Reactant.jl/issues/2920

sbrantq:
(looks like we had a CI fix upstream)

PR #2971, opened 2026-07-22
mlir: derivative for maxnumf, pointer handling for arith.select

pengmai:


enzyme/Enzyme/MLIR/Implementations/ArithAutoDiffOpInterfaceImpl.cpp
```diff
@@ -102,13 +102,99 @@ struct ArithSubFSimplifyMathInterface
...
+
+struct SelectActivityInterface
+    : public ActivityOpInterface::ExternalModel<SelectActivityInterface,
+                                                arith::SelectOp> {
+  bool isInactive(Operation *op) const { return false; }
+  bool isArgInactive(Operation *op, size_t idx) const {
+    // arith.select is not inactive in general, but the condition is always
+    // inactive.
+    auto selectOp = cast<arith::SelectOp>(op);
+    return selectOp.getCondition() == selectOp.getOperand(idx);
```
wsmoses: we should just say return idx == 0;

enzyme/Enzyme/MLIR/Implementations/ArithAutoDiffOpInterfaceImpl.cpp
```diff
@@ -102,13 +102,99 @@ struct ArithSubFSimplifyMathInterface
...
+    // values
+    if (!gutils->isConstantValue(selectOp.getResult())) {
+      auto iface =
+          dyn_cast<AutoDiffTypeInterface>(selectOp.getResult().getType());
+      if (iface && iface.isMutable()) {
+        return success();
+      }
+
+      Value condition = gutils->popCache(caches.front(), builder);
+      Value zero = arith::ConstantOp::create(
```
wsmoses: also we can use create null value of type here instead of arith constantop

enzyme/Enzyme/MLIR/Implementations/ArithAutoDiffOpInterfaceImpl.cpp
```diff
@@ -102,13 +102,99 @@ struct ArithSubFSimplifyMathInterface
...
+          builder, selectOp.getLoc(), FloatAttr::get(selectOp.getType(), 0.0));
+      Value dret = gutils->diffe(selectOp.getResult(), builder);
```
wsmoses, line 127: gutils->setDiffe(selectOp.getResult(), zero);

PR #2973, opened 2026-07-22
mlir: remove dependency on linalg with enzyme.fill_zero op

pengmai:


enzyme/Enzyme/MLIR/Dialect/EnzymeOps.td
```diff
@@ -349,6 +349,14 @@ def StoreOp : Enzyme_Op<"store",
   }];
 }
 
+def FillZeroOp : Enzyme_Op<"fill_zero", []> {
+  let summary = "Fill memory with zero values";
+  let arguments = (ins Arg<AnyMemRef, "the argument to fill", [MemWrite]>:$memref);
+  let assemblyFormat = [{
+    $memref attr-dict `:` type($memref)
```
wsmoses, line 356: we may or may want to keep the index operands to memref.alloca/alloc/etc here [aka the sizes of the non-statically known dimensions], for use in lowering

wsmoses: we can also consider handling this during lowering though by looking at the op passed to it

PR #2975, opened 2026-07-22
mlir: pass to split ifs with multi-results into separate ops

pengmai:
This pass splits `scf.if` ops that produce multiple results into separate ops that each produce one result. Currently, mincut assumes that the results of region cf ops like `scf.if` must be cached, which produces suboptimal results for a number of GPU benchmarks. This pass makes it easier to get mincut to re-compute the results of these ops, though the longer term solution would be to get mincut to properly support partially re-computing region ops. cc: @Pangoraw @ftynse


wsmoses (approved):
lgtm as a pass, but we separately should figure out if we can avoid using this [since copying the recursive ops in a region can be expensive and duplicate subexpressions, etc]

PR #2981, opened 2026-07-22
Use gcp-cache for MLIR build cache

wsmoses:


wsmoses:
@giordano I vibe coded my own github actions/cache, and it does work here. The load from cache is ~30s slower than github itself, but ~infinite space

wsmoses:
okay is now 18s so I'd say we're good [after some more opts]

PR #2996, opened 2026-07-25
Prefer OptionalPassInfoMixin

boomanaiden154:
This became preferred in LLVM 23 and PassInfoMixin will be moving to a detail namespace in LLVM 24, which means compilation errors without this patch.


wsmoses:
```
/home/runner/work/Enzyme/Enzyme/enzyme/Enzyme/FixupJuliaCallingConvention.cpp: At global scope:
/home/runner/work/Enzyme/Enzyme/enzyme/Enzyme/FixupJuliaCallingConvention.cpp:1592:35: error: expected template-name before ‘<’ token
 1592 |     : public OptionalPassInfoMixin<FixupJuliaCallingConventionNewPM> {
 ```

Issue #2998, opened 2026-07-25
Forward mode rejects llvm.exp10


SpeedyTurtle599:
Hi everyone, hope you're well. I've been enjoying using Enzyme in my research group's in-house Kokkos CFD solver. I've run into an issue with `llvm.exp10`, where `__enzyme_fwddiff` over a function containing `std::pow(10.0, x)` fails with an error:

```
error: Enzyme: cannot handle (forward) unknown intrinsic
```

My system specs: M3 Max MacBook Pro, macOS 26.6 RC, Homebrew LLVM 22.1.8. I've been using it with Kokkos 5.1.1, OpenMP 22.1.8, and OpenMPI 5.0.9, but I can reproduce the issue without any of those dependencies on every version of Enzyme from 280-289 inclusive.

Full disclosure: every bit of "diagnosis" and reproduction below was generated using Claude Code's help and does not represent my own original work. I did my best to understand and condense it for easier conversation, but it may contain LLM-generated mistakes.

---

As far as I can tell, `InstructionDerivatives.td` does carry a rule for `exp10`, but it's a `CallPattern` matched on a call to a function literally *named* `exp10`:

```tablegen
def : CallPattern<(Op $x),
                  ["exp10"],
                  [(FMul (FMul (DiffeRet), (Call<(SameFunc), [ReadNone,NoUnwind]> $x)),
                         (ConstantFP<"2.30258509299404568401799145468"> $x))],
                  (ForwardFromSummedReverse),
                  [ReadNone, NoUnwind]>;
```

`log10` does not reproduce this problem because it is declared as an `IntrPattern` matched on `llvm.log10` directly:

```tablegen
def : IntrPattern<(Op $x),
                  [["log10"]],
                  [(FDiv (DiffeRet), (FMul (ConstantFP<"2.302585092994046"> $x), $x))],
                  (ForwardFromSummedReverse)>;
```

Claude thinks that LLVM 19+ changed instcombine to fold `pow(10.0, x)` into the `llvm.exp10` intrinsic. No call to a function named `exp10` survives to match, so the match rule never fires. The fold is intrinsic-to-intrinsic, so `-fno-builtin-exp10` does not prevent it.

Claude made me a repro script:
```cpp
#include <cmath>

extern "C" int enzyme_dup;
template <typename... Args> void __enzyme_fwddiff(void*, Args...);

static void primal(const double* x, double* out) {
[...]

PR #3001, opened 2026-07-25
Add exp10 intrinsic for LLVM >=18 (#2998)

SpeedyTurtle599:
Previously, `__enzyme_fwddiff` over a function containing `std::pow(10.0, x)` fails with an error:

```
error: Enzyme: cannot handle (forward) unknown intrinsic
```

This small PR adds the intrinsic, gated behind LLVM >=18. It should close issue #2998.


wsmoses (approved):
lgtm

SpeedyTurtle599:
@wsmoses, may I get your eyes on this one again? It passed the checks before I merged in main, and it would be great to be able to move my downstream users back onto main.

PR #3005, opened 2026-07-26
Bound lookupM recompute recursion to avoid compile-time blowup

oschulz:
Generated by Claude Fable, in context of the EnzymeAD/Enzyme.jl#3298 CI hang. EnzymeAD/Enzyme.jl#3298 uses a workaround now, this should fix the underlying cause. This might be fixable in a more fundamental way though, see below:

## Symptom

Reverse-mode differentiation of some functions hangs Enzyme in codegen — no crash, one thread pinned at 100% indefinitely. It surfaced downstream as a reverse gradient of `HypergeometricFunctions._₂F₁` timing out CI for >2 h across Julia 1.10–1.12 on Linux and Windows (EnzymeAD/Enzyme.jl#3298).

## Root cause

The stuck thread is entirely inside `GradientUtils::lookupM` ⇄ `unwrapM`:

```
lookupM     GradientUtils.cpp   (recompute attempt: shouldRecompute -> unwrapM)
unwrapM     GradientUtils.cpp   (AttemptSingleUnwrap; getOp on each operand)
lookupM     GradientUtils.cpp
unwrapM     GradientUtils.cpp
...  dozens of alternating frames, entered from visitBinaryOperator's
     lookup() inside several nested recursivelyHandleSubfunction levels
```

In `AttemptSingleUnwrap`, `getOpFull` routes every operand back through `lookupM`, and phi unwrapping materializes fresh `_phirc` blocks that it then builds into. Because `lookup_cache`/`unwrap_cache` are keyed by the current insert block, those freshly-created blocks keep missing the cache, so sibling recursions re-explore the same phi/binop web once per predecessor. The work is therefore combinatorial in the operand DAG rather than linear.

How badly it blows up is order-sensitive (it depends on traversal order over pointer-keyed maps), which is why the same call can take 20 s on one machine and hang for hours on another, and why it looked so erratic across the CI matrix.

## What this PR does

Adds a per-function budget on recompute *attempts*. `enzyme-lookup-recompute-budget` (default 100000): once `lookupM` has made that many recompute attempts while differentiating one function, it stops attempting to recompute and takes the existing cache-to-tape path, which is legal for any value. A negative value disables the bound (prior behavior).

This is deliberately a *bound*, in the same spirit as the existing `enzyme-max-type-offset` / `enzyme-max-type-depth` / `enzyme-max-int-offset` cutoffs — it does not change the memoization that lets the recursion explode in the first place. If you'd prefer to fix that directly (e.g. so the phi-rebuild recursion memoizes across the new blocks instead of re-walking), I'm happy to take a run at it or to defer to you; the reproducer below makes either approach easy to check.

## Evidence

A/B on GitHub runner hardware (each run prints which libEnzyme it loaded; workflow + staged reproducer on the `debug/gamma-hang` branch of oschulz/Enzyme.jl, [run](https://github.com/oschulz/Enzyme.jl/actions/runs/30174164037)):

| build | ₂F₁ reverse |
| --- | --- |
| stock, no bound (×3) | all hang (>20 min, killed) |
| this patch, budget disabled `=-1` (×2) | all hang (same binary — the bound is the only difference) |
| this patch, budget `=100000` (×3) | complete in **20.9 / 21.4 / 21.7 s** |

Gradients are unchanged in every completing configuration; locally I also ran `budget=5`, which forces the cache fallback on essentially every lookup, and the gradient is still exact — so the fallback path is not the concern, it's only slower. A nontrivial control (the scalar `gamma` reverse, Cephes with loops/branches) stays well under the budget and is byte-for-byte unaffected.

## Notes

- The default (100000) is the value I measured: it leaves the scalar case untouched and bounds the pathological case to ~21 s. It's a starting point — happy to change it.
- Validated through the Enzyme.jl integration above (against the v0.0.288 release, where this change is textually identical) rather than the C++ lit suite; I couldn't build the lit harness locally. A minimized `.ll` reproducer would be a good follow-up — glad to work on one.

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -108,6 +108,13 @@ llvm::cl::opt<bool>
 llvm::cl::opt<bool>
     EnzymeSpeculatePHIs("enzyme-speculate-phis", cl::init(false), cl::Hidden,
                         cl::desc("Speculatively execute phi computations"));
+llvm::cl::opt<int> EnzymeLookupRecomputeBudget(
+    "enzyme-lookup-recompute-budget", cl::init(100000), cl::Hidden,
+    cl::desc("Maximum number of recompute attempts lookupM may make while "
+             "differentiating one function before it stops attempting to "
+             "recompute values and always caches them instead. Bounds a "
+             "combinatorial lookupM/unwrapM recursion; a negative value "
+             "disables the bound."));
 llvm::cl::opt<bool> EnzymeFreeInternalAllocations(
     "enzyme-free-internal-allocations", cl::init(true), cl::Hidden,
     cl::desc("Always free internal allocations (disable if allocation needs "
@@ -7056,6 +7063,15 @@ Value *GradientUtils::lookupM(Value *val, IRBuilder<> &BuilderM,
       return result;
     }
   }
+
+  // The recompute attempt below recurses through lookupM/unwrapM chains whose
+  // cost can grow combinatorially (e.g. deep phi/binop webs re-explored per
+  // predecessor). Once too many attempts have been made for this function,
+  // stop trying to recompute and fall through to caching, which is legal for
+  // any value, so the worst case stays bounded.
+  if (EnzymeLookupRecomputeBudget >= 0 &&
+      ++lookupRecomputeCount > EnzymeLookupRecomputeBudget)
+    tryLegalRecomputeCheck = false;
 
   // TODO consider call as part of
   bool lrc = false, src = false;
--- a/enzyme/Enzyme/GradientUtils.h
+++ b/enzyme/Enzyme/GradientUtils.h
@@ -326,6 +326,10 @@ private:
            llvm::ValueMap<llvm::Value *, llvm::WeakTrackingVH>>
       lookup_cache;
 
+  // Number of recompute attempts lookupM has made for this function, bounded
+  // by EnzymeLookupRecomputeBudget.
+  int64_t lookupRecomputeCount = 0;
+
 public:
   void replaceAndRemoveUnwrapCacheFor(llvm::Value *A, llvm::Value *B);
 
--- a/enzyme/test/Enzyme/ReverseMode/juliahypergeometric.ll
+++ b/enzyme/test/Enzyme/ReverseMode/juliahypergeometric.ll
@@ -1,0 +1,652 @@
+; RUN: if [ %llvmver -ge 16 ]; then %opt < %s %newLoadEnzyme -passes="enzyme" -enzyme-preopt=false -S | FileCheck %s; fi
+
+; Reduced from Enzyme.jl: the reverse gradient of HypergeometricFunctions'
+; _2F1 with respect to all four arguments (EnzymeAD/Enzyme.jl#3298). This is
+; the z -> infinity continuation, BInf, whose derivative Enzyme spends almost
+; all of its time on: lookupM tries to recompute values that live inside the
+; loops here, rebuilding phis into fresh blocks whose siblings do not share
+; lookup results, so the exploration multiplies out.
+;
+; A compile time test. With the recompute bound this differentiates in about
+; a second; without it, this reduced module takes ~50x longer, and the
+; unreduced one does not finish at all.
+
+target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128-ni:10:11:12:13"
+target triple = "x86_64-linux-gnu"
+
+; Function Attrs: nofree
+declare void @julia.safepoint(ptr) local_unnamed_addr #0
+declare nonnull ptr @julia.get_pgcstack() local_unnamed_addr "enzyme_inactive"
+
+; Function Attrs: nocallback nofree nosync nounwind speculatable willreturn memory(none)
+declare double @llvm.fabs.f64(double) #1
+
+define dso_local double @julia_gamma_3408(double %"_x::Float64") local_unnamed_addr {
+top:
+  %pgcstack = call ptr @julia.get_pgcstack()
+  %current_task = getelementptr inbounds i8, ptr %pgcstack, i64 -152
+  %ptls_field = getelementptr inbounds i8, ptr %pgcstack, i64 16
+  %ptls_load = load ptr, ptr %ptls_field, align 8
+  %0 = getelementptr inbounds i8, ptr %ptls_load, i64 16
+  %safepoint = load ptr, ptr %0, align 8
+  fence syncscope("singlethread") seq_cst
+  call void @julia.safepoint(ptr %safepoint)
+  fence syncscope("singlethread") seq_cst
+  %1 = fcmp olt double %"_x::Float64", 0.000000e+00
+  %2 = fsub double %"_x::Float64", %"_x::Float64"
+  %3 = fcmp uno double %2, 0.000000e+00
+  %or.cond.not = or i1 %1, %3
+  ret double 0.000000e+00
+}
+
+; Function Attrs: nocallback nofree nosync nounwind speculatable willreturn memory(none)
+declare double @llvm.rint.f64(double) #1
+
+define hidden fastcc double @julia_G_3448(double %"z::Float64", double %"\CF\B5::Float64") unnamed_addr {
+top:
+  %pgcstack = call ptr @julia.get_pgcstack()
+  %current_task = getelementptr inbounds i8, ptr %pgcstack, i64 -152
+  %ptls_field = getelementptr inbounds i8, ptr %pgcstack, i64 16
+  %ptls_load = load ptr, ptr %ptls_field, align 8
+  %0 = getelementptr inbounds i8, ptr %ptls_load, i64 16
+  %safepoint = load ptr, ptr %0, align 8
+  fence syncscope("singlethread") seq_cst
+  call void @julia.safepoint(ptr %safepoint)
+  fence syncscope("singlethread") seq_cst
+  %1 = call double @llvm.rint.f64(double %"z::Float64")
+  %2 = fcmp ult double %1, 0xC3E0000000000000
+  %3 = fcmp uge double %1, 0x43E0000000000000
+  %narrow.not = or i1 %2, %3
+  %4 = fsub double %1, %1
+  %5 = fcmp une double %4, 0.000000e+00
+  %or.cond = or i1 %narrow.not, %5
+  ret double 0.000000e+00
+}
+
+declare dso_local double @julia___3420(double, i64) local_unnamed_addr #2
+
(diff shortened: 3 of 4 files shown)


oschulz:
I asked Claude to look into this in more detail at max effort - sorry, this is really beyond my Enzyme expertise, I'm posting it here in the hope that it's useful for the experts:

Some context on scope, in case it is useful for deciding how you would like this handled.

**This is a bound rather than a root fix.** The recursion can explode because `unwrapM` memoizes only successful unwraps — the `goto endCheck` paths return null without recording anything — so a subtree that fails to unwrap is re-explored on each later visit, and for phis every failed attempt speculatively builds `_phirc` blocks and erases them again. There is also no in-progress guard, so a node can be re-expanded within its own subtree. The budget caps the total cost; the repeated work still happens below it.

**It is not a new code path.** `lookupM` goes back to 2019, `AttemptSingleUnwrap` to 2020, and the phi-recomputation logic to #524 (2022). What is unusual here is the shape of the workload: a polyalgorithm with a dozen-plus dispatch arms, differentiated with respect to four simultaneously active arguments, about five levels deep through `recursivelyHandleSubfunction` — enough re-exploration for the cost to become visible.

If you would like, I am happy to follow up with the two changes that address the repetition itself rather than capping it:

1. **Memoize failed unwraps.** Record `(val, insert block, unwrap mode) -> failed`, so a repeated probe returns in O(1) instead of re-descending and re-building/erasing IR. The `(inst, block)` key already exists next to it for `UnwrappedWarnings`. The correctness question is `available`: a later call with a richer available-set can succeed where an earlier one failed, so negative entries would need tagging with an availability epoch (bumped whenever `available`/`scopeMap` grows) and ignoring once stale. Getting that wrong would not affect gradients — falling back to caching is always legal — but it would trade tape memory for compile time, so it is worth being careful about.

2. **Add an in-progress set for `(val, block)`**, so re-entering a node inside its own expansion returns null immediately rather than recursing again. Returning null is always legal here. One subtlety: that null is a "cycle" answer rather than a definitive failure, so it must not be written into the negative cache from (1).

Together those turn the repeated-subtree exploration into a DAG traversal. The budget could stay afterwards as a cheap backstop for any shape that still slips through, or be dropped if you would rather not carry the extra option.

Longer term, the more principled version is to separate "is this recomputable here?" from "build it", so the decision becomes a memoizable analysis instead of a speculative build-and-erase. That is a larger refactor of `legalRecompute` / `shouldRecompute` / `unwrapM`, so I would rather follow your preference on whether it is worth doing at all.

wsmoses:
can you have fable add an llvm lit test which reproduces the original problem when this patch is not applied

oschulz:
> can you have fable add an llvm lit test which reproduces the original problem when this patch is not applied

On it.

oschulz:
@wsmoses , I hope this does the trick - Claude Opus-5 max-effort:

Added `enzyme/test/Enzyme/ReverseMode/lookuprecomputeblowup.ll`.

It differentiates a loop whose body is a chain of 20 diamonds. The body's values change every iteration and their blocks do not dominate the return, so the reverse pass cannot reuse them in place and `lookupM` tries to recompute each merge phi. Rebuilding a phi unwraps it into fresh blocks, one per predecessor, and those siblings do not share lookup results, so every diamond multiplies the work.

Same file, same machine, LLVM 18, `opt -passes=enzyme -enzyme-preopt=false`, no extra flags:

| build | time |
| --- | --- |
| with this patch | **0.56 s** (`lit` reports PASS in 0.61 s) |
| without this patch | **still running when killed at 300 s** |

"Without this patch" is a separately built plugin with the change removed from the source, not the bound disabled by flag.

The growth is clean, so the depth is easy to justify. Unpatched, at 10/11/12/13 diamonds: 1.17 / 3.61 / 11.39 / 37.01 s — roughly 3x per level. Patched, the same measurements stay at ~0.58 s all the way from 10 to 24 diamonds, since the bound caps the attempts. I used 20 in the test so the unbounded case is unambiguously stuck rather than merely slow (it extrapolates to about a day).

Gradients are unchanged. On a smaller instance of the same shape the derivative is bit-identical with the bound at its default, with the bound disabled, and with the bound set to 10 — which forces the cache fallback on essentially every lookup — and all three match the analytic derivative and central differences.

The test uses the pre-opaque-pointer spelling for the `__enzyme_autodiff` call so it parses on the LLVM 15 and 16 configurations in CI as well as on 17+.

oschulz:
To make sure I had Claude Fable-5 re-check and polish (it added commit ba02ca878554ab1d6154054f8026249a053e97bb).

Fable reports:

- Ran the full `test/Enzyme` lit suite (1046 tests) against the patched build and against a build with the change removed from the source. On LLVM 15 the suite is green with the patch (1036 passed, 10 expected failures) and the only differing result between the two builds is the new test, which times out unpatched. On LLVM 18 the two builds also agree on every test except the new one (many suite tests can't run there at all — typed-pointer CHECK lines — identically in both builds, so the comparison still holds).
- Built the plugin against LLVM 15 as well and ran the test there, so the pre-16 old-PM RUN line is exercised too (passes in ~1.2 s).
- Added a third RUN with `-enzyme-lookup-recompute-budget=10`, which sends essentially every lookup down the caching fallback (~0.04 s); that keeps the fallback path covered even if a future improvement means the default budget is never reached at this depth.

wsmoses:
can you isolate specifically the ir from the case you had in enzyme.jl that times out without this. In particular I think this is the wrong fix and I want a reproducer to investigate what to do as the right fix

oschulz:
> can you isolate specifically the ir from the case you had in enzyme.jl that times out without this.

I'll try (resp.  have Claude try ...).

> In particular I think this is the wrong fix and I want a reproducer to investigate what to do as the right fix

Yes, it is a bit of a workaround, I know. Claude suggested a few more fundamental ways above, but this is really beyond my expertise, so I can't judge. But I'm happy to produce whatever is helpful.

oschulz:
EnzymeAD/Enzyme.jl#3298 , which triggered this, should not stall anymore though (test was changed there).

oschulz:
Here is the isolated IR: [`pre-ad-dump-x86_64` artifact](https://github.com/oschulz/Enzyme.jl/actions/runs/30284409631) (from the workflow on my fork's `debug/gamma-hang` branch). It contains `pread_julia_F_2471.ll` — the exact module Enzyme.jl hands to `EnzymeCreatePrimalAndGradient` for the ₂F₁ reverse gradient, dumped right after `memcpy_alloca_to_loadstore` / `force_recompute!` / `EnzymeDetectReadonlyOrThrow`, with the `enzyme_type` attributes in place — plus a `.meta` with the call parameters (`Const{F}`, four `Active{Float64}`, `DEM_ReverseModeCombined`, width 1, `returnPrimal=false`).

It reproduces through plain `opt`, no Julia involved. Append this to the module:

```llvm
declare { double, double, double, double } @__enzyme_autodiff(i8*, ...)

define { double, double, double, double } @df(double %a, double %b, double %c, double %z) {
entry:
  %r = call { double, double, double, double } (i8*, ...) @__enzyme_autodiff(i8* bitcast (double (double, double, double, double)* @julia_F_2471 to i8*), double %a, double %b, double %c, double %z)
  ret { double, double, double, double } %r
}
```

and run

```
opt -load-pass-plugin=LLVMEnzyme-18.so -load=LLVMEnzyme-18.so --enzyme-attributor=0 \
    -passes=enzyme -enzyme-preopt=false -disable-verify -S input.ll -o /dev/null
```

(`-disable-verify` because Enzyme puts a function-level `readonly` on `strcmp`, which the LLVM 18 verifier rejects; the Enzyme.jl pipeline never runs that verifier.)

On a GitHub `ubuntu-24.04` runner (x86_64, LLVM 18.1.7, Enzyme v0.0.288 built clean): three runs, all still inside the `lookupM`/`unwrapM` recursion when killed at 20 minutes (gdb stacks in the [workflow log](https://github.com/oschulz/Enzyme.jl/actions/runs/30284409631)). Same file with the budget branch: 10 s. The dump job's own Julia-side gradient call also hung until its 40-minute cap on that runner.

One thing that took me a while to see: the trigger is architecture-shaped. The identical procedure on aarch64 — where Julia generates a structurally different module for the same function — finishes in 0.4 s with or without the bound, across 24 layout-perturbed runs. That matches the original CI pattern exactly: every hanging job was x86-64 (ubuntu, windows), while the macOS (aarch64) jobs always passed. So for investigating the real fix, the x86_64 artifact above is the one that matters; an aarch64 dump of the same call is small-timing in every configuration I tried.

Happy to run `llvm-reduce` over it with a timing oracle if a minimized version would help, and fine with this PR being superseded by whatever you conclude from the IR.

wsmoses:
add it as a lit test?

oschulz:
> add it as a lit test?

Working on it.

oschulz:
> add it as a lit test?

This took a long time, due to many stalls, but here it is - it's added to the PR for now, but I assume you don't want it in the repo long term.

Claude says:

Added as `enzyme/test/Enzyme/ReverseMode/juliahypergeometric.ll`.

It is the real thing, reduced: starting from the module Enzyme.jl hands to Enzyme for the ₂F₁ reverse gradient, `llvm-reduce` took it from 4.0 MB / 210 functions to **29 KB / 7 functions** (637 lines). What survives is the part that matters — `BInf`, the z → ∞ continuation, calling `_recInf` twice and through it `gamma`, `G` and `P`.

Finding the right starting point mattered more than the reduction itself. `-enzyme-print` emits each derivative inner-to-outer, so the last name printed before the stall (`diffejulia__recInf___3527`) identifies the function whose differentiation never finishes — its caller, `BInf`. `llvm-extract --recursive` on that gives a 751 KB module that already reproduces it, and reducing that is far easier than reducing the whole thing.

Two things about the interestingness test are worth mentioning, because the first version got them wrong. It requires **both** that the patched build still produces a derivative **and** that the clean build fails to finish — with only the second condition, `llvm-reduce` happily satisfies it by breaking the module (it nulled out `@julia.get_pgcstack`, which makes the clean build spin and then fail). And the reduced file needed `@julia.get_pgcstack` marked `"enzyme_inactive"`, since outside the Enzyme.jl pipeline nothing else tells Enzyme what that function is.

On this machine (aarch64, LLVM 18.1.7): with the bound, ~1 s; without it, ~50 s. So this file is comfortable to iterate on, which I assume is what you want for looking at the real fix — it is a 50× signal rather than a hang. If you would rather have something that does not finish at all, the 531 KB intermediate still hangs indefinitely and I can add that instead, and the full unreduced module is attached to [this run](https://github.com/oschulz/Enzyme.jl/actions/runs/30284409631).

The synthetic `lookuprecomputeblowup.ll` from earlier is still in the branch; happy to drop it now that there is a real one, or keep both — your call. And as before, no attachment to the bound itself if the IR points you at a better fix.

oschulz:
@wsmoses do you also want the big "531 KB intermediate" that Claude describes above?

oschulz:
Gentle ping @wsmoses .

oschulz:
@wsmoses can I generate some additional stuff that would help with this.

PR #3007, opened 2026-07-27
Update to later llvm

wsmoses:


vchuravy:
IIRC there are two MLIR test that fail due to new canonicalizations.


```quote
141 tests, 139 pass, 2 fail:
- MLIR/ReverseMode/complex_create_im.mlir
- MLIR/ReverseMode/affine_parallel_mincut.mlir

Root cause — both failures are one upstream MLIR change, not an Enzyme regression

llvm-project commit dccdacc2dc49 (2026-07-15), "[mlir][arith] Canonicalize addf(negf(x), y) to subf(y, x)", added the AddFOfNegFLhs/AddFOfNegFRhs patterns. Both tests run --canonicalize and both pin the pre-canonicalization negf + addf spelling.

- complex_create_im: without --canonicalize Enzyme emits %7 = arith.negf %6 / %8 = arith.addf %7, %cst_0, exactly what the test expects. The canonicalizer now rewrites that pair to %7 = arith.subf %cst_0, %6. I hand-traced the gradient through the output — it is still +seed, so the fix from #2899 is intact and the derivative is correct.
- affine_parallel_mincut: same rewrite, arith.negf %[[MULF_9]] + arith.addf %[[MULF_8]], %[[NEGF_0]] collapses to arith.subf %[[MULF_8]], %[[MULF_9]]. I patched a copy replacing those two CHECK lines with the single subf line and the whole test passes — confirming that's the only divergence, so the mincut/cache-placement behavior the test actually guards is unchanged.
````

vchuravy (approved):

PR #3009, opened 2026-07-27
Fixes for mutable memory in scf.for reverse

Pangoraw:


enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -218,7 +227,10 @@ struct ForOpInterfaceReverse
     strides.push_back(b.getIndexAttr(1));
     for (int64_t i = 0, e = rowTy.getRank(); i < e; ++i) {
       offsets.push_back(b.getIndexAttr(0));
-      sizes.push_back(b.getIndexAttr(rowTy.getDimSize(i)));
+      if (rowTy.isDynamicDim(i))
+        sizes.push_back(memref::DimOp::create(b, loc, buf, i + 1).getResult());
```
wsmoses, line 231: I'm worried this may not always be resolvable [e.g. from pointers] if there's another way we can get this information

wsmoses: currently the polygeist lowering turns memref->pointer so size info is lost. if guaranteed to be in the same fn as the memref allocation it will fold though, but split ad in the future might have a problem

enzyme/Enzyme/MLIR/Implementations/SCFAutoDiffOpInterfaceImpl.cpp
```diff
@@ -244,7 +256,11 @@ struct ForOpInterfaceReverse
                               Type valTy) {
     if (auto mt = dyn_cast<MemRefType>(valTy)) {
       Value row = checkpointRow(b, loc, buf, slot, mt);
-      Value fresh = memref::AllocOp::create(b, loc, mt);
+      SmallVector<Value> dynSizes;
+      for (int64_t i = 0, e = mt.getRank(); i < e; ++i)
+        if (mt.isDynamicDim(i))
+          dynSizes.push_back(memref::DimOp::create(b, loc, row, i));
+      Value fresh = memref::AllocOp::create(b, loc, mt, dynSizes);
```
wsmoses, line 263: we should avoid this copy op for the same reason as below, if possible [since memref size is unknown]

enzyme/test/MLIR/ReverseMode/scf_for_mutable_memory.mlir
```diff
@@ -0,0 +1,71 @@
...
+// CHECK:           %[[ALLOC_0:.*]] = memref.alloc() : memref<10xindex>
+// CHECK:           %[[ALLOC_1:.*]] = memref.alloc() : memref<10xindex>
+// CHECK:           %[[FOR_0:.*]] = scf.for %[[IV:.*]] = %[[C0]] to %[[C10]] step %[[C1]] iter_args(%[[ACC:.*]] = %[[CST]]) -> (f64) {
+// CHECK:             memref.store %[[IV]], %[[ALLOC_0]]{{\[}}%[[IV]]] : memref<10xindex>
+// CHECK:             %[[LOAD_0:.*]] = memref.load %[[ARG0]]{{\[}}%[[IV]]] : memref<10xf64>
+// CHECK:             %[[ADDF_0:.*]] = arith.addf %[[ACC]], %[[LOAD_0]] : f64
+// Every iteration caches the shadow memref itself (%[[ARG1]]), not a fresh
+// per-iteration copy -- it is a mutable handle whose identity must be
+// preserved across iterations.
+// CHECK:             memref.store %[[ARG1]], %[[ALLOC]]{{\[}}%[[IV]]] : memref<10xmemref<10xf64>>
```
Pangoraw, line 43: for a dup grad, the multidim optimization was enabled, but this was wrong because we want the accumulations to happen in the duplicated memory.

now we could optimize to hoist the push since the memory handle does not change in the loop:

```mlir
%cache = enzyme.init() -> enzyme.Cache<memref<10xf32>>

scf.for {
  enzyme.push(%cache, %arg1)
}
scf.for {
  %dup = enzyme.pop(%cache) -> memref<10xf32>
}

// optimized to
enzyme.push(%cache, %arg1)
scf.for {}
%dup = enzyme.pop(%cache) -> memref<10xf32>
scf.for {

}

wsmoses (approved):
lgtm minus the size issues, we can solve them in follow up though if needed

wsmoses:
sorry I'm not saying you need to remove the dynamic dims, but just an fyi that the result may not be foldable

wsmoses:
going to merge for now, but @Pangoraw we should revisit the above dyn dim stuff

wsmoses:
well I'll let it automerge once you resolve your comments

PR #3020, opened 2026-07-28
Add more Fortran tests

### LINKED ISSUE #3018: Add more Fortran tests
We could include versions of the `dot` and `norm` examples from https://github.com/EnzymeAD/Enzyme-Tutorial/ but with hand-coded dot product and sum (respectively) with the intrinsics as TODOs.

joewallwork:
Closes #3018 

Add some more Fortran tests based on those found in https://github.com/EnzymeAD/Enzyme-Tutorial.

I also took the liberty of making the naming scheme for the programs, modules, and procedures consistent across the tests and also the formatting.

It's worth noting that differentiation through intrinsics does work, just not at `-O0`. As soon as you apply optimisation, they don't appear in the LLVM IR.

### DIFF
--- a/.github/workflows/fortran.yml
+++ b/.github/workflows/fortran.yml
@@ -80,8 +80,6 @@ jobs:
         versions:
           - llvm: 15
             ifx: 2023.0.0
-          - llvm: 16
-            ifx: 2023.1.0
           - llvm: 17
             ifx: 2023.2.4
     timeout-minutes: 60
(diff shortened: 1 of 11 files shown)


vchuravy (approved):

joewallwork:
@copilot Fix the `-opaque-pointers` issue in the Fortran CI with IFX 2023.1.0 and LLVM 16 that I've attempted to fix in the past few commits.

joewallwork:
Hm so Copilot's 'fix' is to disable those tests. If we are going to do that then it would probably be better to just drop that version from the CI.

joewallwork:
Rebased on top of `main` and included another attempt at getting ifx 2023 working.

joewallwork:
Well that didn't work. I say we should just drop support for IFX 2023.1.0 - see latest commit.

vchuravy (approved):

PR #3022, opened 2026-07-28
Retain unused pointer trait metadata

krystophny:
## Reproducer

Configure Enzyme MLIR against LLVM/Clang/MLIR 23.1.0-rc2 and build `MLIREnzymeTransforms`. Clang diagnoses `NumLowBitsAvailable` as unused, and Enzyme promotes that project warning to an error.

## Root cause

LLVM 23 does not instantiate this `PointerLikeTypeTraits` member on the affected build path, although the member remains part of the pointer-trait contract.

## Minimal fix

Mark the metadata member `[[maybe_unused]]`. This retains the trait contract and changes neither its value nor generated IR.

## Validation

- `MLIREnzymeTransforms` builds with `/usr/lib/llvm23/bin/clang++` 23.1.0-rc2.
- `enzymemlir-opt` builds with the same toolchain.
- Existing `MLIR/Passes/mem2reg.mlir` passes through `FileCheck`, providing a transformation smoke oracle.
- `git diff --check` passes.

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
@@ -219,7 +219,7 @@ template <> struct llvm::PointerLikeTypeTraits<Node> {
   static inline Node getFromVoidPointer(void *p) {
     return Node::getFromOpaqueValue(p);
   }
-  static constexpr int NumLowBitsAvailable =
+  [[maybe_unused]] static constexpr int NumLowBitsAvailable =
       llvm::PointerLikeTypeTraits<Inner>::NumLowBitsAvailable;
 };
 


wsmoses:
Currently EnzymeMLIR [unlike EnzymeLLVM] does not have guarantees to support fixed releases of llvm, instead following a commit close to head. EnzymeLLVM supports all fixed releases llvm 15+, and the fixed commit close to head.

This is in part because EnzymeMLIR is under a lot more moving development, and that MLIR changes its APIs a lot more. Is there a reason you need to support llvm23 and not head?

krystophny:
Thanks, that answers the compatibility question. The motivation was a reproducible LLVM 23 toolchain, but EnzymeLLVM covers that use case; we do not need to carry a fixed-release EnzymeMLIR workaround. Closing this.

PR #3024, opened 2026-07-28
[mlir] Make non-looping, non-nested RegionBranchOpInterface ops movable by min-cut

pengmai:
This change works in conjunction with the `split-multi-results` pass to re-compute `scf.if` ops instead of caching their results.

### DIFF
--- a/enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
+++ b/enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
@@ -13,6 +13,7 @@
 #include "Utils.h"
 #include "mlir/Analysis/TopologicalSortUtils.h"
 #include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/RegionUtils.h"
 #include <cassert>
 #include <deque>
 
@@ -361,9 +362,37 @@ static inline bool isLoadMovable(Operation *op) {
   return op->hasAttr("enzyme.readonly");
 }
 
+static inline bool isMovable(Operation *op);
+static inline bool isRegionBranchMovable(RegionBranchOpInterface regionBranch) {
+  // We can move region ops that only contain pure operations. As a heuristic,
+  // we only consider non-looping ops.
+  if (regionBranch.hasLoop())
+    return false;
+  for (auto &region : regionBranch->getRegions()) {
+    // Regions with multiple blocks potentially contain loops
+    if (!region.hasOneBlock())
+      return false;
+
+    for (auto &bodyOp : region.front()) {
+      if (auto bodyRegionBranch = dyn_cast<RegionBranchOpInterface>(&bodyOp)) {
+        if (!isRegionBranchMovable(bodyRegionBranch))
+          return false;
+        // Terminators are considered not movable, but do not impact our ability
+        // to move their enclosing region op.
+      } else if (&bodyOp != region.front().getTerminator() &&
+                 !isMovable(&bodyOp)) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
 // Whether or not an operation can be moved from the forward region to the
 // reverse region or vice-versa.
 static inline bool isMovable(Operation *op) {
+  if (auto regionBranch = dyn_cast<RegionBranchOpInterface>(op))
+    return isRegionBranchMovable(regionBranch);
   return op->getNumRegions() == 0 && op->getBlock()->getTerminator() != op &&
          (mlir::isPure(op) || isLoadMovable(op));
 }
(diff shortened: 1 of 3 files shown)


enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -361,9 +361,30 @@ static inline bool isLoadMovable(Operation *op) {
   return op->hasAttr("enzyme.readonly");
 }
 
+static inline bool isRegionBranchMovable(RegionBranchOpInterface regionBranch) {
+  // We can move region ops that only contain pure operations. As a heuristic,
+  // we only consider non-nested and non-looping ops.
+  if (regionBranch.hasLoop())
+    return false;
+  for (auto &region : regionBranch->getRegions()) {
+    // Regions with multiple blocks potentially contain loops
```
wsmoses, line 372: I think there's also an interface for a singleexecuteop/region [@ftynse ]

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -361,9 +361,30 @@ static inline bool isLoadMovable(Operation *op) {
...
+    }
+  }
+  return true;
+}
+
 // Whether or not an operation can be moved from the forward region to the
 // reverse region or vice-versa.
 static inline bool isMovable(Operation *op) {
+  if (auto regionBranch = dyn_cast<RegionBranchOpInterface>(op))
+    return isRegionBranchMovable(regionBranch);
```
wsmoses, line 395: make sure to also check the operands of ops within the region too [since those won't be operands of the region op]

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -361,9 +361,30 @@ static inline bool isLoadMovable(Operation *op) {
...
+  // we only consider non-nested and non-looping ops.
+  if (regionBranch.hasLoop())
+    return false;
+  for (auto &region : regionBranch->getRegions()) {
+    // Regions with multiple blocks potentially contain loops
+    if (!region.hasOneBlock())
+      return false;
+
+    for (auto &bodyOp : region.getOps()) {
+      if (!(bodyOp.getNumRegions() == 0 && isPure(&bodyOp))) {
```
wsmoses: should this recur on isMovable?

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
pengmai: I'm not sure that I see the interface you're talking about (while looking at the interfaces implemented by `scf.if`)

wsmoses: hm apparantly not, ignore me

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
pengmai: It can - the numRegions check should stay if we're explicitly choosing to ignore nested ops, and we need an extra check for terminators, but this would enable recomputing loads inside of if ops. I'll change it

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
pengmai: I'm not sure I follow. We're not checking any operands (of the region op or otherwise) while determining if we can move an op. `scf.if` in particular never has any region operands

wsmoses: yeah I moreso mean for the mincut movement workflow, normallyy we just look at the operands of an op. But if we move a whole op, we need to save its operands, but also any uses of operands with the op's region which are defined outside the op

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -361,9 +361,33 @@ static inline bool isLoadMovable(Operation *op) {
...
+  // we only consider non-nested and non-looping ops.
+  if (regionBranch.hasLoop())
+    return false;
+  for (auto &region : regionBranch->getRegions()) {
+    // Regions with multiple blocks potentially contain loops
+    if (!region.hasOneBlock())
+      return false;
+
+    for (auto &bodyOp : region.front()) {
+      if (!(bodyOp.getNumRegions() == 0 &&
```
wsmoses: this condition is confusing

To simplify this is the same, right?:

```
if numRegions != 0
   returnFalse

if op != terminator && !isMovable
  return False
```

also instead of the 
```
if numRegions != 0
   returnFalse
```
should we recur on this function?

wsmoses: [handle nested ifs]

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
pengmai: Gotcha, I believe I've made the necessary changes (and updated the test to check for this)

enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
pengmai: This is updated/simplified

wsmoses (approved):

PR #3025, opened 2026-07-28
Handle constant operands on active RegionBranch edges in forward mode

vimarsh6739:
This is a quick correctness argument

  Consider successor input $y_i$, with incoming value $v_{e,i}$ on edge $e$.
  Forward mode requires:
$$\dot{y_i} = \phi_{\forall_e} \left(  \dot{v_{e,i}} \right)$$
  The successor has one fixed argument signature, independent of which incoming
  edge is taken.

  - If every predecessor value is constant, $\dot{v_{e,i}} = 0$ on every edge, the shadow position can be omitted entirely.
  - If any predecessor is active, the successor requires a shadow argument. Consequently every incoming edge must provide that position:
    - an active predecessor supplies its computed tangent;
    - a constant predecessor supplies zero.

The implementation adds a zero after a constant primal operand only when the corresponding successor position needs a shadow, and adds the matching region block argument. Since the source operand is constant, no `invertedPointers` mapping is created for it.


vimarsh6739:
@wsmoses

PR #3030, opened 2026-07-29
mlir: loop-invariant cache for small stack allocas

pengmai:


enzyme/Enzyme/MLIR/Passes/RemovalUtils.cpp
```diff
@@ -1225,10 +1229,174 @@ void mlir::enzyme::minCutCache(Block *forward, Block *reverse,
...
 
+static LogicalResult loopInvariantCacheImpl(CacheInfo info,
+                                            LoopLikeOpInterface fwdLoop,
+                                            LoopLikeOpInterface revLoop,
+                                            IRMapping &mapping,
+                                            int64_t threshold) {
+  Operation *definingOp = info.pushedValue().getDefiningOp();
+  if (!definingOp)
+    return failure();
+  auto loadOp = dyn_cast<memref::LoadOp>(definingOp);
```
wsmoses, line 1243: we'll also want affine load here? [or some other load like?]

wsmoses (approved):
approved for now, but as mentioned on zoom we'll do some follow up work after next week to generalize

PR #3034, opened 2026-07-30
Remove deprecated use of make_scope_exit

boomanaiden154:
Wrap it in an ifdef. This was removed in upstream LLVM recently (289a4568f7d18d1d42687f3a2fc2c72bcbadaa28).


wsmoses:
honestly that's a simple enough function with one exit that we might as well just call gutils->deregisterCacheCreatorHook(cacheCreator) before return and avoid the whole make_scope issue entirely

PR #3036, opened 2026-07-31
mlir: Fix Binomial checkpointing schedule and support for mutable memory

Pangoraw:
The schedule was suboptimal resulting in poor memory / compute trade-offs. This is now better.

For mutable memory implement a few things:

- cloning pointers is memory space aware using a hack to call enzymexla.memref2pointer / enzymexla.pointer2memref and gpu.alloc /dealloc. Would be interesting to move these ops definition here instead of in E-JAX.
- When replaying the computation in the binomial scheme, the mutable memories clones are used but their shadow are the original shadows. This is valid since shadow mutation is only performed during the reverse part.
- Mutable refs referenced in the loop body should *not* alias each other, since otherwise the clones would be wrong (not alias). As such, a pass has been added to sink viewops to the loop body. 
- I have not looked at periodic checkpointing.

image


enzyme/Enzyme/MLIR/Dialect/LLVMExt/LLVMExtOps.td
```diff
@@ -6,9 +6,21 @@ include "mlir/Interfaces/SideEffectInterfaces.td"
 
 def LLVMPtr : Type<CPred<"::llvm::isa<::mlir::LLVM::LLVMPointerType>($_self)">>;
 
+// The memory space a pointer's underlying allocation lives in. Space 0 is host
+// memory, reachable by plain loads/stores and copyable with memcpy. A non-zero
+// space is device memory: it is only addressable by the accelerator, so
+// allocating, copying and freeing it all require runtime calls rather than
+// their host equivalents. Defaults to 0 so IR written before this attribute
+// existed keeps its host meaning.
+def MemorySpace : DefaultValuedAttr<I64Attr, "0">;
```
wsmoses, line 15: `

wsmoses: Long term we may want to fuse this with the alloc/free/memcpy ops in enzymexla, but a task for post monday

enzyme/Enzyme/MLIR/Dialect/LLVMExt/LLVMExtOps.td
```diff
@@ -6,9 +6,21 @@ include "mlir/Interfaces/SideEffectInterfaces.td"
...
+
 def AllocOp : LLVMExtOp<"alloc", []> {
   let summary = "Allocates memory";
-  let arguments = (ins I64 : $size);
+  let description = [{
+    Allocates `size` bytes in `memory_space`. Host allocations (space 0) lower
+    to `malloc`, device allocations lower to a `gpu.alloc` in that space.
+  }];
+  let arguments = (ins I64 : $size, MemorySpace : $memory_space);
```
wsmoses, line 23: why do we need this as an arg when llvm pointers have an address space in its type?

enzyme/Enzyme/MLIR/Dialect/LLVMExt/LLVMExtOps.td
```diff
@@ -18,17 +30,56 @@ def AllocOp : LLVMExtOp<"alloc", []> {
...
+    deallocator (or vice versa) corrupts the heap.
+
+    The attribute is optional because a free is usually emitted far from the
+    matching alloc -- reverse-mode AD frees a checkpoint buffer it recovered
+    from a cache, so the pointer's defining op is an `enzyme.pop` and the space
+    isn't apparent locally. When it is omitted, lower-llvm-ext recovers the
+    space by tracing the pointer back to its allocation, and fails if it can't:
+    guessing "host" would silently free device memory.
+  }];
+  let arguments = (ins Arg<LLVMPtr, "ptr to free", [MemFree]> : $ptr,
```
wsmoses, line 45: same comment here about type memory space / address space

enzyme/Enzyme/MLIR/Dialect/LLVMExt/LLVMExtOps.td
```diff
@@ -18,17 +30,56 @@ def AllocOp : LLVMExtOp<"alloc", []> {
...
+  let summary = "Copies memory between two pointers in a given memory space";
+  let description = [{
+    Copies `size` bytes from `src` to `dst`, both of which live in
+    `memory_space`. Host-to-host copies (space 0) lower to `llvm.memcpy`;
+    copies within a device space lower to a device-to-device runtime copy.
+  }];
+  let arguments = (ins Arg<LLVMPtr, "destination", [MemWrite]> : $dst,
+                       Arg<LLVMPtr, "source", [MemRead]> : $src,
+                       I64 : $size,
+                       MemorySpace : $memory_space);
```
wsmoses, line 64: and here

enzyme/Enzyme/MLIR/Passes/LowerLLVMExtPass.cpp
```diff
@@ -130,19 +427,57 @@ struct LowerLLVMExtPass
...
     for (auto alloc : allocs)
-      lowerAlloc(alloc, lowerToAllocaThreshold);
+      if (failed(lowerAlloc(alloc, lowerToAllocaThreshold)))
+        return signalPassFailure();
+
+    // After the allocs, so a memcpy whose alloca-promoted destination went away
+    // still sees a valid pointer.
+    SmallVector<llvm_ext::MemcpyOp> memcpys;
+    op->walk([&](llvm_ext::MemcpyOp memcpy) { memcpys.push_back(memcpy); });
+
```
wsmoses, line 465: this pass currently shares a ton with the enzymexla alloc/free/copy [though that does memrefs not pointers]. Can/should consolate after deadlines

PR #3051, opened 2026-08-03
Add pass to hoist allocations

Pangoraw:


wsmoses (approved):

enzyme/Enzyme/MLIR/Passes/HoistLoopAllocations.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+/// Only sequential loops qualify: hoisting out of a parallel loop gives every
+/// concurrent iteration the same buffer.
+///
+/// The allowlist is deliberate rather than a `HasParallelRegion` check --
+/// `affine.parallel` carries no such trait, so a trait-only test would silently
+/// accept it.
+static bool isSupportedLoop(Operation *op) {
+  return isa<scf::ForOp, affine::AffineForOp>(op) &&
+         !op->hasTrait<OpTrait::HasParallelRegion>();
```
wsmoses: if we have the isa allowlist we don't need the parallel check?

PR #3053, opened 2026-08-03
Add batch mode test

### LINKED ISSUE #3052: Demonstrate batching in Fortran
We should document how to use `__enzyme_batch` in Fortran. Unfortunately, it's not something we can straightforwardly provide an interface for because one of the arguments has to be passed by value. (The integer for the width.)

joewallwork:
Closes #3052
Merges into #3020 

Here is a Fortran version of the batch mode tutorial here: https://github.com/EnzymeAD/Enzyme-Tutorial/blob/main/8_batch/batch.c.

I'm not 100% sure it's actually using `__enzyme_batch`, though. In earlier iterations, I was making use of a derived type
```fortran
  type vector                                                                                                                                                                                                                         
    real :: x1, x2, x3, x4                                                                                                                                                                                                            
  end type vector
```
(the equivalent of the struct used in the C example), but the error messages I got about the numbers of arguments led me towards the version presented in this PR. Using `type(vector) :: y` for output instead of `real :: y1, y2, y3, y4` I get
```
error: <unknown>:0:0: in function _QQmain void (): Enzyme: __enzyme_batch missing vector argument at index 9, need argument of type ptr at call   call void @square__enzyme_batch_(ptr nonnull @_QMmath_modPsquare, i32 %12, i32 4, i32 %13, ptr nonnull %4, ptr nonnull %3, ptr nonnull %2, ptr nonnull %1, ptr nonnull %5)
```

Worth noting that I couldn't get this to work with `square` defined as a function (as in the other similar tests).

### DIFF
--- a/enzyme/Fortran/README.md
+++ b/enzyme/Fortran/README.md
@@ -36,7 +36,7 @@ $ flang -flto program-enzyme.bc -o program
 Both routes are exercised by the tests in `enzyme/test/Fortran`. The plugin route is
 flang-only; with ifx use the `opt` pipeline above.
 
-## Function hooks
+## Function hooks for differentiation
 
 We provide bindings for the `__enzyme_fwddiff` and `__enzyme_autodiff` function
 hooks using implicit interfaces. Some Fortran compilers disallow procedure names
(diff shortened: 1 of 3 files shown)


isaacaka:
I think the issue is that it is expecting 4 pointers, 1 for each of the inputs that it applies the square function to, but only sees 1 pointer to the entire struct when using type(vector) :: y

enzyme/Fortran/README.md
```diff
@@ -105,3 +105,22 @@ then you can make use of activity descriptors like so:
...
+
+We do not currently provide bindings for the `__enzyme_batch` function hook
+because it requires `enzyme_width` to be passed-by-value as an integer and this
+is not supported by the implicit interfacing approach used for the other
+function hooks. As such, you will need to write your own explicit `interface`
+block to handle the batching. See the Fortran
+[batching test ](../test/Fortran/BatchMode/square_with_explicit_interface.f90)
+for an example.
+
+> [!NOTE]
```
vchuravy, line 119: Is this normal markdown?

I am more used to admonition in the style of

```
!!! note
    In C...
```

vchuravy (approved):

enzyme/Fortran/README.md
joewallwork: It seems it's GitHub-specific (see https://blog.markdowntools.com/posts/markdown-admonitions-callouts-complete-guide), so renders nicely in the branch preview [here](https://github.com/EnzymeAD/Enzyme/tree/3052_fortran-batch-test/enzyme/Fortran) at least. I opened it in VSCode's Markdown preview mode and it makes it a callout but doesn't format the 'Note' like it does in GitHub. The one you mention is MkDocs I guess?

joewallwork:
Rebased on top of main after merging #3054

Issue #2822, opened 2026-05-11
Enzyme is not aware of Fortran descriptors


joewallwork:
Flang represents allocatable arrays using descriptor structs containing both differentiable and non-differentiable metadata but Enzyme isn't aware of this distinction. This means that it currently tries to differentiate with respect to implicit integer arguments that come with 'assumed shape' array arguments. This is the deeper issue underpinning #2820.

Making Enzyme aware of Fortran descriptor information would likely be greatly beneficial for improving the user experience in Fortran.

joewallwork:
Places we might need to modify Enzyme:
* Add to the `KnownInactiveFunctionsStartingWith` array in `enzyme/Enzyme/ActivityAnalysis.cpp`
* Say how to handle certain Fortran-specific strings by modifying `getTypeFromTBAAString` in `enzyme/Enzyme/TypeAnalysis/TBAA.h`
* Add Fortran-specific code in `enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp`

joewallwork:
A more recent recommendation was to edit `enzyme/Enzyme/CallDerivatives.cpp`.

PR #3054, opened 2026-08-03
Separate out Fortran tests at different optimisation levels

joewallwork:
Merges into #3020
Related to #2822

This PR gets us to a point of having no conditional scripting logic in the lit run commands for the Fortran tests by separating out the `-O0` tests from the rest. I noted in #3020 that for higher optimisation levels, Fortran intrinsics can already be handled fine because they don't appear in the LLVM IR.

The downside of this PR is further duplication between tests. However, with the proposed changes it is (in my opinion) much clearer what needs to be done to get things working in each case.


vchuravy:
Can you rebase first? Or is the base branch off?

joewallwork:
> Can you rebase first? Or is the base branch off?

The base was a few commits back because I didn't want to pick up the recent commits attempting to fix the IFX 2023 CI workflow. Force-pushed a rebased version now.

joewallwork:
The IFX failures here are the same as in #3020.

joewallwork:
The benchmark CI job failure seems to be unrelated.

vchuravy (approved):

joewallwork:
Rebased on top of `main` to fix conflicts after merging #3020.

PR #3072, opened 2026-08-05
Add clang enzyme_notypeanalysis and enzyme_ta_norecur attributes

markabate:
Adding clang attributes for the enzyme_notypeanalysis  and enzyme_ta_norecur flags used by Enzyme.

enzyme_notypeanalysis:
https://github.com/search?q=repo%3AEnzymeAD%2FEnzyme+enzyme_notypeanalysis&type=code

enzyme_ta_norecur:
https://github.com/search?q=repo%3AEnzymeAD%2FEnzyme+enzyme_ta_norecur&type=code

### DIFF
--- a/enzyme/Enzyme/Clang/EnzymeClang.cpp
+++ b/enzyme/Enzyme/Clang/EnzymeClang.cpp
@@ -319,18 +319,20 @@ static clang::FrontendPluginRegistry::Add<EnzymeAction<EnzymePlugin>>
 #if LLVM_VERSION_MAJOR > 10
 namespace {
 
+bool isGlobalDecl(const Decl *D) {
+  auto VD = dyn_cast<VarDecl>(D);
+  return VD && VD->hasGlobalStorage();
+}
+
 /// Shared check for the registration attributes which apply to both functions
 /// and global variables.
 bool appertainsToFunctionOrGlobal(Sema &S, const ParsedAttr &Attr,
                                   const Decl *D) {
-  if (isa<FunctionDecl>(D))
+  if (isa<FunctionDecl>(D) || isGlobalDecl(D))
     return true;
-  if (auto VD = dyn_cast<VarDecl>(D)) {
-    if (VD->hasGlobalStorage())
-      return true;
-  }
+
   S.Diag(Attr.getLoc(), diag::warn_attribute_wrong_decl_type_str)
-      << Attr << "functions and globals";
+      << Attr << " applies to functions and globals only";
   return false;
 }
 
@@ -681,71 +683,19 @@ struct EnzymeNoTypeAnalysisAttrInfo : public ParsedAttrInfo {
 
   bool diagAppertainsToDecl(Sema &S, const ParsedAttr &Attr,
                             const Decl *D) const override {
-    // Attribute appertains to functions
-    if (isa<FunctionDecl>(D))
-      return true;
-    S.Diag(Attr.getLoc(), diag::warn_attribute_wrong_decl_type_str)
-        << Attr << "appertains only to functions";
-    return false;
+    return appertainsToFunctionOrGlobal(S, Attr, D);
   }
 
   AttrHandling handleDeclAttribute(Sema &S, Decl *D,
                                    const ParsedAttr &Attr) const override {
-    if (Attr.getNumArgs() != 0) {
-      S.Diag(Attr.getLoc(), diag::err_attribute_too_many_arguments)
-        << Attr << 0;
-      return AttributeNotApplied;
-    }
-    D->addAttr(AnnotateAttr::Create(
-        S.Context, "enzyme_notypeanalysis", nullptr, 0, Attr.getRange()));
-    return AttributeApplied;
+    // For now enzyme::notypeanalysis corresponds to the internal attribute enzyme_ta_norecur
+    return handleEnzymeMarkerAttr(S, D, Attr, "enzyme_notypeanalysis",
+                                    /*FnAnnotation*/ "enzyme_ta_norecur",
+                                    /*VarAnnotation*/ "enzyme_ta_norecur");
   }
 };
 
 static ParsedAttrInfoRegistry::Add<EnzymeNoTypeAnalysisAttrInfo> enzyme_notypeanalysis("enzyme_notypeanalysis", "");
-
-struct EnzymeTANoRecurAttrInfo : public ParsedAttrInfo {
-  EnzymeTANoRecurAttrInfo() {
-    static constexpr Spelling S[] = {
-      {ParsedAttr::AS_GNU, "enzyme_ta_norecur"},
-#if LLVM_VERSION_MAJOR > 17
-      {ParsedAttr::AS_C23, "enzyme_ta_norecur"},
-#else
-      {ParsedAttr::AS_C2x, "enzyme_ta_norecur"},
-#endif
-      {ParsedAttr::AS_CXX11, "enzyme_ta_norecur"},
-      {ParsedAttr::AS_CXX11, "enzyme::ta_norecur"}
-    };
-    Spellings = S;
-  }
-
-  bool diagAppertainsToDecl(Sema &S, const ParsedAttr &Attr,
-                            const Decl *D) const override {
-    // Attribute appertains to functions and globals
-    if (isa<FunctionDecl>(D))
-      return true;
-    if (auto VD = dyn_cast<VarDecl>(D); VD->hasGlobalStorage())
-      return true;
-    S.Diag(Attr.getLoc(), diag::warn_attribute_wrong_decl_type_str)
-        << Attr << "appertains only to globals";
-    return false;
-  }
-
-  AttrHandling handleDeclAttribute(Sema &S, Decl *D,
-                                   const ParsedAttr &Attr) const override {
-    if (Attr.getNumArgs() != 0) {
-      S.Diag(Attr.getLoc(), diag::err_attribute_too_many_arguments)
-        << Attr << 0;
-      return AttributeNotApplied;
-    }
-    D->addAttr(AnnotateAttr::Create(
-        S.Context, "enzyme_ta_norecur", nullptr, 0, Attr.getRange()));
-    return AttributeApplied;
-  }
-};
-
-static ParsedAttrInfoRegistry::Add<EnzymeTANoRecurAttrInfo> enzyme_ta_norecur(
-    "enzyme_ta_norecur", "");
 
 } // namespace
 
--- a/enzyme/Enzyme/Clang/EnzymeClang.cpp
+++ b/enzyme/Enzyme/Clang/EnzymeClang.cpp
@@ -663,6 +663,90 @@ struct EnzymeSparseAccumulateAttrInfo : public ParsedAttrInfo {
 
 static ParsedAttrInfoRegistry::Add<EnzymeSparseAccumulateAttrInfo>
     SparseX("enzyme_sparse_accumulate", "");
+
+struct EnzymeNoTypeAnalysisAttrInfo : public ParsedAttrInfo {
+  EnzymeNoTypeAnalysisAttrInfo() {
+    static constexpr Spelling S[] = {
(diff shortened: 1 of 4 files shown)


enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -1150,6 +1150,7 @@ void TypeAnalyzer::updateAnalysis(Value *Val, TypeTree Data, Value *Origin) {
 
   if (auto GV = dyn_cast<GlobalVariable>(Val)) {
     if (hasMetadata(GV, "enzyme_ta_norecur"))
+      llvm:errs() << "Skipping updateAnalysis (enzyme_ta_norecur) " << GV->getName() << "\n";
```
ax3l: Codex:

[P1] Keep this return conditional on `enzyme_ta_norecur`

This inserted statement makes the following `return` fall outside the inner `if`, so `updateAnalysis` now returns for every `GlobalVariable`, including globals without the metadata. That prevents normal type propagation through all ordinary globals and can change generated derivatives. Please restore braces around the metadata condition (and remove the stray debug output) so only marked globals are skipped.

enzyme/Enzyme/Clang/EnzymeClang.cpp
```diff
@@ -851,6 +851,90 @@ struct EnzymeSparseAccumulateAttrInfo : public ParsedAttrInfo {
...
+    };
+    Spellings = S;
+  }
+
+  bool diagAppertainsToDecl(Sema &S, const ParsedAttr &Attr,
+                            const Decl *D) const override {
+    // Attribute appertains to functions and globals
+    if (isa<FunctionDecl>(D))
+      return true;
+    if (auto VD = dyn_cast<VarDecl>(D); VD->hasGlobalStorage())
```
ax3l: Codex:

[P2] Null-check the `dyn_cast` before dereferencing it

`diagAppertainsToDecl` is also called for unsupported declaration kinds. For a record or field, `dyn_cast<VarDecl>(D)` returns null and `VD->hasGlobalStorage()` dereferences it, crashing Clang instead of issuing the intended warning. Please guard `VD` first, following the existing attribute implementations above.

wsmoses:
@markabate can you rebase on main, I just landed an infra change [that among other things should fix the use of these within templates].

Also can we come up with a test, e.g. one where an internal function uses a union type or something

ax3l:
**_Codex:_**

Given [Billy’s union/template suggestion](https://github.com/EnzymeAD/Enzyme/pull/3072#issuecomment-5197148809) and the [MFIter construction site](https://github.com/BLAST-WarpX/gradient-harness/blob/main/src/amrex_heat_equation/main.cpp#L144), I’d add two lit tests.

[Background](https://github.com/BLAST-WarpX/gradient-harness/blob/main/src/amrex_heat_equation/main.cpp#L144): The AMReX detail worth preserving is that `MFIter` constructors are declared in `AMReX_MFIter.H` but defined out of line in `AMReX_MFIter.cpp`. A test containing only locally defined functions would miss that important path.

### 1. Clang attribute lowering test

Add `enzyme/test/Integration/ReverseMode/typeanalysis_attrs.cpp`.

It should cover:

- `enzyme_notypeanalysis` on a declaration-only constructor, modeling `MFIter`.
- An internal templated function containing mixed union storage, exercising the newly landed template infrastructure.
- `enzyme_ta_norecur` on a declaration-only function.
- `enzyme_ta_norecur` on a global.
- Both GNU and `[[enzyme::...]]` spellings.
- Final IR attributes/metadata after `PreserveNVVM`.

Sketch:

```cpp
// RUN: if [ %llvmver -ge 12 ]; then %clang++ -std=c++17 -O0 %s \
// RUN:   -S -emit-llvm -o - %newLoadClangEnzyme | FileCheck %s; fi

union IterState {
  double real;
  unsigned long long bits;
};

struct Iter {
  IterState state;

  // Declaration only, like amrex::MFIter.
  [[enzyme::notypeanalysis]]
  explicit Iter(int);
};

extern "C" void advance(IterState *)
    __attribute__((enzyme_ta_norecur));

extern IterState global_state
    __attribute__((enzyme_ta_norecur));

template <typename T>
[[enzyme::notypeanalysis]]
__attribute__((noinline))
static int opaque_union(T input) {
  union {
    T real;
    unsigned long long bits;
  } value;

  value.real = input;
  value.bits = 0;
  return static_cast<int>(value.bits);
}

double use(double x) {
  Iter iterator(1);
  advance(&global_state);
  return x + opaque_union(x);
}

// CHECK: @global_state = external global {{.*}}, !enzyme_ta_norecur
// CHECK: declare {{.*}}Iter{{.*}} #[[CTOR:[0-9]+]]
// CHECK: declare void @advance({{.*}}) #[[ADVANCE:[0-9]+]]
// CHECK: define internal {{.*}}opaque_union{{.*}} #[[TEMPLATE:[0-9]+]]
// CHECK-DAG: attributes #[[CTOR]] = { {{.*}}"enzyme_notypeanalysis"{{.*}} }
// CHECK-DAG: attributes #[[ADVANCE]] = { {{.*}}"enzyme_ta_norecur"{{.*}} }
// CHECK-DAG: attributes #[[TEMPLATE]] = { {{.*}}"enzyme_notypeanalysis"{{.*}} }
```

The exact mangled-constructor check should remain loose across LLVM versions.

This test also forces the rebased implementation to integrate the new attributes with `handleEnzymeMarkerAttr`, `enzymeRegistrationKind`, and the registration-global handling in `PreserveNVVM`. Otherwise the declaration-only constructor attribute will silently disappear—the precise failure relevant to AMReX.

### 2. Direct TypeAnalysis behavior test

Add `enzyme/test/TypeAnalysis/skipattrs.ll`.

Use two internal callees whose bodies access the same eight-byte storage as both `double` and `i64`:

- One with `"enzyme_notypeanalysis"`.
- One with `"enzyme_ta_norecur"`.

Have `@caller` invoke both, then run:

```llvm
; RUN: %opt < %s %newLoadEnzyme -passes="print-type-analysis" \
; RUN:   -type-analysis-func=caller -S -o /dev/null | FileCheck %s
```

The important checks are:

```llvm
; CHECK-LABEL: caller -
; CHECK-NOT: notype_union -
; CHECK-NOT: norecur_union -
```

That proves the attributes affect TypeAnalysis rather than merely appearing in emitted IR.

The same file should contain:

```llvm
@ordinary = internal global i64 0
@ignored = internal global i64 0, !enzyme_ta_norecur !0
```

Propagate a floating-point type through `@ordinary`, access both globals, and assert that:

- `@ordinary` still acquires `Float@double`.
- `@ignored` does not.

The unannotated control global specifically catches the missing-braces regression from the review: if `updateAnalysis` returns for every global, the `@ordinary` check fails.

### Optional diagnostic test

A small `-Xclang -verify` test should also include:

```cpp
struct [[enzyme::ta_norecur]] Invalid {};
// expected-warning: attribute only applies to functions and globals
```

This ensures invalid placement produces a diagnostic rather than dereferencing a null `dyn_cast<VarDecl>`.

markabate:
@wsmoses Do you have more information on the difference between enzyme_notyeanalysis and enzyme_ta_norecur, to help with the unit tests?

My understanding is enzyme_notypeanalysis is only applied to functions and exits from the visitCallBase function [here](https://github.com/EnzymeAD/Enzyme/blob/84c6fae0fc907757c318164fc8ac86281fdd655e/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L5390). But visitCallBase only seems to be used with Julia types [here](https://github.com/EnzymeAD/Enzyme/blob/84c6fae0fc907757c318164fc8ac86281fdd655e/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L1590).

enzyme_ta_norecur seems to exit you from the [updateAnalysis loop](https://github.com/EnzymeAD/Enzyme/blob/84c6fae0fc907757c318164fc8ac86281fdd655e/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L1156).

wsmoses:
visitCallBase is called by call call instructions [that's just an explicit dispatch].

wsmoses:
and honestly I forgot enzyme_notypeanalysis existed and it should be replaced with the standard enzyme_ta_norecur in all cases

wsmoses:
note we can call the c++ attribute enzyme::notypeanalysis or something still

markabate:
@wsmoses @ax3l I added a unit test and removed the clang attribute I was using to set the enzyme_notypeanalysis internal function attribute - so now there's just one clang attribute called enzyme_notypeanalysis, which sets enzyme_ta_norecur.

If you remove \_\_attribute\_\_((enzyme_notypeanalysis)) from the unit test, you get the updateAnalysis error we've been seeing:

`global/u2/m/mabate/src/Enzyme/enzyme/test/Integration/ReverseMode/notypeanalysis.c:16:8: error: Enzyme: <analysis>
  %3 = alloca double, align 8: {[-1]:Pointer, [-1,-1]:Float@double}, intvals: {}
  %8 = fdiv double %6, %7: {[-1]:Float@double}, intvals: {}
  %14 = fadd double %11, %13: {[-1]:Float@double}, intvals: {}
  %5 = alloca double, align 8: {[-1]:Pointer, [-1,-1]:Float@double}, intvals: {}
double %0: {[-1]:Float@double}, intvals: {}
double %1: {[-1]:Float@double}, intvals: {}
  %4 = alloca double, align 8: {[-1]:Pointer, [-1,-1]:Float@double}, intvals: {}
  %6 = load double, ptr %3, align 8: {[-1]:Float@double}, intvals: {}
  %7 = load double, ptr %4, align 8: {[-1]:Float@double}, intvals: {}
  %9 = load double, ptr %3, align 8: {[-1]:Float@double}, intvals: {}
@dPtr = internal global %union.DoublePtr zeroinitializer, align 8: {[-1]:Pointer, [-1,-1]:Pointer, [-1,-1,0]:Float@double}, intvals: {}
  %10 = load ptr, ptr @dPtr, align 8: {[-1]:Pointer, [-1,0]:Float@double}, intvals: {}
  %11 = load double, ptr %5, align 8: {[-1]:Float@double}, intvals: {}
  %12 = load i64, ptr @dPtr, align 8: {[-1]:Integer}, intvals: {}
  %13 = sitofp i64 %12 to double: {[-1]:Float@double}, intvals: {}
</analysis>
Illegal updateAnalysis prev:{[-1]:Pointer, [-1,-1]:Pointer, [-1,-1,0]:Float@double} new: {[-1,-1]:Integer}
val: @dPtr = internal global %union.DoublePtr zeroinitializer, align 8 origin=  %12 = load i64, ptr @dPtr, align 8
   16 | double divide(double x, double y) {
`

I'm still wondering if there's a way Enzyme could gracefully handle situations like this. Could Enzyme leave some types undecided until the activity analysis happens? We can also discuss more at the next design meeting.

wsmoses (approved):

wsmoses:
yeah there are ways we can try to improve support even without this -- but let's get in one step at a time

wsmoses:
also @markabate can you fix the format and remove the old enzyme_notypeanalysis references?

markabate:
@wsmoses It looks like the enzyme_notypeanalysis annotation was only used in TraceInterface and TraceGenerator:

https://github.com/EnzymeAD/Enzyme/blob/33c495b1c04968d81e5e5705a5157fc064d1f885/enzyme/Enzyme/TraceInterface.cpp#L205-L217
https://github.com/EnzymeAD/Enzyme/blob/33c495b1c04968d81e5e5705a5157fc064d1f885/enzyme/Enzyme/TraceGenerator.cpp#L173
https://github.com/EnzymeAD/Enzyme/blob/33c495b1c04968d81e5e5705a5157fc064d1f885/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp#L5390

Do you think this is safe to remove?

markabate:
@wsmoses Would you mind merging this, or do you think we should still remove the `enzyme_notypeanalysis` code I linked above?

PR #3079, opened 2026-08-05
Differentiate an llvm.call

wsmoses:
Differentiating a call is the same work whichever dialect spelled it, but only `func::CallOp` had `AutoDiffCallFwd`/`AutoDiffCallRev` attached, so an `llvm.call` stopped at:

```
error: could not compute the adjoint for this operation
  "llvm.call"(%2, %1) <{callee = @_ZNK4mfem6future14tensor_ndarrayIdLi1EJEE10set_layoutESt5arrayImLm1EE, ...}>
```

MFEM's unit tests are *all* `llvm.call` — 213 of them in the one translation unit — and get through on the ones activity analysis rules out. The first that mattered stopped the build.

What differs between the dialects is how a call names its callee and how a call to the derivative is written, and both are already asked through interfaces: `CallOpInterface` for the first and `AutoDiffFunctionInterface::createCall` for the second (`AutoDiffLLVMFuncOpFunctionInterface` already implements it). So the models move to `CallAutoDiffImplementations`, written once over the call op type, and are attached from func and from LLVM.

## Three things that only this path reaches

**An alloca can say its own extent.** The reverse pass keeps a copy of a pointer argument, and `findPtrExtent` could only size one that `llvm_ext.alloc` made or that carried a `ptr_size_hint`. Everything else got `cannot find size of ptr` and a null from `cloneValue`, which the caller then dereferenced. An alloca says how much it is — the size of the element type, that many times — so it no longer has to be told.

**The pass builds `llvm_ext` ops without declaring the dialect.** That copy is written as `llvm_ext.alloc`/`llvm_ext.memcpy`, and `DifferentiatePass` never listed `llvm_ext` among its dependent dialects. A pipeline that raised through `llvm_ext` has it loaded already, which is why only running the pass on its own ever noticed (`llvm_ext.alloc created with unregistered dialect`).

**`insertInit`/`getCacheType` read out of an empty block.** They took a location from `initializationBlock->rbegin()` and a context from `->begin()`, both of which read off the end while the block is still empty. The location can be unknown, and the context is the type's own.

## Test

`test/MLIR/ReverseMode/llvm_call.mlir` — an `llvm.call` to a function that scales through a pointer, with an alloca as the argument. The alloca's 8 bytes are worked out, the copy is made, and the reverse calls `@diffescale` with it.

161/162 of `check-enzymemlir` pass; the one failure is `ReverseMode/linalg.mlir`, which fails on main too.

## What this does not fix

The MFEM file this came from — `tests/unit/enzyme/test_enzyme_reverse_tape.cpp` (EnzymeAD/Enzyme-JAX#2778) — still does not build. The adjoint is now produced, and the failure moves to a later one: the reverse function generated for that callee ends in a `llvm.intr.trap` branch, and the reverse-CFG branch caches for that shape

```mlir
%4 = "enzyme.init"() : () -> !enzyme.Cache<i32>
"enzyme.push"(%4, %c-1_i32) : (!enzyme.Cache<i32>, i32) -> ()
^bb1:
  llvm.intr.trap
  %9 = "enzyme.pop"(%4) : (!enzyme.Cache<i32>) -> i32   // unused
```

survive every cache-lowering pass and reach `failed to translate MLIR to LLVM IR`. Those are Enzyme's own control-flow caches, not the argument caches added here, and dead in this shape. Worth a separate look.

So this is a strict improvement — nothing that built before stops building — but it is not by itself enough for that file.


enzyme/Enzyme/MLIR/Implementations/CallAutoDiffImplementations.cpp
```diff
@@ -0,0 +1,228 @@
...
+      gutils->postpasses, gutils->verifyPostPasses, gutils->strongZero,
+      /*markReadonly=*/false);
+
+  SmallVector<Value> revArguments;
+
+  for (auto [arg, act, cache] :
+       llvm::zip_equal(orig->getOperands(), ArgActivity, caches)) {
+    revArguments.push_back(gutils->popCache(cache, builder));
+    if (act == DIFFE_TYPE::DUP_ARG)
+      revArguments.push_back(gutils->invertPointerM(arg, builder));
```
Pangoraw: The current fn call impl is wrong in the sense that these shadows need to go through caches as well. consider `f(float* p)` where `p` is a gep not dominating the reverse.

we can fix this in a follow-up.

PR #3081, opened 2026-08-05
Route getFirstNonPHI through a Utils.h wrapper

vchuravy:
LLVM removed the deprecated BasicBlock::getFirstNonPHI (llvm-project
62c5ede9fd14), leaving only the iterator-flavoured getFirstNonPHIIt.
Add a getFirstNonPHI(BasicBlock*) shim alongside the existing
getFirstNonPHIOrDbg wrappers and move the 19 call sites onto it. The
iterator overload has been available since LLVM 18, so older LLVMs keep
the old spelling.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


wsmoses (commented):
can you ask claude to add this as a lint check so we don't have this recur

wsmoses (approved):

PR #3092, opened 2026-08-05
mlir: unify checkpointing implementations

Pangoraw:
as a claim from the paper


wsmoses (approved):
didn't look too closely, but I'll assume its up to your level of happiness [and works]

PR #3097, opened 2026-08-06
Say an unknown summary set the way the readers of summaries read it

wsmoses:
`serializeSetNaive` says an unknown or undefined set as the bare marker string, and every reader of these summaries looks for exactly that. `serializeMapOfSetsNaive` said it as a one-element list instead:

```mlir
[#distinct, ["<unknown>"]]     rather than     [#distinct, "<unknown>"]
```

So the marker branch in both `deserializePointsTo` functions was dead, and the list read as a set of one known element. That element, a `StringAttr`, was then `cast` unchecked to `DistinctAttr` / `OriginAttr` and its arg number used as an index — `argNumber=336934464` into a six-element vector, in MFEM's dFEM tests, which segfaulted three of them (EnzymeAD/Enzyme-JAX#2778):

```
#4  mlir::enzyme::SetLattice<mlir::enzyme::OriginAttr>::join(...)
#5  mlir::enzyme::DenseActivityAnnotationAnalysis::processCallToSummarizedFunc(...)
#6  mlir::enzyme::DenseActivityAnnotationAnalysis::visitCallControlFlowTransfer(...)
```

One of the two readers was also comparing against `"unknown"`/`"undefined"` without the angle brackets, so even given the right shape it would not have recognised them.

Reading the marker correctly then hands the consumers an unknown set where they had only ever seen defined ones, and `getElements` asserts on those. So the three places that translate a callee's summary to its caller now answer it:

- forward dense origins: unknown origins stay unknown
- backward dense origins: memory whose sinks the callee could not name leaves every one of the caller's classes a candidate
- points-to: an argument the callee could not track leaves the caller pointing to unknown

Dropping it would have been the unsound direction — "came from nowhere" reads as inactive.

Test is `test/MLIR/ActivityAnalysis/Summaries/unknown.mlir`; it fails on `main` with the assertion (release: the segfault) and checks both that the marker round-trips and that the caller inherits the unknown rather than an empty set.

### DIFF
--- a/enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
+++ b/enzyme/Enzyme/MLIR/Analysis/ActivityAnnotations.cpp
@@ -502,10 +502,10 @@ static void deserializePointsTo(
     auto pointer = cast<DistinctAttr>(pair[0]);
     auto pointsToSet = enzyme::ValueOriginSet::getUndefined();
     if (auto strAttr = dyn_cast<StringAttr>(pair[1])) {
-      if (strAttr.getValue() == "unknown") {
+      if (strAttr.getValue() == enzyme::unknownSetString) {
         (void)pointsToSet.markUnknown();
       } else {
-        assert(strAttr.getValue() == "undefined" &&
+        assert(strAttr.getValue() == enzyme::undefinedSetString &&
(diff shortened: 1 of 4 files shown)


wsmoses:
@pengmai @ftynse can you review this in post

PR #3110, opened 2026-08-08
PreserveNVVM: let the math preservation skip the linkage promotion

wsmoses:
Preserving a libdevice definition does two things: `noinline`, so calls survive recognizably to Enzyme, and a promotion to external linkage, so nothing internalizes or drops the definition first. A consumer whose pipeline keeps the definitions alive its own way needs only the first — Reactant's raising consumes the math calls itself — and under it the promotion is a cost: every translation unit exports a strong copy of a formerly-internal function.

`PreserveNVVMNewPM` and `createPreserveNVVMPass` gain `PromoteMathLinkage` (default true; only the libdevice/half-wrapper sections consult it — custom derivatives and inactive-fn registrations promote as before). The `Begin=false` restore now checks `prev_linkage` is present before reading it, and takes the attribute away once restored. Exposed as `preserve-nvvm-inline-only`; the new lit test pins all three behaviors (promote, inline-only, restore).

Companion to EnzymeAD/Reactant#67.

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -3480,6 +3480,11 @@ extern "C" void registerEnzymeAndPassPipeline(llvm::PassBuilder &PB,
           MPM.addPass(PreserveNVVMNewPM(/*Begin*/ true));
           return true;
         }
+        if (Name == "preserve-nvvm-inline-only") {
+          MPM.addPass(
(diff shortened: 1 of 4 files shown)


wsmoses:
Withdrawn per review — the begin=false equivalent moves into the Reactant/Enzyme-JAX pipeline instead of an upstream knob.

PR #3121, opened 2026-08-09
MLIR: honor enzyme_dupnoneed by eliding primal stores

### LINKED ISSUE #3120: MLIR forward mode does not elide primal stores for enzyme_dupnoneed (dfem derivative kernels do ~50% dead work)
## MLIR forward mode does not elide primal work for `enzyme_dupnoneed`

MFEM's dfem derivative actions pass every output as `enzyme_dupnoneed` (fem/dfem/backends/util.hpp:745) — the caller wants only the tangent. LLVM-Enzyme omits the primal stores (and lets the primal compute chain DCE) for DUP_NONEED. The MLIR path parses the marker correctly into the op's activity, but forward mode then treats it as plain `dup`:

```mlir
func.func @f(%x: f64, %out: memref<f64>) {
  %sq = arith.mulf %x, %x : f64
  memref.store %sq, %out[] : memref<f64>
  return
}
```
`--enzyme-wrap="infn=f outfn= argTys=enzyme_dup,enzyme_dupnoneed retTys= mode=ForwardMode"` produces:
```mlir
    %2 = arith.addf ...            // tangent
    %3 = arith.mulf %arg0, %arg0   // primal — licensed to be omitted
    memref.store %2, %arg3[]       // tangent store (correct)
    memref.store %3, %arg2[]       // primal store — licensed to be omitted
```

## Why it matters (measured)

In MFEM's `bench_dfem_linearized_simple`, the fwddiff'd diffusion qfunc costs 3.0× its primal (the theoretical floor for primal+tangent is ~2×, and the interleaved liveness causes 106 stack spills in the 2D kernel). The primal half — a full 3×3 `inv`/`det` chain per quadrature point — is exactly the dead work `dupnoneed` should remove; eliding it should take the tangent kernel toward ~1× primal cost (only the tangent chain remains) and cut the register pressure with it. The derivative-action path of every dfem operator hits this.

## Implementation sketch

Forward mode clones the primal function and inserts tangents per op. For DUP_NONEED the store handlers (LLVM/memref/affine `createForwardModeTangent`) should erase the cloned primal store when the destination pointer traces (through casts, e.g. `enzymexla.pointer2memref`) to a DUP_NONEED entry argument; the primal compute chain then dies by DCE. `MGradientUtils` carries the per-argument activity; the base-pointer chase is the same walk the shadow lookup already does.

wsmoses:
Fixes #3120.

MLIR forward and reverse mode ignored `enzyme_dupnoneed`: every tangent/adjoint function recomputed and stored the full primal alongside the derivative, making the derivative pay for a primal the caller declared it will not read. (LLVM-Enzyme already exploits this; measured on MFEM dFEM it is why the MLIR tangent kernel cost ~3x its own primal.)

**Base-object helpers on `MGradientUtils`** (`GradientUtils.{h,cpp}`):
- `getBaseObject`: peels `ViewLikeOpInterface` ops and LLVM GEP/bitcast/addrspacecast, analogous to LLVM's `getUnderlyingObject`. Going through `ViewLikeOpInterface` means downstream view-like casts (e.g. Enzyme-JAX's `enzymexla.pointer2memref`, which already implements it) are peeled without this repo needing to know them — no new interface required.
- `getDiffeTypeOfBase`: the activity the caller declared for that object (entry block argument → `ArgDiffeTypes`).
- `mayReadBase` (cached): does anything in the function possibly read the object; calls count as reads of their pointer operands (callee-side elision is handled by activity forwarding instead).
- `primalStoreElidable(ptr)` = declared `DUP_NONEED` **and** never read back.

**Forward mode**: `memoryIdentityForwardHandler` — which the tblgen `MemoryIdentityOp` route funnels `llvm.store`/`memref.store`/`affine.store` through — erases the primal store clone when all written pointer operands are elidable. `llvm.intr.memset` keeps only its shadow clear.

**Reverse mode**: the store reverse handlers of all three dialects skip the augmented-forward primal store under the same predicate; the adjoint reads only the shadow.

**Calls**: forward and reverse call handlers forward `DUP_NONEED` into `CreateForwardDiff`/`CreateReverseDiff` for pointer arguments whose base was declared unneeded, so callees can elide their own stores. Call sites still pass primal+shadow — dupnoneed changes what may be skipped, not the signature — which required treating `DUP_NONEED` like `DUP_ARG` in the call-site operand assembly and reverse cache pop/free.

**Tests**: `ForwardMode/dupnoneed_store.mlir` (memref/llvm/affine/memset + through-cast + negative read-back case), `ForwardMode/dupnoneed_call.mlir` (forwarding through a call, with a cast in between), `ReverseMode/dupnoneed_store.mlir` (all three dialects). Full `test/MLIR` sweep: no regressions vs base.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
@@ -1010,6 +1010,12 @@ struct AffineStoreOpInterfaceReverse
             zeroStoreOp->setAttr("alignment", alignAttr);
         }
       }
+
+      // A store into memory whose primal contents the caller declared
+      // unneeded (enzyme_dupnoneed) need not happen in the augmented
+      // forward pass either.
+      if (gutils->primalStoreElidable(storeOp.getMemref()))
+        gutils->erase(gutils->getNewFromOriginal(op));
     }
     return success();
   }
--- a/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
@@ -256,6 +256,13 @@ LogicalResult mlir::enzyme::detail::memoryIdentityForwardHandler(
     gutils->setDiffe(oval, sval, builder);
   }
 
+  // A store into memory whose primal contents the caller declared unneeded
+  // (enzyme_dupnoneed) need not happen: the shadow store above is the whole
+  // of the derivative.
+  if (auto store = dyn_cast<enzyme::StoreLikeInterface>(orig))
+    if (gutils->primalStoreElidable(store.getStoredPointer()))
+      gutils->erase(primal);
+
   return success();
 }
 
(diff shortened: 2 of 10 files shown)


wsmoses:
Amended per review: the forward elision in the generic memory-identity handler now keys on `enzyme::StoreLikeInterface` (`getStoredPointer`) instead of scanning for pointer-typed operands on zero-result ops — all three store dialects already attach it. Behavior unchanged; tests unchanged and passing.

wsmoses:
Simplified per review: dropped the `mayReadBase` walk — declaring a pointer `enzyme_dupnoneed` is the caller saying nothing (the function's own reads included) needs the primal contents of that memory, so `primalStoreElidable` is now just `getDiffeTypeOfBase(ptr) == DUP_NONEED`. `ReverseMode/memref.mlir`'s `subview_in_loop` case (accumulation through a dupnoneed output) now elides its entire forward accumulation loop — CHECK lines updated; the read-back there fed only the unneeded primal, and the adjoint reads only the shadow.

wsmoses:
End-to-end validation on MFEM (raising path, `REACTANT_LOWER_INVOKE=0`): full rebuild of punit + benchmarks with this branch under Enzyme-JAX — **punit 103/103 (88,493 assertions)**, identical to baseline. `bench_dfem_linearized_simple -o 3`: plain-codegen and Enzyme paths unchanged-to-slightly-better vs vanilla clang (local derivative ~0.93–1.00× quad, ~0.98× hex); the elision is correctness-neutral on the suite. The dfem forward tangent's remaining deriv/fwd gap is register pressure in the tangent kernel, not primal stores — tracked separately.

PR #3124, opened 2026-08-10
Use original function's LoopInfo in legalRecompute

### LINKED ISSUE #3123: ParentPtr == BBParent && "loop info queried with block of other function"
explorer is too outdated to reproduce it.

@wsmoses 
`opt repro.ll -load-pass-plugin=enzyme23.so -passes="enzyme" -enzyme-strict-aliasing=0`

```llvm

@enzyme_dupnoneed = external global ptr

define fastcc void @inner(ptr %0, ptr %1, i1 %2) {
  br label %4

4:                                                ; preds = %15, %3
  %5 = load i32, ptr %1, align 4
  br label %7

6:                                                ; preds = %15
  tail call void %1(ptr null, ptr null, ptr null, ptr null)
  ret void

7:                                                ; preds = %7, %4
  %8 = phi i64 [ 0, %4 ], [ %12, %7 ]
  %9 = load double, ptr null, align 8
  %10 = load double, ptr %0, align 8
  %11 = fmul double %9, %10
  store double %11, ptr %0, align 8
  %12 = add i64 %8, 1
  %13 = zext i32 %5 to i64
  %14 = icmp eq i64 %8, %13
  br i1 %14, label %15, label %7

15:                                               ; preds = %7
  br i1 %2, label %4, label %6
}

define void @f(ptr %0) {
  call fastcc void @inner(ptr %0, ptr null, i1 false)
  ret void
}

define {} @entry() {
  %1 = tail call {} (...) @__enzyme_autodiff(ptr @f, ptr @enzyme_dupnoneed, ptr null, ptr null)
  ret {} %1
}

declare {} @__enzyme_autodiff(...)
```

ZuseZ4:
GradientUtils::legalRecompute maps `origStart` and the load back into the original function via isOriginal, and the dominance check on the line just above already uses OrigDT -- but the allInstructionsBetween call that follows passed CacheUtility's `LI`, which is LoopInfo over newFunc. Querying it with an oldFunc block asserts in LoopInfoBase::verifyBlockNumberEpoch:

  ParentPtr == BBParent && "loop info queried with block of other function"

All ten other allInstructionsBetween call sites already pass *OrigLI.

Found while differentiating a Rust integral kernel: an inner loop whose trip count is loaded from memory the loop itself may clobber is uncacheable, so the load is recomputed, and that recompute reaches legalRecompute with a forward-mode builder.

fixes https://github.com/EnzymeAD/Enzyme/issues/3123
@wsmoses

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -4173,8 +4173,8 @@ bool GradientUtils::legalRecompute(const Value *val,
               bool failed = false;
 
               allInstructionsBetween(
-                  const_cast<GradientUtils *>(this)->LI, origStart,
-                  const_cast<Instruction *>(orig), [&](Instruction *I) -> bool {
+                  *OrigLI, origStart, const_cast<Instruction *>(orig),
+                  [&](Instruction *I) -> bool {
                     if (I->mayWriteToMemory() &&
                         writesToMemoryReadBy(
                             &TR, *OrigAA, TLI,
--- a/enzyme/test/Enzyme/ReverseMode/legalrecompute-origli.ll
+++ b/enzyme/test/Enzyme/ReverseMode/legalrecompute-origli.ll
@@ -1,0 +1,69 @@
+; RUN: if [ %llvmver -ge 15 ]; then %opt < %s %OPnewLoadEnzyme -passes="enzyme,function(mem2reg,instsimplify,%simplifycfg)" -enzyme-preopt=false -S | FileCheck %s; fi
+
+; GradientUtils::legalRecompute maps `origStart` and the load back into the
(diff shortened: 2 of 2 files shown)


enzyme/Enzyme/GradientUtils.cpp
```diff
@@ -4172,8 +4172,14 @@ bool GradientUtils::legalRecompute(const Value *val,
             if (OrigDT->dominates(origStart, const_cast<Instruction *>(orig))) {
               bool failed = false;
 
+              // `origStart` and `orig` were both mapped back into oldFunc just
```
wsmoses: fix the format and remove the comment, otherwise lgtm

wsmoses (approved):

Issue #3029, opened 2026-07-29
PHI nodes not grouped at top of basic block


ZuseZ4:
https://tinyurl.com/24tvv237
```llvm
define double @outer(ptr %0) {
  call void @loops(ptr %0, i1 false)
  ret double 0.000000e+00
}

define void @leaf(i32 %0) {
  store double 0.000000e+00, ptr null, align 8
  ret void
}

define void @loops(ptr %0, i1 %1) {
  %3 = load i32, ptr %0, align 4
  br label %4

4:                                                ; preds = %8, %2
  %5 = phi i32 [ %10, %8 ], [ %3, %2 ]
  %6 = phi i32 [ %5, %8 ], [ 0, %2 ]
  %7 = phi i32 [ %9, %8 ], [ 0, %2 ]
  br i1 %1, label %8, label %11

8:                                                ; preds = %11, %4
  %9 = phi i32 [ %3, %11 ], [ %7, %4 ]
  %10 = add i32 %5, 1
  br label %4

11:                                               ; preds = %4
  call void @leaf(i32 %7)
  call void @leaf(i32 %6)
  br label %8
}

define double @entry() {
  %1 = tail call double (...) @__enzyme_autodiff(ptr @outer, ptr null, ptr null)
  ret double %1
}

declare double @__enzyme_autodiff(...)
```

`opt input.ll -load-pass-plugin=enzyme.so -passes=enzyme -enzyme-loose-types=1 -enzyme-strict-aliasing=0` 

@wsmoses  another repetition of this error.

PR #3125, opened 2026-08-10
Use hasTerminator() in freeCache

ZuseZ4:
Noticed this was missing @wsmoses 

LLVM's BasicBlock::getTerminator() asserts hasTerminator() rather than returning null, so the null-check idiom aborts with

  BasicBlock.h: getTerminator:
  `hasTerminator() && "cannot get terminator of non-well-formed block"'

whenever freeCache runs on a reverse block that is still under construction and so legitimately has no terminator yet. Reached via forceAugmentedReturns -> getContext -> getDynamicLoopLimit -> createCacheForScope, differentiating a loop whose trip count cannot be derived statically.

This was the last remaining use of the old idiom; the other sites already go through the Utils.h hasTerminator() wrapper, which this now matches. The preceding size() check becomes redundant, since hasTerminator() already implies a non-empty block.

### DIFF
--- a/enzyme/Enzyme/DiffeGradientUtils.cpp
+++ b/enzyme/Enzyme/DiffeGradientUtils.cpp
@@ -915,8 +915,7 @@ CallInst *DiffeGradientUtils::freeCache(BasicBlock *forwardPreheader,
   tbuild.setFastMathFlags(getFast());
 
   // ensure we are before the terminator if it exists
-  if (tbuild.GetInsertBlock()->size() &&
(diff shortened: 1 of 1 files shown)


wsmoses (approved):
I'm surprised our linter didn't find this [it's supposed to]

ZuseZ4:
I can add an integration test which triggered the last ~10 bugs once at least https://github.com/EnzymeAD/Enzyme/issues/3029 is fixed (so it builds again)

wsmoses:
sure!

PR #3132, opened 2026-08-10
MLIR: memcpy derivatives and shadow stores of pointers

wsmoses:
Differentiating a CUDA kernel launch through the raising pipeline walks
the capture struct the launcher packs its arguments into: active data
pointers are stored into alloca'd structs, the structs are memcpy'd
together, and the kernel reads the pointers back out. None of that had
a derivative registered, so both modes stopped at 'could not compute
the adjoint'.

Forward mode: the tangent of a memcpy is a memcpy of the shadows --
float bytes carry their tangents, pointer bytes their shadow pointers,
one copy serves both. A source nothing differentiates has no shadow to
copy from: float-free copies (destination provably an alloca of a
float-free type) take the primal bytes so structural fields stay usable
through the shadow object; otherwise the shadow is cleared, a zero
tangent for wherever the floats sit.

Reverse mode: a float-free copy only relocates pointers and integers,
whose derivative story is structural: one copy of the shadow bytes in
the forward sweep puts the shadow pointers at the same offsets, and the
reverse sweep has nothing to accumulate. Copies that may hold floats
still refuse, now saying why.

The same structural story applies to llvm.store of a mutable value:
the shadow memory must hold the shadow pointer, which the forward sweep
now writes; previously nothing did, so a shadow struct read back a null
data pointer.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/FuncAutoDiffOpInterfaceImpl.cpp
@@ -36,8 +36,16 @@ public:
   void transformResultTypes(Operation *self,
                             SmallVectorImpl<Type> &types) const {}
 
-  // A func.func carries no linkage of its own for the clone to have inherited.
-  void detachFromPrimalDefinition(Operation *self) const {}
+  // A func.func carries no linkage of its own, but one raised from an
+  // llvm.func still holds the primal's comdat as a plain attribute, and the
+  // clone inherits it. The derivative is not part of the primal's
+  // deduplication group (see the llvm.func model above for why that group is
+  // wrong for it), and a func.func cannot express a comdat anyway -- what
+  // remains is a dangling nested symbol reference that breaks symbol-use
+  // walks such as gpu-kernel-outlining's. Drop it.
+  void detachFromPrimalDefinition(Operation *self) const {
+    self->removeAttr("comdat");
+  }
 
   Operation *createCall(Operation *self, OpBuilder &builder, Location loc,
                         ValueRange args) const {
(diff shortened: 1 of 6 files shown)


wsmoses:
Split per review: this PR now carries only the forward-mode memcpy tangent (plus the func.func comdat detach). The reverse-mode memcpy rule is dropped entirely — doing it right needs type analysis to know where the floats sit in the copied bytes, so reverse mode keeps refusing for now. The shadow store of pointers moved to its own PR.

wsmoses:
Re where the func.func gets a comdat: it's round-trip baggage, not an error. Enzyme-JAX's llvm-to-memref-access rebuilds llvm.funcs as func.func (to memref-type their arguments) and copies the entire attribute dictionary — `linkage`, `CConv`, `passthrough`, `target_cpu`, and `comdat` all ride along — because convert-polygeist-to-llvm restores them when converting back to llvm.func (it consumes `linkage`/`CConv` explicitly and the rest, comdat included, pass through into the recreated llvm.func). So the *primal* legitimately keeps its comdat across the round trip; only the derivative clone must not inherit the primal's group key — the same situation the llvm.func model's `detachFromPrimalDefinition` already handles by re-keying. These func.func derivatives are internal linkage, where a comdat buys nothing, so the func.func model just drops it.

wsmoses:
Addressed all three follow-ups:
- `getUnderlyingAllocaType` is now one line over `oputils::getBaseObject` (this stacks the PR on #3133, whose commits appear in the diff until it merges).
- An unresolvable destination type now refuses conservatively instead of guessing; the float-holding case keeps the memset-zero behavior but emits a warning that the byte meaning is assumed, as the TODO marker for future type analysis.
- A memcpy into dupnoneed'd memory now elides the primal copy, like stores and memsets already did.

PR #3134, opened 2026-08-10
MLIR: derivatives for enzyme.affine_atomic_rmw

wsmoses:
The op is emitted as the derivative of affine loads under atomic-add
accumulation, but appearing in a primal -- a kernel that itself
accumulates with atomics, as MFEM's dfem hyperelasticity tests do --
neither mode had a rule for it and differentiation stopped.

Reverse: the location's adjoint serves both reads. The added value takes
it (d_v += dm[i], a plain load: the reverse sweep mirrors the forward
structure, so the barriers that ordered the primal's atomics order the
shadow's), and it keeps flowing to whatever wrote the location before.
A used result read the pre-add value, so its adjoint joins the
location's, atomically for the same reason the primal add was atomic.

Forward: the tangent of the added value joins the shadow location
through the same atomic add, whose result carries the tangent of the
value the primal read out.

Non-add kinds still refuse, now saying why.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/EnzymeAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/EnzymeAutoDiffOpInterfaceImpl.cpp
@@ -20,6 +20,7 @@
 
 #include "Dialect/Dialect.h"
 #include "Dialect/Ops.h"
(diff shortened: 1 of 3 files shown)


wsmoses:
Moved the interfaces and registration into EnzymeAutoDiffOpInterfaceImpl.cpp (the op's own dialect file) per review; the affine file is untouched now.

PR #3135, opened 2026-08-10
Fortran function like [WIP]

isaacaka:
# Add Fortran `enzyme_function_like` support

## Summary

This adds two ways to register a Fortran function as behaving like a
mathematical function known to Enzyme.

```fortran
use enzyme, only: enzyme_function_like, enzyme_log1p

call enzyme_function_like(log1p_like_function, enzyme_log1p)
```

The declaration-style alternative is closer to C's static registration:

```fortran
procedure(log1p_like_function), pointer, private :: &
  fn__enzyme_function_like__log1p => log1p_like_function
```

This declaration creates a module global that points at
`log1p_like_function`. Because Fortran cannot construct the equivalent of C’s 
statically initialized {function pointer, string pointer} struct the function whose rule is being copied is instead encoded in the global’s name. PreserveNVVM searches for `__enzyme_function_like__` and reads `log1p` from its
suffix.

## Trade-offs

- Call-style registration, such as
  `call enzyme_function_like(target, enzyme_log1p)`, simpler and clearer. 
  However, every math function name needs a global binding such as
  `integer(c_int), bind(C, name="enzyme_math_log1p") :: enzyme_log1p`.
  Users can declare missing bindings themselves.

- Procedure-pointer registration doesn't need a binding per function. Its
  less simple and the marker name must be unique, and the rule name
  must fit a Fortran identifier. Names containing punctuation etc won't work
  
### C-compatible static record

The closest translation of C's `{ function pointer, string pointer }` global
didn't work:

```fortran
type, bind(C) :: function_like_record
  type(c_funptr) :: target
  type(c_ptr) :: name
end type function_like_record

type(function_like_record), bind(C, name="__enzyme_function_like") :: &
  registration = function_like_record( &
    c_funloc(log1p_like_function), c_loc(log1p_name))
```

Flang rejects the initialiser because `c_funloc` and `c_loc` are not valid
constant initialisation expressions here:

```text
error: Must be a constant value
c_funloc(log1p_like_function), c_loc(log1p_name)
```

C is able to handle things like the above but Fortran cannot

### Passing `"log1p"` directly

```fortran
call enzyme_function_like(log1p_like_function, "log1p")
```

does not lower to a direct constant-string argument. Flang copies the literal through a stack temporary variable

## Tests

- End-to-end Fortran differentiation through the call-style registration.
- End-to-end differentiation through an independently registered
  procedure-pointer marker.
- IR checks that the declaration form adds `"enzyme_math"="log1p"`
- Add a warning for Conflicting registrations  while retaining current behaviour


enzyme/Fortran/README.md
```diff
@@ -124,3 +124,88 @@ for an example.
...
+
+The `enzyme_function_like` hook tells Enzyme to differentiate a function as if
+it were a known mathematical function. For example, Enzyme can use the
+derivative of `log1p` for `log1p_like_function`, regardless of its
+implementation.
+
+### Call-style registration
+
+The call-style interface follows the same pattern as `enzyme_autodiff`:
+
```
joewallwork, line 139: Does 'call-style' refer to the fact it's a subroutine?
```suggestion
### Function registration

The interface follows the same pattern as `enzyme_autodiff`:

```

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,88 @@ for an example.
...
+```fortran
+use enzyme, only: enzyme_function_like, enzyme_log1p
+
+call enzyme_function_like(log1p_like_function, enzyme_log1p)
+```
+
+`enzyme_log1p` supplies the symbolic function name `log1p`; its value is not
```
joewallwork, line 146:
```suggestion
Here `enzyme_log1p` supplies the symbolic function name `log1p`; its value is not
```

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,88 @@ for an example.
...
+`enzyme_function_like` hook before differentiation:
+
+```console
+$ opt -load-pass-plugin=/path/to/LLVMEnzyme-21.so \
+    -passes='preserve-nvvm,enzyme,preserve-nvvm-end' input.bc -o output.bc
+```
+
+When using this separate `opt` workflow, compile the Fortran source to LLVM
+ with `-O0`. Otherwise, Flang may inline calls to the function before
+`preserve-nvvm` processes the `enzyme_function_like` hook. The `FlangEnzyme`
```
joewallwork, line 163:
```suggestion
> [!WARNING]
> When using this separate `opt` workflow, compile the Fortran source to LLVM
> with `-O0`. Otherwise, Flang may inline calls to the function before
> `preserve-nvvm` processes the `enzyme_function_like` hook.

The `FlangEnzyme`
```

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,88 @@ for an example.
...
+
+```fortran
+module enzyme_math_names
+  use iso_c_binding, only: c_int
+  implicit none
+
+  integer(c_int), bind(C, name="enzyme_math_sin") :: enzyme_sin
+end module enzyme_math_names
+```
+
```
joewallwork, line 179: How about we just implement them all in the bindings ourselves so that users don't have to?

enzyme/test/Fortran/ReverseMode/function_like.f90
joewallwork, line 1: To avoid the risk of module collision (which has caused issues previously), we have switched to just putting functions or subroutines after a `contains` statement in the main program, rather than using modules. You can get away with this so long as you don't need to define any interfaces.

enzyme/test/Fortran/ReverseMode/function_like.f90
```diff
@@ -0,0 +1,44 @@
...
+  implicit none
+
+  double precision :: x, dx
+
+  x = 2.0d0
+  dx = 0.0d0
+  call enzyme_function_like(log1p_like_function, enzyme_log1p)
+  call enzyme_autodiff(test, x, dx)
+
+  if (abs(dx - 1.0d0 / 3.0d0) > 1.0d-10) error stop
```
joewallwork, line 40: Instead, encode this as a lit `CHECK` statement at the bottom of the file. Print out `dx` to a given number of decimal places and then check it matches 1/3 to the same precision. See other tests for examples of this.

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
joewallwork, line 1: Similar suggestions as for the other test.

joewallwork (changes requested):
This looks great so far. I left a few comments and suggestions on the Fortran side of things.

I was wondering: is there any particular reason you use `double precision` rather than `real` in the test and docs here?

enzyme/Fortran/README.md
isaacaka: I agree

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
```diff
@@ -0,0 +1,48 @@
...
+! RUN: %fc -flto -O0 -c %loadFortran %s -o /dev/stdout | %opt %loadEnzyme -passes="preserve-nvvm" -S | FileCheck %s --check-prefix=IR --implicit-check-not=__enzyme_function_like__
+! RUN: %fc -flto -O0 -c %loadFortran %s -o /dev/stdout | %opt %loadEnzyme -passes="preserve-nvvm,enzyme,preserve-nvvm-end" -o %t.ll && %fc -flto -O0 %t.ll -o %t1 && %t1 | FileCheck %s
+! RUN: %if flangenzyme %{ %fc -O0 %loadFortran %loadFlangEnzyme %s -o %t2 && %t2 | FileCheck %s %}
+! RUN: %if flangenzyme %{ %fc -O2 %loadFortran %loadFlangEnzyme %s -o %t2 && %t2 | FileCheck %s %}
+
+module function_like_procedure_pointer_test
+  implicit none
+
+  procedure(log1p_like_function), pointer, private :: &
+    fn__enzyme_function_like__log1p => log1p_like_function
```
vchuravy: My only question is this idiomatic? Could we use pragmas?

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
isaacaka: Done

enzyme/test/Fortran/ReverseMode/function_like.f90
isaacaka: Done

enzyme/test/Fortran/ReverseMode/function_like.f90
isaacaka: Done

enzyme/Fortran/README.md
isaacaka: I mean to distinguish between registering a function via a 'procedure pointer vs a 'call'

isaacaka:
> This looks great so far. I left a few comments and suggestions on the Fortran side of things.
> 
> I was wondering: is there any particular reason you use `double precision` rather than `real` in the test and docs here?

Both work, I think it was left over from when I was trying out different things. I've changed it to use real.

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
isaacaka: Yes, this is modern idiomatic Fortran. I've added additional examples in the README
Pragmas could work as well but it would require some changes to Flang

enzyme/Fortran/README.md
```diff
@@ -103,7 +103,7 @@ example, if you have a subroutine
 then you can make use of activity descriptors like so:
 ```fortran
   call enzyme_autodiff(my_subroutine, enzyme_const, n, &
-                       enzyme_dup, x, dx, enzyme_dup, y, dy
+                       enzyme_dup, x, dx, enzyme_dup, y, dy)
```
joewallwork, line 106: Good catch, thanks.

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,299 @@ for an example.
...
+| Function location and interface | When to use each form |
+|---|---|
+| Module function | Use a pointer declaration before the module's `contains` to keep registration with the function. The compiler supplies its explicit interface. A registration call in executable code also works. |
+| External function with an explicit interface | Put the pointer declaration after the interface block in the declaration section. A registration call in executable code also works. |
+| Internal function | Use call registration only if Flang supplies a direct function reference. Access to variables from the containing program or procedure can prevent registration. See the restriction below. The current pointer mechanism cannot register an internal function. |
+| External function with an implicit interface | Use `procedure(real)` for an ordinary pointer to a function with a real result. For scalar math registration, provide the required value arguments and an explicit interface. |
+
+### Call-style registration
+
+The call-style interface follows the same pattern as `enzyme_autodiff`:
```
joewallwork, line 150: Not really? except that it's called like a subroutine.

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,299 @@ for an example.
...
+
+You can declare other symbolic names in user code. Use the `enzyme_math_`
+prefix followed by a function name that Enzyme supports:
+
+```fortran
+module enzyme_math_names
+  use iso_c_binding, only: c_int
+  implicit none
+
+  integer(c_int), bind(C, name="enzyme_math_fmin") :: enzyme_fmin
```
joewallwork, line 234: Is this something that Enzyme supports?

enzyme/Fortran/README.md
joewallwork, line 1: It's good to have all this detail. Once we set up proper documentation for the Fortran bindings, the function-like stuff should probably be on its own page, though.

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,299 @@ for an example.
...
+
+An external function is defined outside any program, module, or other procedure.
+It can be in the same source file as its caller.
+
+Use an explicit interface when the function has a `value` argument, as required
+by the scalar math rules shown here. Put the pointer declaration after the
+interface block, before executable statements. Omit `private` outside a module.
+
+In this example, `double_value` follows `end program main`. Its interface block
+describes its value argument and result.
```
joewallwork, line 315: I'm not sure we should include this approach. It's generally considered bad practice to have procedures that don't belong to a module or program. I think if you applied the Fortitude linter to this source it would complain.

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,299 @@ for an example.
...
+
+Keep the explicit interface consistent with the external function definition.
+The internal `test` wrapper is allowed here because the initialized pointer
+targets the external `double_value`, not `test`.
+
+#### External function with an implicit interface
+
+A simple external function with scalar arguments passed by reference can use
+an implicit interface. Use this form for ordinary Fortran pointer calls when
+the function has no features that require an explicit interface.
```
joewallwork, line 367: Similarly here.

enzyme/Fortran/README.md
```diff
@@ -124,3 +124,299 @@ for an example.
...
+the explicit interface and registration declaration from the preceding example.
+Changing the argument to `value` changes the calling convention. Update the
+interfaces at all call sites and recompile the callers.
+
+#### Register a module function from a subroutine
+
+The declaration can instead appear in a subroutine's declaration section,
+before executable statements. For example, remove the module-level registration
+from `function_like_example` above and add this subroutine alongside its
+`double_value` and `test` functions:
```
joewallwork, line 403: I feel this could be combined with other variants. At the end of the day, you just need to include the declaration in the declaration block for the subroutine/module/program/function, no?

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
```diff
@@ -0,0 +1,52 @@
...
+
+end program main
+
+function double_value(x) result(y)
+  implicit none
+  real, value :: x
+  real :: y
+
+  y = 2.0 * x
+end function double_value
```
joewallwork, line 49: Again, I don't think we should include code that has procedures that don't belong to programs or modules.

joewallwork (changes requested):
Thanks for your work on this @isaacaka. I have a concern about including procedures that don't belong to programs or modules because it's considered bad practice.

enzyme/Fortran/README.md
isaacaka: Yes

enzyme/Fortran/README.md
isaacaka: Thanks, removed

enzyme/Fortran/README.md
isaacaka: Removed

enzyme/Fortran/README.md
isaacaka: Fixed the wording

enzyme/Fortran/README.md
isaacaka: Fixed the wording

enzyme/Fortran/README.md
isaacaka: Done

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
isaacaka: I can move it back into a module since procedure pointers can only go in modules or outside programs. I think for this test case we can avoid the module collision case that you mentioned by using a unique module name.

joewallwork (approved):
I'm happy with this now, thanks @isaacaka!

enzyme/Fortran/enzyme.f90
```diff
@@ -36,7 +38,76 @@ module enzyme
   integer(c_int), public, bind(C, name="enzyme_width")     :: enzyme_width
   integer(c_int), public, bind(C, name="enzyme_vector")    :: enzyme_vector
 
+  ! Symbolic function names for enzyme_function_like
+  integer(c_int), public, bind(C, name="enzyme_math_sin")   :: enzyme_sin
```
vchuravy: Can we automatically attach to `sin` to Fortran intrinsic.

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
```diff
@@ -0,0 +1,50 @@
...
+
+end module enzyme_test_reverse_function_like_procedure_pointer
+
+program main
+  use enzyme, only: enzyme_autodiff
+  use enzyme_test_reverse_function_like_procedure_pointer, only: double_value
+  implicit none
+
+  procedure(double_value), pointer :: &
+    fn__enzyme_function_like__log1p => double_value
```
vchuravy, line 29: Can we have this definition in the `module enzyme_test_reverse_function_like_procedure_pointer` instead?

enzyme/test/Fortran/ReverseMode/function_like.f90
```diff
@@ -0,0 +1,38 @@
...
+
+program main
+  use enzyme, only: enzyme_autodiff, enzyme_function_like, enzyme_log1p
+  implicit none
+
+  real :: x, dx
+
+  x = 2.0
+  dx = 0.0
+  call enzyme_function_like(double_value, enzyme_log1p)
```
vchuravy, line 23: Do we have to place this call in `main`.

enzyme/Fortran/enzyme.f90
```diff
@@ -36,7 +38,76 @@ module enzyme
...
+  integer(c_int), public, bind(C, name="enzyme_math_j0")    :: enzyme_j0
+  integer(c_int), public, bind(C, name="enzyme_math_j0f")   :: enzyme_j0f
+  integer(c_int), public, bind(C, name="enzyme_math_j1")    :: enzyme_j1
+  integer(c_int), public, bind(C, name="enzyme_math_j1f")   :: enzyme_j1f
+  integer(c_int), public, bind(C, name="enzyme_math_y0")    :: enzyme_y0
+  integer(c_int), public, bind(C, name="enzyme_math_y0f")   :: enzyme_y0f
+  integer(c_int), public, bind(C, name="enzyme_math_y1")    :: enzyme_y1
+  integer(c_int), public, bind(C, name="enzyme_math_y1f")   :: enzyme_y1f
+  integer(c_int), public, bind(C, name="enzyme_math_ldexp") :: enzyme_ldexp
+  integer(c_int), public, bind(C, name="enzyme_math_ldexpf") :: enzyme_ldexpf
```
vchuravy: I don't think they hurt, but I am also unsure how what user would need them. Most of these should be directly handled internally for intrinsics.

vchuravy (approved):

enzyme/test/Fortran/ReverseMode/function_like.f90
isaacaka: Yes, for subroutine calls

enzyme/test/Fortran/ReverseMode/function_like_procedure_pointer.f90
isaacaka: Yes, added a test for it

enzyme/Fortran/enzyme.f90
isaacaka: I've removed them

enzyme/Fortran/enzyme.f90
isaacaka: It would require additional changes to handle it. I have removed the bindings now though.

PR #3136, opened 2026-08-10
MLIR: shadow stores of pointers in reverse mode

wsmoses:
Storing an active pointer into memory has no float adjoint to
accumulate; its derivative story is structural, like llvm.getelementptr:
the shadow memory must hold the shadow pointer at the same spot, so
shadow loads traverse shadow structures. Previously nothing wrote the
shadow slot, and reverse-differentiating code that round-trips a data
pointer through a struct -- the capture struct a CUDA kernel launch
packs its arguments into -- read back a null shadow data pointer. A
pointer nothing differentiates is its own shadow, keeping inactive
fields readable through the shadow object.

Split out of #3132 per review.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
@@ -1042,12 +1042,24 @@ struct AffineStoreOpInterfaceReverse
     return SmallVector<Value>();
   }
 
+  // Same structural story as memref.store: a stored mutable value's shadow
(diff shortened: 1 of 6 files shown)


wsmoses:
All three were missing: llvm.store handled only non-mutable values in reverse, and memref.store/affine.store had the same gap with createShadowValues an explicit TODO stub. Extended the PR to cover all three consistently, each with a lit test.

wsmoses:
Added one more case of the same convention: the memory-identity *forward* handler refused a stored constant of mutable type (MFEM's compatibility test stores an inactive data pointer into differentiated memory). An inactive pointer is its own shadow there too; test included.

PR #3137, opened 2026-08-10
Keep the shadow of a constant a constant in invertPointerM

### LINKED ISSUE #3128: assertion cast<Ty>() argument of incompatible type!
https://tinyurl.com/27srmb39
```llvm
; Enzyme: cast<Constant>() argument of incompatible type
;
;   opt castconst-reduced.ll -load-pass-plugin=libEnzyme-23.so -passes=enzyme -S
;
; opt: llvm/Support/Casting.h:572: decltype(auto) llvm::cast(From *)
;   [To = llvm::Constant, From = llvm::Value]:
;   Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"'
;
; GradientUtils.cpp, the ConstantDataArray case of invertPointerM:
;
;   Value *val = invertPointerM(el, BuilderM, subTT);
;   Vals.push_back(cast<Constant>(val));      ; <-- fires here
;
; @tab holds two packed doubles as bytes, so its ConstantDataArray elements are
; i8. Element 0 is byte 0x55 ('U') of 1/3 = 0x3FD5555555555555. Its one-byte
; TypeTree window overlaps a float without covering the whole double, so
; inverting that element falls into the generic partially-float path, which
; emits an alloca/store/load and hands back a LoadInst rather than a Constant.
;
; The store makes @tab active, so Enzyme builds a shadow global and inverts its
; initializer; without it the global stays constant and is never inverted.
;
; Reduced with llvm-reduce from repro/castconst-single.rs (a no_std Rust file).

@tab = global [16 x i8] c"UUUUUU\D5?\00\00\00\00\00\00\D0?"
@enzyme_dup = external global ptr

define void @f(ptr %0, i64 %1, ptr %2, i64 %3, i32 %4) {
  store i32 0, ptr @tab, align 4
  ret void
}

define {} @entry() {
  %1 = tail call {} (...) @__enzyme_autodiff(ptr @f, ptr @enzyme_dup, ptr null, ptr null, ptr @enzyme_const, i64 0, ptr @enzyme_dup, ptr null, ptr null, ptr @enzyme_const, i64 0, ptr @enzyme_const, i32 0)
  ret {} %1
}

declare {} @__enzyme_autodiff(...)
```
@wsmoses

ZuseZ4:
Rust always emits arrays as `[N x i8]`, which might be why it showed up repeatedly.

The ConstantDataArray / ConstantArray / ConstantStruct / ConstantVector cases of invertPointerM invert each element and feed the results straight into ConstantArray::get and friends, so they do cast<Constant>() on whatever the recursive call returned. That call is not guaranteed to hand back a constant: when the element's TypeTree says it overlaps a float but does not cover all of it, invertPointerM falls through to the generic partially-float path, which emits an alloca/store/store-zero/load sequence and returns the LoadInst. The cast then trips

  Assertion `isa<To>(Val) && "cast<Ty>() argument of incompatible type!"'

in llvm::cast<llvm::Constant>.

A partial float window is exactly what the individual i8s of a [N x i8] blob holding packed doubles produce. Differentiating a Rust `static` of type [16 x i8] holding f64 tables reaches invertPointerM as the initializer of a global, recurses into the ConstantDataArray case, and inverting the element `i8 85` (byte 0 of 0x3FD5555555555555) returns a LoadInst.

Fix it at the source rather than at the four call sites: when the value being inverted is a scalar constant, compute the same byte-wise nulling by folding instead of by emitting instructions. nullFloatBytesOfConstant() zeroes every byte the TypeTree types as a float (or as anything other than an integer or a pointer) and keeps the rest, which is what the alloca sequence does, and returns null for anything it cannot reinterpret bitwise so those inputs still take the old path.

None of the 1204 existing lit tests reach the new branch, so this only affects inputs that used to assert.

@wsmoses 
Closes #3128

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -5441,6 +5441,65 @@ static bool allNullOrUndef(Value *C, const DataLayout &dl, TypeTree TT) {
   return false;
 }
 
+/// Return a copy of the scalar constant `C` with every byte that `TT` types as
+/// a float (or as something other than an integer or a pointer) replaced by
+/// zero, keeping the remaining bytes as they were. This is the constant-folded
+/// equivalent of the alloca/store/store-zero/load sequence invertPointerM emits
+/// for a value that is only partially float, and it lets the shadow of a
+/// constant stay a constant. Returns null if `C` cannot be reinterpreted
+/// bitwise, in which case the caller falls back to emitting that sequence.
+static Constant *nullFloatBytesOfConstant(Constant *C, const TypeTree &TT,
+                                          const DataLayout &DL) {
+  auto *Ty = C->getType();
+
+  APInt bytes(8, 0);
+  if (auto CI = dyn_cast<ConstantInt>(C))
+    bytes = CI->getValue();
+  else if (auto CFP = dyn_cast<ConstantFP>(C))
+    bytes = CFP->getValueAPF().bitcastToAPInt();
+  else
+    return nullptr;
+
+  // Only a whole number of bytes can be masked byte-wise, and only a type
+  // whose value fills its in-memory footprint can be rebuilt from those bytes
+  // (this rules out i1, i24, x86_fp80, ...).
+  unsigned bits = bytes.getBitWidth();
+  size_t size = bits / 8;
+  size_t storeSize = (DL.getTypeStoreSizeInBits(Ty) + 7) / 8;
+  if (size == 0 || size * 8 != bits || size != storeSize)
+    return nullptr;
+
+  APInt mask(bits, 0);
+  for (size_t i = 0; i < size;) {
+    auto CT = TT[{(int)i}];
+    size_t chunk = 1;
+    if (CT == BaseType::Pointer) {
+      // A pointer is carried through unchanged, like an integer.
+      i += DL.getPointerSize(0);
+      continue;
+    } else if (auto flt = CT.isFloat()) {
+      chunk = (DL.getTypeSizeInBits(flt) + 7) / 8;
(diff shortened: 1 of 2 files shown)


wsmoses (commented):
this doesn't feel like the right fix, why is it not a constant in some cases?

ZuseZ4:
@wsmoses pushed an update. I'd guess the first commit should be dropped, just left it while you're reviewing. But also feel free to close here and fix it yourself.

Shortened AI  answer: 


  The value is fine — the TypeTree is empty. invertPointerM's GlobalVariable case inverts an initializer using the tree it holds for the pointer, which says nothing about the
  pointee, and the aggregate cases slice that empty tree per element. Instrumenting the reproducer: aggregate arrives as {}, every element sliced to {}.

  An all-Unknown tree is the one input the predicates disagree on — anyFloat() says "may be float", allFloat() says "not all float" — which is exactly the guard on the
  partially-float path. So an untyped constant gets routed into emitting alloca/store/load.

  Type analysis already knows: TR.query() returns {[-1]:Anything}, which slices to Anything per element, and the existing isa<ConstantData> && allFloat(anythingIsFloat) case
  folds each to a null constant. So:

  if (!TT.isKnown() && isa<Constant>(oval))
    TT = TR.query(oval);

  This alone fixes #3128 — with commit 1's fold disabled both reproducers pass and @tab_shadow is zeroinitializer either way. I kept the fold as a backstop for constants that
  really are typed partially float; happy to drop it for an EmitFailure at the four cast<Constant> sites if you prefer the smaller diff.

  Across the 1205 lit tests the new branch is reached by 1 file (the new test), and nothing else changes.

PR #3139, opened 2026-08-10
MLIR: an inactive pointer stored into shadow memory is its own shadow

wsmoses:
The memory identity forward handler refused a stored constant of mutable
type outright. The same structural story as the reverse-mode shadow
stores applies: a mutable value nothing differentiates is its own
shadow, so the shadow slot takes the primal pointer and structural
fields read the same through the shadow object. MFEM's enzyme
compatibility test stores an inactive data pointer into differentiated
memory and hit the refusal.

If the given activity was wrong about the pointer, only runtime
activity could catch it; a warning says so until that exists.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
@@ -16,6 +16,7 @@
 #include "Implementations/LoopCheckpointing.h"
 #include "Interfaces/AutoDiffOpInterface.h"
 #include "Interfaces/GradientUtilsReverse.h"
+#include "Interfaces/Utils.h"
 #include "Passes/RemovalUtils.h"
 #include "Passes/Utils.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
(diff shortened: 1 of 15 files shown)


wsmoses:
Round of review changes:
- `inactiveStoredValueShadow` now takes the builder, returns zero (`createNullValue` on the width-batched shadow type) for immutable types, and concatenates the primal across the batch width for mutable ones (`getConcatValue`).
- The store `createShadowValues` guards stop conflating their cases: a type without autodiff semantics is a hard failure (the interface method now returns `LogicalResult`, threaded through visitChild, the tblgen emitter, and all models) instead of a silent skip; an undifferentiated destination or an immutable value (whose shadow the reverse-sweep adjoint owns) return success without work.
- The memory-identity handler's two constant branches collapsed into one helper call.

Note: the `createShadowValues` signature change needs a paired Enzyme-JAX update to its external models (Pointer2Memref/Memref2Pointer/GPUWrapper etc.) when bumping — I have that ready to PR.

wsmoses:
Load-side guards restructured per review:
- `!iface` (a type without autodiff semantics) is now a hard failure on both llvm.load and affine.load — silently skipping was the null-shadow bug pattern.
- Constant `loadOp` stays success (nothing active to stand for); a constant address/memref under an *active mutable* result is now a failure — there is no shadow to hand out, so the claimed activity cannot be honored.
- One deliberate divergence from the quoted suggestion: `!iface.isMutable()` keeps returning success in `createShadowValues` — an immutable (float) load's derivative is the reverse sweep's adjoint accumulation, and this hook has nothing to place for it; failing there would refuse every float load. The float-load reverse tests confirm.

Same structure as the store side (fail unknown type / success undifferentiated destination / immutable belongs to the adjoint).

wsmoses:
Fixed: the load-side restructure had been lost in a branch shuffle before the last push — it's actually in now, for all three load models (llvm.load, memref.load, affine.load):
- `!iface` → hard failure;
- immutable → success (the adjoint's to accumulate);
- constant `loadOp` → success;
- non-constant load out of a constant address/memref → error ("cannot load a non-constant value out of a constant ...").

Also ran clang-format over every touched file — the CI format failure came from the interface-migration files I had missed.

PR #3147, opened 2026-08-11
mlir: Allocate the $b$ checkpoints at once in binomial checkpointing

Pangoraw:
the honest summary: the change removes ~30 fixed allocations from b=16, worth ~15 ms — a real 4.9% at nt=50 that fades to under 1% by nt=800, and nothing measurable for b=4. It won't move the long-running checkpointed cases that motivate binomial in the first place.


wsmoses:
offhand looks reasonable, but mlir tests fail

wsmoses (approved):

PR #3152, opened 2026-08-14
Don't push a poison aggregate onto the sret rooting worklist

kshyatt:
Another 🤖 assisted PR, I ran into this while trying to fix things for a very long tape fed into Enzyme.jl:

`needsReRooting` walks every value stored into an sret argument to check each tracked pointer is also parked in the returnRoots array. Walking back through an insertvalue chain, the branch for an insert of an untracked field pushed the aggregate operand unconditionally:

    CountTrackedPointers tracked(IVI->getInsertedValueOperand()->getType());
    if (tracked.count == 0) { storedValues.push_back(IVI->getAggregateOperand()); continue; }

When that operand is the poison the chain was built from, the bare poison lands on the worklist. The struct handling below then looks for an extractvalue covering poison's tracked fields, finds none, and reports

     failed to find extracted pointer for %tape poison at index 1

before falling through to the pointer-shaped handling, whose assert(0) fires on any non-pointer value. Julia hits this whenever a nested sub-tape survives SROA as a whole-aggregate store and its first field is untracked -- reverse mode over PEPSKit's CTMRG aborts the process this way, since the shipped Enzyme_jll RelWithDebInfo build leaves assertions live.

A poison, undef, or zeroinitializer aggregate has no live pointer in any field, so it needs no root regardless of which field the insertvalue overwrites. Hoist the trivial-base test above both branches and drop such a base, which is the reasoning the tracked branch three lines down already relies on. In the Julia case the tracked pointers are separately stored into the returnRoots, so the walk then completes and the pass correctly concludes it has no work to do.

Also un-shadow `legal` in the struct handling. The inner declaration meant the coverage failure set a variable nobody read and fell into that same assert; report it through CustomErrorHandler like the pointer path does and reroot conservatively. This is not a behaviour change for NDEBUG builds, which already fell through to exactly that handler.


oscardssmith (approved):
not read, but to run ci

kshyatt:
I got added to the repo so you can unapprove now :D

enzyme/Enzyme/FixupJuliaCallingConvention.cpp
```diff
@@ -336,13 +343,27 @@ bool needsReRooting(llvm::Argument *arg, bool &anyJLStore,
...
+          // rooted separately. Report it and reroot conservatively rather than
+          // falling into the pointer-shaped handling below, which asserts on
+          // any non-pointer value.
+          if (hasReturnRootingAfterArg) {
+            std::string s;
+            llvm::raw_string_ostream ss(s);
+            ss << "Could not find use of stored value\n";
+            ss << " sv: " << *sv << "\n";
+            CustomErrorHandler(ss.str().c_str(), wrap(sv), ErrorType::GCRewrite,
+                               nullptr, wrap(arg), nullptr);
```
vchuravy, line 363: needs `if (CustomErrorHandler)`

vchuravy (approved):

wsmoses (changes requested):
blocking merge for me to also review

enzyme/Enzyme/FixupJuliaCallingConvention.cpp
```diff
@@ -336,13 +343,30 @@ bool needsReRooting(llvm::Argument *arg, bool &anyJLStore,
...
+            ss << "Could not find use of stored value\n";
+            ss << " sv: " << *sv << "\n";
+            if (CustomErrorHandler) {
+              CustomErrorHandler(ss.str().c_str(), wrap(sv),
+                                 ErrorType::GCRewrite, nullptr, wrap(arg),
+                                 nullptr);
+            }
+          }
+          legal = false;
+          break;
```
vchuravy, line 369: Once https://github.com/EnzymeAD/Enzyme/pull/3244 is merged this needs to be updated

PR #3153, opened 2026-08-14
Update LLVM 19 guards for trigonometric functions

nkoukpaizan:
This marks several trigonometric functions as `Intrinsic` for `LLVM >= 19`

Closes #3101 

I double-checked the list of functions to tag on the [LLVM repo](https://github.com/llvm/llvm-project) as follows:
```console 
git grep -h 'DefaultAttrsIntrinsic' llvmorg-18.1.0 -- llvm/include/llvm/IR/Intrinsics.td | grep llvm_anyfloat_ty > 18.log
git grep -h 'DefaultAttrsIntrinsic' llvmorg-19.1.0 -- llvm/include/llvm/IR/Intrinsics.td | grep llvm_anyfloat_ty > 19.log
diff -b 18.log 19.log

4a5,7
>   def int_asin : DefaultAttrsIntrinsic<[llvm_anyfloat_ty], [LLVMMatchType<0>]>;
>   def int_acos : DefaultAttrsIntrinsic<[llvm_anyfloat_ty], [LLVMMatchType<0>]>;
>   def int_atan : DefaultAttrsIntrinsic<[llvm_anyfloat_ty], [LLVMMatchType<0>]>;
6a10,13
>   def int_tan  : DefaultAttrsIntrinsic<[llvm_anyfloat_ty], [LLVMMatchType<0>]>;
>   def int_sinh : DefaultAttrsIntrinsic<[llvm_anyfloat_ty], [LLVMMatchType<0>]>;
>   def int_cosh : DefaultAttrsIntrinsic<[llvm_anyfloat_ty], [LLVMMatchType<0>]>;
>   def int_tanh : DefaultAttrsIntrinsic<[llvm_anyfloat_ty], [LLVMMatchType<0>]>;
49a57,59
>   def int_experimental_constrained_asin  : DefaultAttrsIntrinsic<[ llvm_anyfloat_ty ],
>   def int_experimental_constrained_acos  : DefaultAttrsIntrinsic<[ llvm_anyfloat_ty ],
>   def int_experimental_constrained_atan  : DefaultAttrsIntrinsic<[ llvm_anyfloat_ty ],
51a62,65
>   def int_experimental_constrained_tan  : DefaultAttrsIntrinsic<[ llvm_anyfloat_ty ],
>   def int_experimental_constrained_sinh  : DefaultAttrsIntrinsic<[ llvm_anyfloat_ty ],
[...]

PR #3145, opened 2026-08-11
Add GitHub Actions linting workflow

joewallwork:
Okay one last linting workflow! This PR uses the security-oriented [Zizmor](https://docs.zizmor.sh/) linter to identify potential issues with GitHub Actions workflows.

I heeded its warnings and fixed a syntax error that was stopping the `enzyme-bazel` workflow from running previously.

PR #3154, opened 2026-08-14
Fix Bazel workflow file at OS definition

nkoukpaizan:
I noticed in #3153 that the Bazel workflow did not run. See https://github.com/EnzymeAD/Enzyme/actions/runs/31821292498/workflow. This fixes the definition of the OS in the workflow file. 

Draft to see if this breaks anything else.


nkoukpaizan:
Closing for now, as I am not able to ascertain that the issue is fixed. Should be covered by #3145.

PR #3156, opened 2026-08-19
Preserve types on reverse cache reloads

vimarsh6739:
Reverse cache reloads created by `lookupM` dropped the original TypeTree. Preserve it as `enzyme_type` metadata so nested differentiation can classify cached integer loop indices.

Adds a focused check to `bsearch2.ll`. Tracks EnzymeAD/Enzyme.jl#3472.


wsmoses (approved):

vchuravy:
Ah claude came up with a similar fix in https://github.com/EnzymeAD/Enzyme/pull/3202

PR #3158, opened 2026-08-20
Partial update of unit tests to use newer regex

markabate:
Many of the Enzyme unit tests seem to fail with the latest version of LLVM on Mac (target triple: arm64-apple-darwin25.5.0). This updates some of the regex for a few of the ForwardMode tests. Most changes are allowing for opaque pointers.

For some of the tests, I added LL16 and LL24 FileCheck prefixes to handle the different IR you get between LLVM pre-version 16 and mainline (v 24). I'm open to suggestions if there are different ways we want this to be handled or if we don't want to support LLVM mainline in the unit tests yet

### DIFF
--- a/enzyme/test/Enzyme/ForwardMode/atomicfadd.ll
+++ b/enzyme/test/Enzyme/ForwardMode/atomicfadd.ll
@@ -22,9 +22,9 @@ attributes #0 = { norecurse nounwind readonly uwtable }
 attributes #1 = { nounwind uwtable }
 attributes #2 = { nounwind }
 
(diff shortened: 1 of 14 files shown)


markabate:
@wsmoses The Julia integration tests are failing on this, but the error looks unrelated?

markabate:
@wsmoses Do you think making changes like this to the unit tests is advisable, or should we only test against <= LLVM 16 for now? I was just motivated to write this since I was trying to develop with LLVM mainline. We could also discuss at the next Enzyme meeting.

PR #3160, opened 2026-08-21
Add Fortran OpenMP tests

### LINKED ISSUE #3065: Add tests for Fortran + OpenMP
We should add tests for the supported OpenMP functionality and see if Fortran+OpenMP can be handle already or if there is still a gap

joewallwork:
Closes #3065

This PR adds a variant of the `square` example for Fortran+OpenMP.

It loops over the same batch of values in the `BatchMode` test, dispatching to OpenMP, and computes the corresponding derivatives via reverse mode.

Currently, this seems to work fine at `-O0` and `-O1`, but not at `-O2` and `-O3`.

Note that I also enabled the Fortran `BatchMode` tests as I noticed they weren't hooked up in `enzyme/test/Fortran/CMakeLists.txt`.


joewallwork:
The failing Julia tests don't seem to be relevant to this PR.

vchuravy (approved):

PR #3167, opened 2026-08-25
Incorrect gradient: Zero the shadow in the reverse sweep for untyped zeroing memsets

ZuseZ4:
Claude forgot to use strict-aliasing=0 for the llvm-ir+opt combo, which accidentally gave the right result on gpt2.rs, whereas cargo+enzyme didn't (since it correctly sets strict-aliasing=0). The code's probably easier to follow than all the fluff explanation below, and it seems reasonable that zero memset is independent of the type? Left the prose in case it has anything interesting. @wsmoses 

visitMemSetCommon splits on whether TypeAnalysis could determine the element type of the region. With a known float type the shadow memset is emitted into the reverse sweep; with an unknown type it is emitted only into the forward sweep, which is right for zeroing a freshly allocated buffer but wrong for a memset that re-initialises a buffer inside a loop. The shadow accumulator is then never re-zeroed between reverse iterations and any gradient accumulated into it grows with the trip count.

This is reachable from any frontend whose allocas are untyped byte arrays. Rust is the motivating case: it must run with EnzymeStrictAliasing=false because its memory model is untyped, and a tiled reduction such as

    let mut acc = [0.0f32; 8];
    for i in 0..c { for k in 0..8 { acc[k] += inp[k*c+i] * w[i]; } }

lowers to `alloca [32 x i8]` plus `llvm.memset`. With strict aliasing off TypeAnalysis will not propagate the float type from the nested loop back into the alloca, since that block does not post-dominate the memset, so secretty is null and the shadow is zeroed only in the forward sweep. On GPT-2 this made the weight gradient of the final projection over-accumulate by a factor of the trip count -- roughly 1e5 -- while the loss stayed correct.

A zeroing memset kills every prior value in the region, so no adjoint can flow past it, and zeroing the shadow in the reverse sweep is correct regardless of what the element type turns out to be. Restricted to stack slots, whose shadow the caller cannot observe, so a shadow pointer held elsewhere is never clobbered. Enzyme's preprocessing rewrites some allocas into heap allocations tagged !enzyme_fromstack, so both spellings are accepted.

The added test is reduced from rustc output and fails without this change. Verified end to end: with the fix, GPT-2 124M differentiated whole-model under -enzyme-strict-aliasing=0 matches central finite differences and reproduces llm.c's PyTorch reference gradients for all 16 parameter tensors (tol 2e-2, max deviation 1.02e-3). No change in the Enzyme/ReverseMode suite pass/fail set.


enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -3347,6 +3347,38 @@ class AdjointGenerator : public llvm::InstVisitor<AdjointGenerator> {
...
+    // declared as
+    //     let mut acc = [0.0f32; 8];
+    // lowers to exactly this pattern.
+    //
+    // A zeroing memset kills every prior value in the region, so no adjoint can
+    // flow past it, and zeroing the shadow in the reverse sweep is correct
+    // whatever the element type turns out to be. Restricted to stack slots,
+    // whose shadow the caller cannot observe, so that a shadow pointer held
+    // elsewhere is never clobbered.
+    bool zeroShadowInReverse = false;
```
wsmoses, line 3369: this is a hack, merely checking if it is/was a stack pointer isn't the actual cause of a correctness issue

wsmoses (changes requested):

enzyme/Enzyme/AdjointGenerator.h
ZuseZ4: Great, sounds like you see the problem. opened https://github.com/EnzymeAD/Enzyme/issues/3174 so you can add a proper fix and closing here.

wsmoses: I do not see the problem, but I can say this is not a fix

ZuseZ4: https://tinyurl.com/29wz8udt (toggle the strict-aliasing flag to see the bug)

PR #3170, opened 2026-08-25
Fix the ancestor check in overwritesToMemoryReadByLoop

ZuseZ4:
This was encountered while trying to fix some >100GB caching of const arguments in gpt2, effectively automating the impact of manually applying enzyme_nocache. You might be able to make more out of it, or close it if hallucinated.  @wsmoses 

Both legality gates read

  for (const Loop *L = anc; anc != scope; anc = anc->getParentLoop()) {
    if (!visitedAncestors.count(L))
      legal = false;
  }

`L` is initialised to `anc` and never advanced, while the condition and increment operate on `anc`.  So instead of requiring every common loop in [anc, scope) to have been seen as an induction variable -- what the adjacent comment asks for, "lest we have a repetition of the store" -- the gate tests the innermost one repeatedly, and an unvisited outer common loop still leaves `legal` true.

`anc` is also captured by reference by the `skipLoop` lambda, and the first gate walks it up to `scope` (or null when `scope` is null).  The second `hasOverlap` call therefore runs with a mutated `anc`: `skipLoop` returns false unconditionally, `visitedAncestors` stays empty, and the second gate's loop never even enters, making the load-before-store direction vacuously legal under different assumptions than the store-before-load direction.

Walk `L` and leave `anc` alone.

Doing so exposes that `skipLoop` only inserts when `anc == L || anc->contains(L)` -- `L` at or nested inside `anc` -- while the gates walk outward, so a proper ancestor of `anc` could never be marked visited and the fixed gates would only pass when `anc->getParentLoop() == scope`.  Accept `L->contains(anc)` too, which restores that precision while still rejecting the unsound case.

The new test is a two-deep nest with an opaque per-outer-iteration offset, so the store address is an AddRec of the inner loop only: the inner loop is marked visited, the outer never can be.  With offs[i+1] = offs[i]+M+1 the store in iteration i+1 overwrites exactly what iteration i read, so before this change the load is judged recomputable and the reverse sweep reloads clobbered memory. Observed on a compiled run: dA[5] and dA[10] came back 6 and 6 where the exact values are 12 and 22, silently and without a crash.

Introduced by 313c9898 ("Loop-level rematerialization", #518).


wsmoses:
ci fails

enzyme/test/Enzyme/ReverseMode/cache-unvisited-ancestor-loop.ll
```diff
@@ -0,0 +1,105 @@
...
+  ret double %r
+}
+
+; CHECK: define internal void @diffef(ptr {{.*}}%A, ptr {{.*}}%"A'", ptr {{.*}}%offs, i64 %N, i64 %M, double %differeturn)
+
+; The load must be cached: performed in the forward sweep and spilled into a
+; cache allocation ...
+;
+; The allocation's SSA name is deliberately not checked.  Enzyme derives it from
+; the cached value on some LLVM versions (%v_malloccache) and falls back to a
```
wsmoses: please look at other tests and add check lines for all ir, like elsewhere

ZuseZ4: Also asked it once more to try & reduce it a little further, but that didn't went anywhere.

ZuseZ4:
CI failures are pre-existing.

enzyme/Enzyme/Utils.cpp
```diff
@@ -2763,7 +2763,12 @@ bool overwritesToMemoryReadByLoop(
     if (scope && L->contains(scope))
       return false;
 
-    if (anc && (anc == L || anc->contains(L))) {
+    // `L->contains(anc)` matters because the legality gates below walk
+    // *outward* from `anc` towards `scope`.  Without it a proper ancestor of
+    // `anc` can never be marked visited, so those gates could only ever pass
+    // when `anc->getParentLoop() == scope` -- sound, but needlessly pessimistic
+    // for a strided access whose outer loop is genuinely accounted for.
+    if (anc && (anc == L || anc->contains(L) || L->contains(anc))) {
```
wsmoses, line 2771: I don't trust this logic, can you @ZuseZ4 [not the LLM] explain to me why this is correct/walk me through things

ZuseZ4: https://tinyurl.com/2233rh7g
Change it between O0 and O1, that should show different numbers.

wsmoses: I'm not asking about the test case. I'm asking you to explain the logic of this line in the PR in the context of the rest of the code here

ZuseZ4: Enzyme does not have nearly enough docs for me to understand the code, so I've given up on that long ago. Since there doesn't seem to be capacity to fix bugs from maintainers I'm happy to provide LLM suggestions that solve the issue at hand, but Enzyme isn't part of my job and I don't have capacity to engage beyond llm bugfixes. Shall I convert it into an issue so you can find your own solution in the future?

wsmoses: okay without actually understanding what the logic is behind this line change, it'll be hard for me to review it properly.

what happens if it is restored back?

if you don't have cycles yourself to diagnose, maybe one of your collaborators from the rust side can offer some thoughts?

if not (though ideally after testing its reversion), leave it open and I'll try to understand it later

PR #3173, opened 2026-08-25
Forward terminator handler: shadow only results the terminator forwards

wsmoses:
Fixes the pre-existing `-c dbg` failure of Enzyme-JAX's `diffrules_enzymexla_gpu_wrapper_fwd.mlir`: `regionTerminatorForwardHandler`'s fallback path asserts `parentOp->getNumResults() == origTerminator->getNumOperands()`, but a parent op may legitimately carry more results than its terminator forwards — enzymexla's `gpu_wrapper` yields nothing for its token-like index result. Release builds happened to work (the loop only touched paired results); debug builds aborted. Pair up to the terminator's operand count; unpaired trailing results have nothing to shadow.

### DIFF
--- a/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
+++ b/enzyme/Enzyme/MLIR/Implementations/CoreDialectsAutoDiffImplementations.cpp
@@ -333,15 +333,23 @@ void mlir::enzyme::detail::regionTerminatorForwardHandler(
           successor.isOperation()
               ? parentOp->getResults()
               : regionBranchOp.getSuccessorInputs(successor);
-      assert(operandRange.size() == targetValues.size());
-      for (auto &&[i, target] : llvm::enumerate(targetValues)) {
+      // The parent may carry more results than the terminator forwards to it
+      // (a gpu wrapper's token-like result has no yield operand); only the
+      // paired prefix can be shadowed.
+      size_t numPaired = std::min(operandRange.size(), targetValues.size());
+      for (auto &&[i, target] :
+           llvm::enumerate(targetValues.take_front(numPaired))) {
         if (!gutils->isConstantValue(target))
           operandsToShadow.insert(operandRange.getBeginOperandIndex() + i);
       }
     }
   } else {
-    assert(parentOp->getNumResults() == origTerminator->getNumOperands());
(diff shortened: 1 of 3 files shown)


wsmoses:
Amended: the assert that actually fires (line 336 at the Enzyme-JAX pin) is on the **region-branch-interface path** — enzymexla's `gpu_wrapper` implements `RegionBranchOpInterface`, so the parent-successor's `targetValues` (the token-like index result) outnumber the terminator's successor operands there, not in the fallback. Both paths now pair up to the shorter count.

On a test: the shape needs a registered region-terminator (`scf.yield`/`scf.condition`/`affine.yield`/`linalg.yield` are the only ones wired to this handler) whose parent carries results the terminator does not forward — and every upstream parent of those pairs exactly by verifier, so the construct is not expressible in this repo's op vocabulary; it only exists downstream (`enzymexla.gpu_wrapper` + `polygeist_yield`). Regression coverage is Enzyme-JAX's existing `diffrules/enzymexla/gpu_wrapper_fwd.mlir`, which aborts on assert-enabled builds without this fix and is exercised by EJ's `-c dbg` CI lane on every run once the ENZYME_COMMIT bump lands (verified the test passes against this branch). If in-repo coverage is wanted regardless, the clean route would be a minimal test-dialect op with an unforwarded token result — happy to add that as a follow-up if preferred.

(While probing alternatives, found a separate latent issue worth its own ticket: forward AD of *buffer-semantics* `linalg.generic` produces invalid IR in release — the generic's rule doubles the outs but the yield never gets a tangent operand, because shadowing is keyed on result activity and the memref form has no results; additionally the yielded value's activity reads constant through the region args.)

wsmoses:
Added an in-repo regression test by wiring up forward-mode AD of `gpu.launch` (its operands and optional token result carry no tangents, so the tangent is the launch itself with a differentiated body) and registering `gpu.terminator` through the region-terminator handler. `gpu.terminator` has no operands at all, so an async launch is exactly the shape this PR fixes — a region op with more results than its terminator forwards. Without the fix, `test/MLIR/ForwardMode/gpulaunch.mlir` dies on the old `parentOp->getNumResults() == origTerminator->getNumOperands()` assert (verified against a Debug build at the base commit); with it, the test passes and the full `check-enzymemlir` suite is green (205 pass, 1 pre-existing XFAIL).

PR #3182, opened 2026-09-04
Add simple Fortran+MPI tests

### LINKED ISSUE #3066: Add Fortran MPI Integration Tests
related: https://github.com/EnzymeAD/Enzyme/pull/3064#issuecomment-5179233415

Fortran CI currently installs intel MPI dependency but no integration tests in .github/workflows/fortran.yml actually check Fortran MPI with Enzyme. Note, however, that MPI tests are covered at the rule/transformation level (see `enzyme/test/**/*mpi*`) and also by the C integration tests (`enzyme/test/Integration/*mpi*`).

Here are some tentative TODOs

- [x] MPI integration tests should be added (this should at least include Fortran MPI tests... in principle if such integration tests pass one can infer that Enzyme is stable also for more complex C/C++ programs containing MPI calls)
- [x] An MPI PR would also have to update the CI so that the Flang builds OpenMPI (see below for section that would need to add OpenMPI install) since at the moment the Flang build does not install OpenMPI

https://github.com/EnzymeAD/Enzyme/blob/1db6e649de614009527b24004ee85360d6a81de1/.github/workflows/fortran.yml#L52-L58

joewallwork:
Closes #3066

This PR adds simple tests for Fortran+MPI.

- [x] Rank
- [x] Size
- [x] Gather
- [x] Scatter
- [x] Allreduce
- [x] Reduce

To get it working, I needed to:
* Canonicalize MPI calls to support the different equivalent variants (Fortran ones have trailing underscores).
* (Attempt to) handle MPI Fortran ABI inconsistenties.
* Set the `HOME` environment variable in the lit config because this is needed by MPI.
* Install OpenMPI in the Fortran CI.

### DIFF
--- a/.github/workflows/fortran.yml
+++ b/.github/workflows/fortran.yml
@@ -56,9 +56,16 @@ jobs:
         sudo apt-add-repository "deb http://apt.llvm.org/`lsb_release -c | cut -f2`/ llvm-toolchain-`lsb_release -c | cut -f2`-${{ matrix.llvm }} main"
         sudo apt-get update && sudo apt-get install -y cmake gcc g++ llvm-${{ matrix.llvm }}-dev libomp-${{ matrix.llvm }}-dev ninja-build pip clang-${{ matrix.llvm }} flang-${{ matrix.llvm }}
         sudo python3 -m pip install --upgrade pip lit
+    - name: install openmpi
+      run: |
+        git clone https://github.com/spack/spack.git
+        . spack/share/spack/setup-env.sh
+        spack compiler find
+        spack install openmpi ^llvm
     - uses: actions/checkout@v4
     - name: generate build system
       run: |
+        spack load openmpi
         rm -rf build && mkdir build && cd build
         cmake ../enzyme -GNinja -DCMAKE_BUILD_TYPE=${{ matrix.build }} -DLLVM_DIR=/usr/lib/llvm-${{ matrix.llvm }}/lib/cmake/llvm -DLLVM_EXTERNAL_LIT=`which lit` -DENZYME_FORTRAN=ON -DENZYME_FLANG=ON -DCMAKE_Fortran_COMPILER=${{ matrix.compiler }}-${{ matrix.llvm }}
     - name: build enzyme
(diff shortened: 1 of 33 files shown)


enzyme/Enzyme/ActivityAnalysis.cpp
```diff
@@ -712,24 +727,45 @@ bool ActivityAnalyzer::isFunctionArgumentConstant(CallInst *CI, Value *val) {
       CI->getArgOperand(0) != val && CI->getArgOperand(1) != val)
     return true;
 
+  // Canonicalize MPI names across calling conventions (C "MPI_Recv",
+  // "PMPI_Recv" and Fortran "mpi_recv_" etc.): the leading argument indices
+  // match between the ABIs, the Fortran ABI only appends an `ierr` argument.
+  StringRef CanonicalMPIName = canonicalizeMPIName(Name);
```
vchuravy, line 733: I recently fixed a performance issue where we paid a lot for string allocations. Can we early exit when we know that it can't be an MPI name?

enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -1148,12 +1162,20 @@ void AdjointGenerator::handleMPI(llvm::CallInst &call, llvm::Function *called,
...
-        ss << " call: " << call << "\n";
-        ss << " unhandled mpi_reduce op: " << *orig_op << "\n";
-        EmitNoDerivativeError(ss.str(), call, gutils, BuilderZ);
-        return;
+        if (fortranABI) {
+          // Integer MPI operator handles of the Fortran ABI are
+          // implementation-defined and cannot be mapped portably at compile
+          // time; warn and assume the common case of MPI_SUM.
+          llvm::errs() << "warning: cannot determine MPI op used in `" << call
+                       << "`, assuming MPI_SUM\n";
```
vchuravy: Is this still true with MPI 5, and the new ABI?

enzyme/Enzyme/CallDerivatives.cpp
```diff
@@ -2204,6 +2244,81 @@ void AdjointGenerator::handleMPI(llvm::CallInst &call, llvm::Function *called,
...
   }
 
+  // Simple MPI functions that don't affect differentiation - just duplicate
+  // them in both forward and reverse passes. These functions query MPI
+  // state (communicator size, rank, processor name, etc.) but don't
+  // participate in the computation being differentiated.
+  if (funcName == "MPI_Comm_size" || funcName == "PMPI_Comm_size" ||
+      funcName == "MPI_Comm_rank" || funcName == "PMPI_Comm_rank" ||
+      funcName == "MPI_Get_processor_name" ||
+      funcName == "PMPI_Get_processor_name") {
```
vchuravy: Should this use the canoncalized names?

vchuravy: Why are they duplicated into the reverse pass?

enzyme/Enzyme/ActivityAnalysis.cpp
joewallwork: Is 3f1a139060b71bc8e5b5089c1faca2392fbf8e09 sufficient?

joewallwork: Proposed alternative in f098bd2acdb7c0c5fcdf57e1259e283be27c473f.

enzyme/Enzyme/CallDerivatives.cpp
joewallwork: You're right - it's not required. Removed in 03d8f591c91a6c75880d2ecb8ee4e64cfbc91338.

enzyme/Enzyme/CallDerivatives.cpp
joewallwork: Mostly yes - see b8c6d6d5fede05290e3386c567930cf99e1f12f8.

joewallwork:
Isend, Irecv, Wait seem to be harder to get working so I think it would be best to consider those in a follow-up PR.

.github/workflows/fortran.yml
joewallwork, line 1: This probably isn't the best solution long-term. Might be a good idea to create a container with OpenMPI and other dependencies preinstalled.

vchuravy (approved):

.github/workflows/fortran.yml
joewallwork: See #3196

PR #3184, opened 2026-09-05
Compute AugmentWithJuliaObjectType offsets from the DataLayout

vchuravy:
`AugmentWithJuliaObjectType` walks aggregate types to find the offsets of their Julia pointer fields, and obtained each offset by building a throwaway GEP and folding it:

```cpp
auto ud = UndefValue::get(getUnqual(ST));
auto g2 = GetElementPtrInst::Create(T, ud, vec);
APInt ai(DL.getIndexSizeInBits(g2->getPointerAddressSpace()), 0);
g2->accumulateConstantOffset(DL, ai);
delete g2;
```

That allocates and frees an LLVM instruction (plus a constant index vector and an undef operand) per struct field and per array type, to recover information the `DataLayout` already stores.

It matters because the function runs on **every** `TypeAnalyzer::getAnalysis(Value*)` call when `-enzyme-julia-addr-load` is set, so it is hot on Julia frontends. In a sampling profile of Enzyme differentiating an Oceananigans ocean model, 11s was charged to `GetElementPtrInst::Create` / `accumulateConstantOffset` underneath this function.

This replaces both with the layout queries directly:
- structs: `DL.getStructLayout(ST)->getElementOffset(i)`
- arrays: `DL.getTypeAllocSize(AT->getElementType())` as the stride, which is exactly what a `gep [0, 1]` into the array computes

Net: 9 insertions, 19 deletions, no allocation.

## Verification

Built old and new against LLVM 16 and diffed `-print-type-analysis` output with `-enzyme-julia-addr-load=1` over aggregates that exercise the changed lines — padded structs, packed structs, nested structs, arrays of pointers, arrays of structs, and nested arrays. **Output identical in every case.** For example `{ i8, ptr addrspace(10), i32, double, ptr addrspace(11) }` yields offsets 8 and 32 both before and after, and `[4 x [5 x ptr addrspace(10)]]` yields all 20 offsets 0…152.

The lit suite is also unchanged: same 1250 passed / 293 failed set before and after (the failures are pre-existing in my local config, which has MLIR/Fortran/Integration disabled; `TypeAnalysis`, `ActivityAnalysis` and `Enzyme/*` are fully green).

https://claude.ai/code/session_01Rj8Zm33cndJf3HGAKKeSHP

### DIFF
--- a/enzyme/Enzyme/GradientUtils.cpp
+++ b/enzyme/Enzyme/GradientUtils.cpp
@@ -6056,18 +6056,8 @@ Value *GradientUtils::invertPointerM(Value *const oval, IRBuilder<> &BuilderM,
     IRBuilder<> bb(newi->getNextNode());
 
     auto AggTy = arg->getAggregateOperand()->getType();
-    SmallVector<Value *, 4> vec;
-    vec.push_back(ConstantInt::get(Type::getInt64Ty(arg->getContext()), 0));
-    for (auto ind : arg->getIndices()) {
-      vec.push_back(ConstantInt::get(Type::getInt32Ty(arg->getContext()), ind));
-    }
-    auto ud = UndefValue::get(getUnqual(AggTy));
-    auto g2 = GetElementPtrInst::Create(AggTy, ud, vec);
-    APInt ai(DL.getIndexSizeInBits(g2->getPointerAddressSpace()), 0);
-    g2->accumulateConstantOffset(DL, ai);
-    delete g2;
-
-    unsigned Off = (unsigned)ai.getLimitedValue();
+    unsigned Off =
+        (unsigned)getAggregateElementOffset(DL, AggTy, arg->getIndices());
     auto ObjSize = (DL.getTypeSizeInBits(arg->getType()) + 7) / 8;
     auto AggSize = (DL.getTypeSizeInBits(AggTy) + 7) / 8;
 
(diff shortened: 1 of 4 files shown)


wsmoses (approved):
we definitely have other places in the code like this. @vchuravy can you search throughf or other cases of "delete g2" for similar replacements?

vchuravy:
Searched for the rest of them — there were 16 more sites building a throwaway GEP just to recover an offset the `DataLayout` already knows, now all converted in 05adcc99:

- `TypeAnalysis.cpp`: `getConstantAnalysis` (ConstantAggregate + ConstantDataSequential), `visitShuffleVectorInst` (×3), `visitExtractValueInst`, `visitInsertValueInst` (×2), the struct-returning libm call handling, and `defaultTypeTreeForLLVM`'s struct/array/vector cases
- `GradientUtils.cpp`: the `ExtractValueInst` and `InsertValueInst` shadows in `invertPointerM`
- `Utils.cpp`: `simplifyLoad` of an `extractvalue`

They all go through one new helper in `Utils.h`:

```cpp
static inline uint64_t getAggregateElementOffset(const llvm::DataLayout &DL,
                                                 llvm::Type *T,
                                                 llvm::ArrayRef<unsigned> Idxs);
```

which walks the indices using `getStructLayout()->getElementOffset()` for a struct field and `getTypeAllocSize()` of the element as the stride for an array or vector index (what a gep applies to a sequential index). There is a single-index overload for the many `for (i : 0..NumElements)` loops. No site allocates any more.

`insertvalue` needed a bit more care: its "offset of the next logical element" walk carried by mutating a vector of `ConstantInt`s, so that now carries over plain `unsigned` indices, with carrying past the outermost index landing one past the aggregate — what the old `gep [1]` over the aggregate computed.

Three sites I deliberately left alone, since they fold real GEPs rather than synthetic ones: `FunctionUtils.cpp:8846`, `FixupJuliaCallingConvention.cpp:491` and `SimpleGVN.cpp:289`.

Verified against a build of the parent commit: type-analysis output is identical on padded/packed/nested structs, nested extractvalue and insertvalue including carries out of the last element, shufflevector over double/float/ptr/i1 vectors, constant structs/arrays/vectors/i1-vectors, and struct-returning libm calls. Lit reports the same failure set before and after.

PR #3186, opened 2026-09-07
Make per-call lookup tables static

vchuravy:
`isInactiveCall` built a `StringSet` of ~130 known-inactive function names, a `std::set` of intrinsic IDs and several string arrays on every call, as the tables were function-local but not `static`. Activity analysis calls it from several hot paths. The same applied to `JuliaKnownTypes` in the type-analysis preparation loop (once per call instruction) and to the `NoFrees` / `NoFreeDemangles` sets in `CreateNoFree`.

Make all of them `static`.

Benchmark: Enzyme.jl reverse mode over 2 time steps of an 80×160×32 Oceananigans `HydrostaticFreeSurfaceModel` (Julia 1.12, LLVM 18), timing the first `autodiff` call. All numbers are from the same machine under `perf record`.

Measured together with #3187: Enzyme compile time 1054 s → 975 s. `isInactiveCall` and the `StringMap` rehash/lookup it caused disappear from the profile entirely. The lit suite has the same set of (pre-existing, opaque-pointer related) failures before and after.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/enzyme/Enzyme/ActivityAnalysis.cpp
+++ b/enzyme/Enzyme/ActivityAnalysis.cpp
@@ -225,7 +225,7 @@ struct {
 bool isInactiveCall(CallBase &CI) {
 
   // clang-format off
-const char *KnownInactiveFunctionsStartingWith[] = {
+static const char *KnownInactiveFunctionsStartingWith[] = {
     "f90io",
     "$ss5print",
     "strcpy",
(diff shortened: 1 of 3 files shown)


wsmoses (approved):

wsmoses:
clang-format fails

PR #3190, opened 2026-09-07
complex blas reverse mode gemv/gemm

ymardoukhi:
During JuliaCon, Billy told me that the infrastructure already exists for handling BLAS over complex numbers. So I gave it a shot with the help of Claude. This Intends to extend the support for AD over complex numbers. 

`lit` tests are skipped because I don't have the knowledge. If someone could guide me on how to write them, I will take care of it.

`cuBLAS` and support for the `transA=T` case for Fortran fall outside the scope of this PR.

* I relaxed the guard for `c/z` chars. 
* CBLAS uses values for real numbers; had to handle complex alpha/beta values via pointers
* `trans` flag swaps are checked and addressed
* `Constant<>` for complex values created `1.0 + i1.0`. Have it fixed
* `checkMemoryTrace` uses `elem_size` instead of `sizeof(double)`
* `zgemvTests` and `zgemmTests`


vchuravy:
Related #3181

PR #3193, opened 2026-09-08
Hand the frontend's external context to the callbacks that need it

vchuravy:
Hands the frontend's per-request context to the four callbacks that need it, so Enzyme.jl can stop keeping that state on the side.

## The problem

Enzyme calls back into the frontend through global function pointers, and none of them is told which differentiation request it is serving. Enzyme.jl needs to know: turning an LLVM type back into a Julia one only has an answer in a world age. Its workarounds today are a task-local variable holding the world, published around every `EnzymeCreate*` call and read back inside the callbacks, and reporting "unknown world" from the error handler when it is invoked without a `GradientUtils`, which loses the source location in the resulting Julia exception. Task-local state is also wrong under a frontend that differentiates from several tasks against separate logics.

## The design

`EnzymeLogic` already carries an opaque `void *ExternalContext` for the frontend, installed with `EnzymeLogicSetExternalContext`. This PR gives it a name, `EnzymeContextRef`, and makes it the first parameter of the callbacks whose Enzyme.jl implementations consult the world (EnzymeAD/Enzyme.jl#2643):

| Callback | Why it needs the context |
| --- | --- |
| `CustomErrorHandler` | the paths that reach it with a null `data` (the `Create*` bodies, batching, truncation, tracing, the calling-convention fixup, internal errors) have nothing else to recover the world from |
| `CustomAllocator` | boxes the tape type |
| `EnzymeFixupReturn` | materializes `nothing` |
| `EnzymeSanitizeDerivatives` | raises a Julia error |

The other callbacks (zeroing, deallocation, runtime-inactive errors, cache stores, tape type, undefined values, shadow allocation rewriting) keep their signatures: their Enzyme.jl implementations never consult the world, and the last already receives the `GradientUtils`.

The work is in getting the context from the logic down to each call site, reaching it through the object that already owns it wherever one exists:

| Carrier | Reaches |
| --- | --- |
| `EnzymeLogic::externalContext()` | the `CreateAugmentedPrimal` / `CreatePrimalAndGradient` / `CreateForwardDiff` / `CreateBatch` / `CreateTruncateFunc` / `CreateTrace` bodies |
| `CacheUtility::externalContext()`, read from the `EnzymeLogic &Logic` it now holds (moved down from `GradientUtils`) | `GradientUtils`, `DiffeGradientUtils`, `AdjointGenerator`, the BLAS derivatives and the caching machinery |
| `PreProcessCache::Logic`, the owning logic taken by reference at construction | preprocessing, which runs before any `GradientUtils` exists and reaches the allocator and the error handler itself |
| `TypeAnalyzer::externalContext()`, through its `TypeAnalysis`'s logic | type analysis errors |
| a `GradientUtils *` first parameter | the free helpers only ever called with one in hand: `getOrInsertDifferentialWaitallSave`, `SanitizeDerivatives`, `transpose` |
| an explicit `EnzymeContextRef` first parameter | `CreateAllocation`, `CreateReAllocation`, `EmitNoDerivativeError(RequestContext)`, `RecursivelyReplaceAddressSpace`, and the calling-convention fixup |

No copy of the context is stored anywhere below the logic: every carrier reads `EnzymeLogic::externalContext()` at the time of the call.

Two sites have no route to a logic and pass a null context: the type-depth warnings in `TypeTree`, which is a value type not owned by an analyzer. Their `data` is the tree, as before.

The Julia calling-convention fixup is a module pass, so it has nowhere to take a context from either. `EnzymeFixupJuliaCallingConventionModule` runs it over a module with one supplied, calling the per-function fixup directly; the named passes (`enzyme-fixup-julia`, `enzyme-fixup-julia-sret`) keep working and pass null. The batched fixup reaches no callback and is untouched.

## Breaking change

The ABI of the four callbacks above changes: a frontend must add the leading context parameter to each of them. Nothing else about the callbacks changes, and the existing `data` argument of the error handler (the `GradientUtils` or `TypeAnalyzer`, where there is one) is untouched.

The Julia CI job here builds Enzyme.jl main against this branch, so it will fail until the companion Enzyme.jl PR lands. That companion PR adds the parameter to the four `@cfunction` signatures, deletes the task-local world and its accessor, and switches the calling-convention fixup from the named pass to the new entry point.

## Test

Built against LLVM 20. `lit` over `enzyme/test/Enzyme`: 346 passed, 703 failed, 10 expectedly failed, an identical set of failures to `origin/main` built and run the same way (they are opaque-pointer expectation mismatches in this local configuration, not regressions).

https://claude.ai/code/session_01WZ9ygd17aLfCESakQx36ot

### DIFF
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -5241,12 +5241,11 @@ public:
           newcalled = BuilderZ.CreateExtractValue(newcalled, {0});
         }
 
-        ErrorIfRuntimeInactive(gutils->externalContext(), BuilderZ,
-                               gutils->getNewFromOriginal(callval), newcalled,
-                               "Attempting to call an indirect active function "
-                               "whose runtime value is inactive",
-                               gutils->getNewFromOriginal(call.getDebugLoc()),
-                               &call);
+        ErrorIfRuntimeInactive(
+            gutils, BuilderZ, gutils->getNewFromOriginal(callval), newcalled,
+            "Attempting to call an indirect active function "
+            "whose runtime value is inactive",
+            gutils->getNewFromOriginal(call.getDebugLoc()), &call);
 
         auto ft = call.getFunctionType();
         bool retActive = subretType != DIFFE_TYPE::CONSTANT;
@@ -5645,8 +5644,7 @@ public:
 
         if (Mode != DerivativeMode::ReverseModeGradient)
           ErrorIfRuntimeInactive(
-              gutils->externalContext(), BuilderZ,
-              gutils->getNewFromOriginal(callval), newcalled,
+              gutils, BuilderZ, gutils->getNewFromOriginal(callval), newcalled,
               "Attempting to call an indirect active function "
               "whose runtime value is inactive",
               gutils->getNewFromOriginal(call.getDebugLoc()), &call);
--- a/enzyme/Enzyme/AdjointGenerator.h
+++ b/enzyme/Enzyme/AdjointGenerator.h
@@ -4897,7 +4897,7 @@ public:
 
         if (tape && shouldFree()) {
           for (auto idx : subdata->tapeIndiciesToFree) {
-            CreateDealloc(gutils->externalContext(), Builder2,
+            CreateDealloc(Builder2,
                           idx == -1 ? tape
                                     : Builder2.CreateExtractValue(tape, idx));
           }
@@ -5122,8 +5122,7 @@ public:
             (writeOnlyNoCapture && readOnly);
 
         if (replace) {
-          argi = getUndefinedValueForType(gutils->externalContext(), M,
-                                          argi->getType());
+          argi = getUndefinedValueForType(M, argi->getType());
         }
         argsInverted.push_back(argTy);
         args.push_back(argi);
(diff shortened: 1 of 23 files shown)


vchuravy:
CI notes, so nobody has to dig:

- **Julia Enzyme CI fails, as expected.** `Build libEnzyme` succeeds; the test step segfaults everywhere, always the same way:

  ```
  [3856] signal (11.1): Segmentation fault
  _ZNK4llvm4Type19getScalarSizeInBitsEv at .../libLLVM-15jl.so
  ```

  That job checks out Enzyme.jl `main`, whose `@cfunction` callbacks still have the old arity, so the context lands in the register the callback reads as its `LLVMBuilderRef` and the first dereference goes through garbage. It is the change reaching the callbacks, and it is what the companion Enzyme.jl PR fixes. The two have to land together.

- **Benchmarking fails for an unrelated reason**, and has on every run of `main` for the past week: the external `cpu_benchmarks` lit config uses `execute_external=True`, which the lit in that image now rejects with `ValueError: execute_external=True is deprected as of LLVM-23`.

- Locally, `lit` over `enzyme/test/Enzyme` gives an identical pass/fail set on this branch and on a pristine `main` built the same way, and `clang-format` 16 (the version the Format job pins) reformats none of the changed files.

PR #3168, opened 2026-08-25
WIP: Grab bag of fortran fixes

vchuravy:

PR #3194, opened 2026-09-08
Mark inactive Fortran calls

joewallwork:
Refinement of #3168

This PR marks a few Fortran-related things as inactive in the activity analysis.

### DIFF
--- a/enzyme/Enzyme/ActivityAnalysis.cpp
+++ b/enzyme/Enzyme/ActivityAnalysis.cpp
@@ -227,8 +227,7 @@ bool isInactiveCall(CallBase &CI) {
   // clang-format off
 const char *KnownInactiveFunctionsStartingWith[] = {
     "f90io",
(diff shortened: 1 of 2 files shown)


vchuravy:
Some Claude review:

-   The `_FortranAio` prefix also matches flang's READ entry points (_FortranAioInputReal32/64, InputComplex*, InputDescriptor, InputUnformattedBlock), which overwrite user floating-point memory, so marking them inactive leaves the destination shadow un-zeroed and turns a loud failure into a silently wrong derivative.

-  Adding these names to the inactive lists is not sufficient for reverse mode, because the constant-fallback path calls EnzymeLogic::CreateNoFree on the declaration-only callee, which hard-errors since no `_FortranA*` name is in its NoFrees list.

-   The new comment's justification that "the comparison above is unsigned, so an entry at the any-offset index -1 is skipped" holds only for a 64-bit size_t, making the block's behavior platform-dependent and documenting an accident as intent.

PR #3195, opened 2026-09-09
Memoize isNoNeed and its direct-use queries in calculateUnusedValuesInFunction

vchuravy:
## Problem

In `calculateUnusedValuesInFunction`, `isNoNeed(v)` walks every user of `v`'s base object, and the `useneeded` / `instneeded` callbacks handed to `calculateUnusedValues` ask it once per (user, value) pair they examine. An allocation with U users therefore costs O(U^2) per visit, and the allocation is revisited whenever one of its users gets marked unnecessary, so the total is O(U^3), with each step potentially a full `is_use_directly_needed_in_reverse` query.

Julia 1.12's "inline roots" codegen produces exactly this shape: `alloca [427 x ptr addrspace(10)]` GC root arrays filled by 426 GEP+store pairs, dozens of them per function.

## Change

The recursive `std::function` lambda becomes a local `NoNeedQuery` struct holding both caches (no change to semantics or to the callbacks' interface — call sites still read `isNoNeed(v)`):

- `isNoNeed` results are cached per value with an epoch = `unnecessaryValues.size() + unnecessaryInstructions.size()`. The result depends only on those two sets. They only grow, and every check against them can only turn a user from "blocking" into "benign", so a `true` result is final and a `false` result stays valid until the epoch changes.
- `is_use_directly_needed_in_reverse(gutils, cur, mode, I, oldUnreachable, QueryType::Primal)` is cached per (value, user) pair: all of its inputs are fixed for the duration of the call.

The struct is declared inside the function, so the walker body is unchanged apart from indentation; `git diff -w` against main is +56/-9.

## Measured effect

SpeedyWeather.jl's Enzyme CI test (reverse-mode `time_step!` of `PrimitiveWetModel`, Julia 1.12.7, Enzyme.jl 0.13.202): with the unpatched library, the process sat in `calculateUnusedValuesInFunction` for the entire 90 minute CI limit. With this patch, the gradient-mode analysis of the nested `timestep!` (6,286 instructions) made 2,227,031 `isNoNeed` calls and 1,656,764 direct-use queries with 99.3% cache hits and finished in 625 ms; every other nested function finished in under 300 ms.

Note that the full SpeedyWeather case then proceeds to a separate activity-analysis hot spot in a 21k-instruction function; that is not addressed here.

Compile-checked against LLVM 18 (`ENZYME_EXTERNAL_SHARED_LIB=ON`, RelWithDebInfo); formatted with `git clang-format`.

### DIFF
--- a/enzyme/Enzyme/EnzymeLogic.cpp
+++ b/enzyme/Enzyme/EnzymeLogic.cpp
@@ -80,6 +80,7 @@
 #include "llvm/Support/AMDGPUMetadata.h"
 #include "llvm/Support/TimeProfiler.h"
 
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/StringSet.h"
 
 #include "DiffeGradientUtils.h"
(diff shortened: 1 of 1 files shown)


enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -855,6 +894,19 @@ void calculateUnusedValuesInFunction(
     }
     return false;
   };
+  isNoNeed = [&](const llvm::Value *v) {
```
wsmoses: can we avoid lambda functions?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -106,6 +107,11 @@ llvm::cl::opt<bool>
                            cl::Hidden,
                            cl::desc("Print unnecessary values in function"));
```
wsmoses, line 109: once demonstrated can we rm this var?

enzyme/Enzyme/EnzymeLogic.cpp
```diff
@@ -756,105 +757,151 @@ void calculateUnusedValuesInFunction(
...
+  struct NoNeedQuery {
+    const SmallPtrSetImpl<const Value *> &unnecessaryValues;
+    const SmallPtrSetImpl<const Instruction *> &unnecessaryInstructions;
+    GradientUtils *gutils;
+    DerivativeMode mode;
+    TargetLibraryInfo &TLI;
+    ArrayRef<DIFFE_TYPE> constant_args;
+    const SmallPtrSetImpl<BasicBlock *> &oldUnreachable;
+    DenseMap<const Value *, std::pair<size_t, bool>> cache{};
+    DenseMap<std::pair<const Value *, const Instruction *>, bool>
```
vchuravy, line 779: @wsmoses the closure free version is not nicer to read :)

PR #3210, opened 2026-09-11
Merge duplicated gc_preserve_ends before differentiating

### LINKED ISSUE #3565: LLVM ERROR: function failed verification — reverse-mode through an OrdinaryDiffEq solution interpolant (ode_interpolation)
## Summary

Reverse-mode `Enzyme.autodiff` aborts the process with `LLVM ERROR: function
failed verification (4)` when the differentiated function evaluates an
`OrdinaryDiffEq` solution's dense interpolant `sol(t)`. The failing IR is
`@preprocess_julia_ode_interpolation_… (…::OrdinaryDiffEqCore.InterpolationData…)`
— Enzyme cannot compile the derivative of `ode_interpolation`. Removing the
interpolant call (or replacing it with a plain analytic function of `t`) makes it
compile and run.

This surfaced through SciMLSensitivity's `EnzymeVJP` adjoint path, but reproduces
with only `OrdinaryDiffEqTsit5` + `Enzyme` — no SciMLSensitivity.

## MWE

```julia
using OrdinaryDiffEqTsit5, Enzyme, LinearAlgebra

A = [-0.3 1.0; -1.0 -0.3]
base!(du,u,p,t) = (mul!(du, A, u); nothing)
ref = solve(ODEProblem(base!, [1.0, 0.5], (0.0, 3.0)), Tsit5();
            abstol=1e-10, reltol=1e-10, dense=true, save_everystep=true)

# in-place RHS that evaluates the interpolant ref(t)
function forced!(du, u, p, t)
    mul!(du, A, u)
    r = ref(t)                 # <-- differentiating through this is what crashes
    du[1] += p[1]*r[1]
    du[2] += p[1]*r[2]
    return nothing
end

u = [1.0, 0.5]; p = [0.0]; t = 1.5
du = zeros(2); bdu = [1.0, 0.0]      # cotangent seed for the reverse pass
bu = zeros(2); bp = zeros(1)

Enzyme.autodiff(Enzyme.Reverse, forced!, Enzyme.Const,
    Enzyme.Duplicated(du, copy(bdu)),
    Enzyme.Duplicated(u, bu),
    Enzyme.Duplicated(p, bp),
    Enzyme.Const(t))
println("SUCCESS: ", bu, " ", bp)     # never reached
```

## Output

```
LLVM ERROR: function failed verification (4)

[..] signal (6): Aborted
```

with the failing function named `preprocess_julia_ode_interpolation_14438`, whose
argument type is `OrdinaryDiffEqCore.InterpolationData{ODEFunction{…}, …,
OrdinaryDiffEqTsit5.Tsit5Cache{…}, …}`. `SUCCESS` is never printed.

Reproduces identically on **Julia 1.11.9** (Enzyme's bundled `libLLVM-16jl`) and
**Julia 1.12.4** (`libLLVM-18.1.7`).

## `versioninfo()`

```
Julia Version 1.12.4
Commit 01a2eadb047 (2026-01-06 16:56 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 128 × AMD EPYC 7502 32-Core Processor
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver2)
```

Enzyme v0.13.202, OrdinaryDiffEqTsit5 v2.1.4.

vchuravy:
Fixes EnzymeAD/Enzyme.jl#3565, where reverse-mode `Enzyme.autodiff` through an OrdinaryDiffEq solution interpolant aborts with `LLVM ERROR: function failed verification (4)`.

## The problem

An `llvm.julia.gc_preserve_begin` can legally end up with more than one `llvm.julia.gc_preserve_end`. The token it produces may have many uses, and LLVM only refuses to clone a block when the block *produces* a token, so a pass such as JumpThreading is free to duplicate a block that merely consumes one. Julia's own pipeline does exactly that — for the `GC.@preserve` around the multi-exit `alwaysinline` `llvmcall` in FindFirstFunctions, whose result the following branch tests — so Enzyme is handed IR with one begin and two ends. (Verified by counting ends per begin at the top of `preprocessForClone`, before any Enzyme pass runs.)

Reverse mode mirrors a preserve region by emitting a `gc_preserve_begin` at the invert of the end and the matching `gc_preserve_end` at the invert of the begin, which is well-formed only when a single end post-dominates the begin. With several ends, each end visited in `CallDerivatives.cpp` RAUWs and erases whatever sits in `invertedPointers[begin_call]`, so exactly one mirrored begin survives, in exactly one of the invert blocks, and it does not dominate the mirrored end:

```
Instruction does not dominate all uses!
  %1435 = call token (...) @llvm.julia.gc_preserve_begin(ptr addrspace(10) %438)
  call void @llvm.julia.gc_preserve_end(token %1435)
LLVM ERROR: function failed verification (4)
```

## The fix

Restore the one-end invariant at the end of `preprocessForClone`, before differentiating:

1. Replace the duplicates with a single end at their nearest common post-dominator. The preserved region can only grow, which is always safe.
2. Where that block is also reachable without ever running the begin, split the edges leaving the begin's dominator subtree off into a block of their own first, so the merged end stays dominated by the token it consumes. This is guarded by a check that those edges are the only way out of the subtree; if they are not, the region is left alone rather than moved somewhere a path could bypass.

Post-dominance here has to ignore blocks that can only lead to an abort. LLVM's `PostDominatorTree` counts every `unreachable` as a function exit and Julia ends every bounds check in one, so `findNearestCommonDominator` returns the virtual root for essentially any pair of blocks. Those paths never return, and a preserve region left open along one of them is harmless, so the pass prunes `getGuaranteedUnreachable` blocks and runs Cooper-Harvey-Kennedy over what is left.

## Testing

* New `test/Enzyme/ReverseMode/gcpres-multiend.ll` covers both paths — ends rejoining inside the region, and ends rejoining at a block reachable without the begin. Confirmed to fail (verifier abort) with the call to the new pass disabled, and to pass with it.
* Full Enzyme lit suite: no new failures.
* The issue's MWE now returns `[-0.3, 1.0] [0.3631194734328978]`, matching what an `EnzymeRules.inactive` workaround on the search function produces.
* Full Enzyme.jl test suite against the patched `libEnzyme` (Julia 1.12, LLVM 18): 218194 pass, 60 broken, 0 fail.

Marked draft: I would like a second opinion on the edge-split in step 2, and it has only been exercised on LLVM 18 locally.

https://claude.ai/code/session_01Saygneos7bXvv7kgMh38yU

### DIFF
--- a/enzyme/Enzyme/FunctionUtils.cpp
+++ b/enzyme/Enzyme/FunctionUtils.cpp
@@ -64,6 +64,7 @@
 #include "llvm/CodeGen/UnreachableBlockElim.h"
 
 #include "llvm/Analysis/PhiValues.h"
+#include "llvm/Analysis/PostDominators.h"
 #include "llvm/Analysis/ProfileSummaryInfo.h"
 #include "llvm/Analysis/ScalarEvolution.h"
 #include "llvm/Analysis/ScopedNoAliasAA.h"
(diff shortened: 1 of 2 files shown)


wsmoses (approved):
not checking the impl in depth, @vchuravy I'm going to trust you validated it. If sufficiently complex we can make it a proper pass with separate tests but fine with this for now

PR #3211, opened 2026-09-11
Defer differentiation from an LTO pre-link pipeline to the post-link run

vchuravy:
`augmentPassBuilder` registers Enzyme on two extension points: `OptimizerEarly`, which runs per translation unit, and `FullLinkTimeOptimizationEarly`, which runs inside the linker. With `-flto` **both** fire, and the pre-link one fires first — on a module that is only one translation unit. So a `__enzyme_autodiff` call naming a function defined in another object dies with `Enzyme: failed to find fn to differentiate` before the link-time run ever gets to see the whole program.

That forces an either/or on users: load the plugin in the compiler (`-fpass-plugin=`) or in the linker (`-Wl,--load-pass-plugin=`), but not both — even though on LLVM ≥ 16 `FlangEnzyme`/`ClangEnzyme`/`LLDEnzyme` are the same library behind the same `llvmGetPassPluginInfo`, and `LLDEnzymeFlags` already hands out `-flto` as a *compile* option next to the linker plugin.

Since LLVM 20 the `OptimizerEarly` callback receives the `ThinOrFullLTOPhase`, which Enzyme discarded:

```c++
auto loadPass = [prePass](ModulePassManager &MPM, OptimizationLevel Level,
                          ThinOrFullLTOPhase)
```

This PR uses it: skip the body at `FullLTOPreLink` and `ThinLTOPreLink` and let the post-link run do the work — `FullLinkTimeOptimizationEarly` for full LTO, this same callback re-entered with `ThinLTOPostLink` for ThinLTO. `PreserveNVVM(/*Begin=*/true)` still runs at pre-link, so the functions involved stay alive until then, and `loadLTO` already calls `loadPass` with an explicit `ThinOrFullLTOPhase::None`, so the link-time path is unaffected.

The effect is that "plugin loaded in both the compiler and the linker" becomes the correct default: compile-time Enzyme is a no-op exactly when a link-time run is going to see more.

### Caveats

* This requires the plugin to be loaded by the linker as well. Where that is not possible, `-mllvm -enzyme-lto-prelink=1` restores the old behaviour. Forgetting it is a loud failure (`undefined reference to __enzyme_autodiff...`), not silently wrong derivatives.
* Non-LTO builds are unchanged, and LLVM < 20 is untouched — the phase is not available on that callback there.

### Testing

Verified against LLVM 22 (`opt` standing in for the linker, which is where the post-link pipelines actually run):

| pipeline | before | after |
| --- | --- | --- |
| `default<O2>` | differentiates | differentiates |
| `lto-pre-link<O2>` | `failed to find fn to differentiate` | deferred |
| `thinlto-pre-link<O2>` | `failed to find fn to differentiate` | deferred |
| `lto<O2>` | differentiates | differentiates |
| `thinlto<O2>` | differentiates | differentiates |

End to end, `d/dx square(x)` across two translation units now folds to `fmul double %0, 2.000000e+00` in the post-link module instead of failing at compile time.

`enzyme/test/Integration/ForwardMode/lto_prelink.c` covers all of the above. It needs a new lit substitution, `%optLoadClangEnzyme`, to load a plugin that actually registers the extension points — `%loadEnzyme` points at `LLVMEnzyme`, which is built without `ENZYME_RUNPASS` and so never calls `augmentPassBuilder`.

`check-enzyme`, `check-enzyme-activity` and `check-enzyme-typeanalysis` show identical results with and without the patch (unsurprising: they all run through `LLVMEnzyme`, which does not reach this code).

### Context

Came out of https://github.com/EnzymeAD/flang-test-case, where the flang and lld routes had to be separate fpm profiles for exactly this reason. With this change they compose into one.

https://claude.ai/code/session_01TC5oeSdtsztrJspe4tFCJx

### DIFF
--- a/enzyme/Enzyme/Enzyme.cpp
+++ b/enzyme/Enzyme/Enzyme.cpp
@@ -101,6 +101,11 @@ using namespace llvm;
 llvm::cl::opt<bool> EnzymeEnable("enzyme-enable", cl::init(true), cl::Hidden,
                                  cl::desc("Run the Enzyme pass"));
 
+llvm::cl::opt<bool> EnzymeLTOPreLink(
+    "enzyme-lto-prelink", cl::init(false), cl::Hidden,
+    cl::desc("Run Enzyme in an LTO pre-link pipeline, rather than deferring "
+             "differentiation to the post-link run"));
+
 llvm::cl::opt<bool>
     EnzymePostOpt("enzyme-postopt", cl::init(false), cl::Hidden,
                   cl::desc("Run enzymepostprocessing optimizations"));
@@ -3151,7 +3156,7 @@ void augmentPassBuilder(llvm::PassBuilder &PB) {
 
 #if LLVM_VERSION_MAJOR >= 20
   auto loadPass = [prePass](ModulePassManager &MPM, OptimizationLevel Level,
-                            ThinOrFullLTOPhase)
+                            ThinOrFullLTOPhase Phase)
 #else
   auto loadPass = [prePass](ModulePassManager &MPM, OptimizationLevel Level)
 #endif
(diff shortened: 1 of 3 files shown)


SpeedyTurtle599:
This matches up with a thin LTO issue I've been having with C++. I attacked the problem slightly differently in my private patches: instead of deferring everything until link time, I've been looking at requests. So, at a pre-link phase, I differentiate any request whose primal is defined in a given module, and defer **only** requests whose primal is still a declaration. Right now (on main), a missing body is a hard `failed to find fn to differentiate`, but the surgical deferral lets you leave calls in place and mark functions held. That keeps all of the same-TU requests (for example, no RDC) behaving exactly as they do today. The thin link pass only has to work on cross-module requests. If this approach is interesting, happy to post the PR...

vchuravy:
@SpeedyTurtle599 that might indeed be of interest.

PR #3213, opened 2026-09-11
Allow a constant value to have a cached shadow in forward mode

vchuravy:
`forwardModeInvertedPointerFallback` asserts that a constant value has no entry in `invertedPointers`, and aborts on

```
Assertion `found == gutils->invertedPointers.end()' failed.
```

That invariant does not hold under runtime activity. `invertPointerM` can be asked to invert a constant operand — its `InsertValueInst` case skips the mixed-activity check when `runtimeActivity` is set, and then inverts both operands regardless of their activity — and its `ExtractValueInst` case caches the shadow it builds.

The cached entry is a real shadow rather than a placeholder, since `forceAugmentedReturns` only creates the `'dual_phi` placeholders for non-constant values. So there is nothing to clean up: dropping the assert and returning leaves a valid memoized shadow in place for later reuse, and `invertPointerM` consults that cache before any activity check.

No IR changes — only what a debug build tolerates.

> Per review feedback: an earlier revision of this PR instead suppressed the cache insert in the `ExtractValueInst` case. Fixing the assert is narrower and keeps the memoization.

### How it was found

Batched forward mode over a Turing model that calls `sample` internally, from EnzymeAD/Enzyme.jl#2464. Printing the offending map entry at the assert showed it was an `insertvalue [3 x i64]` built from a `'ipev` value, which ruled out `forceAugmentedReturns` as its origin and pointed at the `ExtractValueInst` branch.

Runtime activity is required. Without `set_runtime_activity` the mixed-activity check is not skipped, and Enzyme reports a proper `EnzymeRuntimeActivityError` instead of aborting.

### Testing

No regression test, and I want to flag that rather than bury it. I could not reduce the trigger below the original Turing program: four Julia variants and several hand-written IR shapes (mixed integer/float aggregates, an all-integer `[4 x i64]` aggregate feeding an `insertvalue` with an active double, width 3, runtime activity on, the aggregate argument const) all compile cleanly on an unpatched build. Whether the `extractvalue` shadow is computed at all depends on the order in which surrounding shadows get materialized, which is what makes it hard to provoke in a small function. Happy to keep trying if you would rather this not land untested.

What I did verify, with a local LLVM 18 build on Julia 1.12 (against the previous revision of the patch, which fixes the same abort):

- the abort is gone, and that program proceeds to a separate, unrelated failure
- `test/basic.jl`, `test/sugar.jl`, `test/mixed.jl`, `test/applyiter.jl` and `test/runtime_calls.jl` in Enzyme.jl all pass against the patched library


wsmoses:
I feel like the assertion is more the problem than the fact this is cached

enzyme/Enzyme/AdjointGenerator.h
```diff
@@ -376,12 +376,18 @@ class AdjointGenerator : public llvm::InstVisitor<AdjointGenerator> {
...
     using namespace llvm;
 
-    auto found = gutils->invertedPointers.find(&I);
     if (gutils->isConstantValue(&I)) {
-      assert(found == gutils->invertedPointers.end());
+      // A constant value may still have an entry here. Under runtime activity
+      // invertPointerM can be asked to invert a constant operand -- its
+      // InsertValueInst case skips the mixed-activity check when
+      // runtimeActivity is set, and then inverts both operands regardless of
+      // their activity -- and it caches the shadow it builds. Such an entry is
```
wsmoses, line 384: I would remove this text, it's historical and not needdd -- but do add a test

PR #3224, opened 2026-09-15
Fix: avoid pruning `Const` arguments needed by a custom derivative rule

Tuna2222:
## Summary

A `Const` argument that a custom `EnzymeRules` derivative rule reads or writes, but that the primal never touches, was being dropped from a differentiated function's signature and replaced with `jl_nothing` whenever the call crossed a non-inlined function. Reported in EnzymeAD/Enzyme.jl#3570.

`shouldDisableNoWrite` (`Utils.h`) trusted write-only/no-capture attribute inference for any call to a resolved callee, as long as neither the call nor the callee carried an explicit custom-rule marker. That inference is computed from the callee's own primal semantics but a custom derivative rule actually needs the argument. `shouldDisableNoWrite` now also disables the no-write shortcut for any resolved callee with a body Enzyme will still analyze. Since a Const argument is potentially be read/written, `readnone`/`readonly`/`writeonly` attributes are also dropped wherever they get copied onto a Const argument. But only where the argument is actually passed to a call `shouldDisableNoWrite` disables trust for.

## Tests

- `ForwardMode/callee_const_arg.ll` — a Const pointer argument of a resolved, non-custom-rule callee must not be pruned to 
  `undef` by `shouldDisableNoWrite`'s strategy.
- `ReverseMode/custom_const_state.ll` — a Const pointer argument read and written by a combined 
  `!enzyme_augment`/`!enzyme_gradient` custom rule must reach the generated call as a real pointer, and the generated derivative's definition and call site must not keep a stale `readnone`/`readonly`/`writeonly` claim. Covers all three attributes as separate variants, with the attribute asserted directly as a parameter attribute (not inferred from a whole-call attribute) so each variant exercises both copy sites independent of pass ordering. 

See EnzymeAD/Enzyme.jl#3570.


vchuravy:
Alternative https://github.com/EnzymeAD/Enzyme.jl/pull/3610

PR #3227, opened 2026-09-15
[MLIR] Add reverse rule for affine if

Pangoraw:
Currently emitted as a `scf.if` with materialized condition.


enzyme/Enzyme/MLIR/Implementations/AffineAutoDiffOpInterfaceImpl.cpp
```diff
@@ -67,6 +67,159 @@ affine::AffineIfOp createAffineIfWithShadows(Operation *op, OpBuilder &builder,
...
+    SmallVector<Value> dims = llvm::map_to_vector(
+        ifOp->getOperands().take_front(is.getNumDims()), getNewFromOriginal);
+    SmallVector<Value> symbols = llvm::map_to_vector(
+        ifOp->getOperands().drop_front(is.getNumDims()), getNewFromOriginal);
+
+    for (auto [eq, E] : llvm::zip_equal(is.getEqFlags(), is.getConstraints())) {
+      AffineExpr lhsExpr = E, rhsExpr = cacheBuilder.getAffineConstantExpr(0);
+
+      while (lhsExpr.getKind() == AffineExprKind::Add) {
+        auto binExpr = cast<AffineBinaryOpExpr>(lhsExpr);
```
wsmoses: There must be this code elsewhere eg during the affine if lowering?

Pangoraw: the easiest is to emit `E >= 0` as the llvm lowering:

https://github.com/llvm/llvm-project/blob/a125e6ed795798ee87376a82f90cb3fd897b25ca/mlir/lib/Conversion/AffineToStandard/AffineToStandard.cpp#L291-L309

but it's not doing the peeling (which could be fine too)

wsmoses: Do we need the peeling for perf, wouldn't the equivalent be resolved by mincut?

Pangoraw: not really needed, it's nothing llvm/mlir opts can't recover probably

wsmoses: Okay let's simplify for now and revisit after if needed for perf

Pangoraw: removed it 👍

wsmoses (approved):
didn't look terribly closely, but lgtm at a high skim if you checked the details

PR #3228, opened 2026-09-15
TypeAnalysis: propagate a gep's pointee per edge through correlated phis

### LINKED ISSUE #3373: 1.13+ illegal type analysis due to memory itself being unstable internally
```julia
using Enzyme

x = Float32[3]
w = Float32[1]
dw = zero(w)

function inactiveArg(w, x, cond)
    if cond
        x = copy(x)
    end
    @inbounds w[1] * x[1]
end

# Throws IllegalTypeAnalysisException
Enzyme.autodiff(set_runtime_activity(Reverse), inactiveArg, Active, Duplicated(w, dw), Const(x), Const(false))
```

vchuravy:
A phi of derived pointers may have been split into a base phi and an offset phi that are correlated per incoming edge (Enzyme.jl's `nodecayed_phis!` does this for addrspace(11) phis):

```llvm
%base = phi ptr addrspace(10) [ %a, %bb1 ], [ %b, %bb2 ]
%off  = phi i64 [ 8, %bb1 ], [ 0, %bb2 ]
%p    = getelementptr i8, addrspacecast(%base), %off
```

`%a` is only ever indexed by 8 and `%b` only by 0, and the two need not share a layout (e.g. a `Memory{T}` and an `Array{T}`). The upward rule of `visitGEPOperator` enumerated the known values of `%off` and applied every offset to the merged base, asserting both layouts on both objects; the phi rule then pushed that onto each incoming, and a fresh `Memory` whose length field is an integer was told its offset 0 is a pointer, aborting with `Illegal updateAnalysis`.

On Julia 1.13 this is reached from `cond && (x = copy(x)); x[1]` (the `Copy Broadcast arg` test of Enzyme.jl): the `Array` wrapper built by `copy` is eliminated and LLVM 20 folds the two data-pointer loads into one load of a phi of addresses. Fixes EnzymeAD/Enzyme.jl#3373.

When the gep has a single variable index that is a phi, and its pointer operand (through casts) is a phi of the same block whose edges pair with constant offsets, push the gep's pointee type into the incoming object of each edge at that edge's offset instead of into the merged base. The base phi then only learns what its incomings agree on through the downward phi rule, which is already an intersection.

Validation:
- new `test/TypeAnalysis/gepphiperedge.ll` fails on main with `Illegal updateAnalysis` and passes with this change; the rest of the TypeAnalysis directory is unchanged
- with a build against Julia 1.13.0's LLVM 20.1.8, the Enzyme.jl reproducer gives the correct gradient and `test/advanced.jl` goes from 217 pass / 1 error / 2 broken to 224 pass / 2 broken


enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -2196,6 +2196,61 @@ void TypeAnalyzer::visitGEPOperator(GEPOperator &gep) {
...
+        offPhi->getNumIncomingValues() == basePhi->getNumIncomingValues()) {
+      APInt scale = VariableOffsets.begin()->second;
+      SmallVector<std::pair<Value *, int64_t>, 4> edges;
+      bool legal = true;
+      for (unsigned i = 0, e = offPhi->getNumIncomingValues(); i < e; ++i) {
+        auto CI = dyn_cast<ConstantInt>(offPhi->getIncomingValue(i));
+        if (!CI ||
+            offPhi->getIncomingBlock(i) != basePhi->getIncomingBlock(i)) {
+          legal = false;
+          break;
```
vchuravy: This assumes that the IR edges are always:

```
  %base = phi ptr [ %mem, %a ], [ %arr, %b ]
  %off = phi i64 [ 8, %a ], [ 0, %b ]
```

and never

```
  %base = phi ptr [ %mem, %a ], [ %arr, %b ]
  %off = phi i64 [ 8, %b ], [ 0, %a ]
```

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -2196,6 +2196,62 @@ void TypeAnalyzer::visitGEPOperator(GEPOperator &gep) {
...
+      ptr = CI->getOperand(0);
+    auto basePhi = dyn_cast<PHINode>(ptr);
+    if (offPhi && basePhi && offPhi->getParent() == basePhi->getParent()) {
+      APInt scale = VariableOffsets.begin()->second;
+      SmallVector<std::pair<Value *, int64_t>, 4> edges;
+      bool legal = true;
+      for (unsigned i = 0, e = offPhi->getNumIncomingValues(); i < e; ++i) {
+        auto CI = dyn_cast<ConstantInt>(offPhi->getIncomingValue(i));
+        // The two phis need not list their predecessors in the same order,
+        // pair the edges by incoming block rather than by operand index.
```
vchuravy:
```suggestion
```

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -2196,6 +2196,60 @@ void TypeAnalyzer::visitGEPOperator(GEPOperator &gep) {
...
+  // then only learns what its incomings agree on.
+  bool upPerEdge = false;
+  if ((direction & UP) && VariableOffsets.size() == 1) {
+    auto offPhi = dyn_cast<PHINode>(VariableOffsets.begin()->first);
+    Value *ptr = gep.getPointerOperand();
+    while (auto CI = dyn_cast<CastInst>(ptr))
+      ptr = CI->getOperand(0);
+    auto basePhi = dyn_cast<PHINode>(ptr);
+    if (offPhi && basePhi && offPhi->getParent() == basePhi->getParent()) {
+      APInt scale = VariableOffsets.begin()->second;
```
wsmoses: I don't think this is the right resolution, instead I would thread the phi around:

  // Stores pair ([whether first offset is zero], offset)
  std::vector<std::pair<bool, int>> offsets;
  Value *firstIdx = *gep.idx_begin();
  if (VariableOffsets.size() == 0) {
    bool firstIsZero = cast<ConstantInt>(firstIdx)->getLimitedValue() == 0;
    offsets.emplace_back(firstIsZero, (int)constOffset.getLimitedValue());
  } else {
    bool firstIsZero = false;
    if (auto CI = dyn_cast<ConstantInt>(firstIdx))
      firstIsZero = CI->getLimitedValue() == 0;
    for (auto vec : getSet<int>(idnext, idnext.size() - 1)) {
      APInt nextOffset = constOffset;
      for (auto [varpair, const_value] : llvm::zip(VariableOffsets, vec)) {
        nextOffset += varpair.second * const_value;
        if (varpair.first == firstIdx)
          firstIsZero = const_value == 0;
      }
      offsets.emplace_back(firstIsZero, (int)nextOffset.getLimitedValue());
    }
  }

  bool seenIdx = false;

  for (auto [firstIsZero, off] : offsets) {
    // TODO also allow negative offsets

and then we would do the offset and other calculation per edge [and we should do it for all edges not just a chosen one]

enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
```diff
@@ -2220,91 +2218,191 @@ void TypeAnalyzer::visitGEPOperator(GEPOperator &gep) {
...
+    if (edge.base != basePhi)
+      updateAnalysis(edge.base, upTree.Only(-1, inst), &gep);
   }
+  if (!seenIdx)
+    return;
   if (direction & DOWN)
     updateAnalysis(&gep, downTree.Only(-1, inst), &gep);
-  if (direction & UP)
-    updateAnalysis(gep.getPointerOperand(), upTree.Only(-1, inst), &gep);
+  if ((direction & UP) && seenMergedUp)
```
wsmoses: I don't understand why this is changed

PR #3229, opened 2026-09-15
Forward fast-math flags instead of overwriting

SpeedyTurtle599:
Currently, Enzyme forces fast math on, changing the semantics of some of my numerics. This PR aims to fix that: each primal carries its own fast flag as specced in the compilation line, with the usual `enzyme-fast-math` flag available to override. Most of the diff is test changes (removing fast flags and changing emitted IR expectations).

I view this as an invasive, philosophically-different change, so I would very much welcome feedback.


wsmoses (commented):
let's try this in stages. I think having a getFastFrom helper that when enzyme-fast-math is off uses the instr's fp flags is good and mergable atm. The question of the right setting of that flag can be separated into a second follow up discussion. Not oppossed but there are arguments on both sides

PR #3231, opened 2026-09-15
Use `enzyme_out` in Fortran dot tests

### LINKED ISSUE #3230: Test `enzyme_out` with Fortran
I saw some issues with this activity descriptor a while ago so we should check whether there are any issues.

joewallwork:
Closes #3230.

This PR converts the `dot` Fortran tests to use subroutines rather than functions and uses `enzyme_out` for the output argument.


joewallwork:
Closing because it appears the test failure is due to the fact that Enzyme doesn't support `enzyme_out` being applied to pointer arguments. (Arguments are passed by reference by default in Fortran.)

PR #3236, opened 2026-09-15
[MLIR] fix non-vector subroutine call tblgen crash

Pangoraw:


enzyme/Enzyme/MLIR/Implementations/ArithDerivatives.td
```diff
@@ -39,11 +39,11 @@ def : MLIRDerivative<"arith", "MaxNumFOp", (Op $x, $y),
                   >;
 
 def IsPosZero : SubRoutine<(Op $x), (CmpF (Arith_OEQ),
-                                          (CopySignF (ConstantFP<"1","arith","ConstantOp">), $x),
-                                          (ConstantFP<"1","arith","ConstantOp">))>;
+                                          (CopySignF (ConstantFP<"1","arith","ConstantOp"> $x), $x),
```
wsmoses, line 42: why didn't we get an error for this earlier? can we fix the error mechanism in tablegen [and also add a test to confirm both these changes are needed/tested]

Pangoraw: The tblgen was making ci red (which I missed when merging with auto queue), I don't think this one is needed since op->getResultType(0) should be the same type as $x

Pangoraw: actually needed for batch ad

